PR #1, opened 2017-04-25
Create empty Symfony 3.2 project

nicwortel:


bin/symfony_requirements
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env php
+<?php
```
DRvanR: is this relevant to keep in the repository?

nicwortel:
Most if not all of the steps you mention were indeed planned for future PRs. I have included some of them in this one and others will be done in future PRs.

As for running tests and building the project I intend to use a Makefile, so setting that up will be done manually.

I prefer to keep app_dev.php in the repository and remove it (along with other dev-only files such as test, composer.json, etc.) as part of the build process. It will need to be filtered anyway when deploying from a development machine. Either way, I will make sure that it doesn't end up on non-development environments.

bin/symfony_requirements
nicwortel: I do think it is useful, to check the PHP CLI configuration from inside the Vagrant box.

DRvanR:
>  I intend to use a Makefile

please keep it in line with the rest of the projects in the platform and use ant by default, optionally supplying a makefile.

nicwortel:
> please keep it in line with the rest of the projects in the platform and use ant by default

Alright :disappointed:

ansible/roles/app/templates/support.conf.j2
```diff
@@ -8,51 +8,25 @@ Listen {{ apache_app_listen_address.support }}:{{ loadbalancing.support.port }}
     ServerAdmin {{ apache_server_admin }}
     Options -indexes +Includes
     DocumentRoot {{ document_root }}
-    Alias /registration {{ support_data_dir }}/current/web
-    SetEnv HTTPS on
 
-    <Directory "{{ support_data_dir }}">
+    <Directory {{ document_root }}>
```
quartje: I'd like to see that the variable {{ document_root }} is changed to something less generic and more SP Dashboard specific to make it usable in the OpenConext deploy playbooks.

PR #3, opened 2017-04-25
Set up various quality assurance tools

nicwortel:


ansible/roles/app/templates/support.conf.j2
```diff
@@ -8,51 +8,25 @@ Listen {{ apache_app_listen_address.support }}:{{ loadbalancing.support.port }}
     ServerAdmin {{ apache_server_admin }}
     Options -indexes +Includes
     DocumentRoot {{ document_root }}
```
quartje: I'd like to change the var {{ document_root }} to something less generic and more SP dashboard specific, since this Ansible role will be used from a playbook in OpenConext-deploy as well.

.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: php
+
+php:
+  - 5.6
+  - 7.0
+  - 7.1
+
+cache:
+  directories:
+    - ~/.composer/cache
```
DRvanR: `~/.composer/cache/files`, see https://github.com/SURFnet/yubikey-api-client-bundle/commit/0c7e47cb7b299127e2413f57bf8da00e88a0be9e

PR #6, opened 2017-05-02
Improve documentation

nicwortel:


README.md
```diff
@@ -4,31 +4,49 @@
...
+[OpenConext-deploy](https://github.com/OpenConext/OpenConext-deploy), so in order to provision the Vagrant box you need
+to have that repository checked out in a directory called `OpenConext-deploy` in the parent directory of where this
+project lives.
+
 ## Getting started
 
-Install Composer dependencies using:
+First, run `composer install`. This will install all PHP dependencies, including the development dependencies.
+
+In order to start the development environment, run `vagrant up`. This will build and start the virtual machine that is
```
MKodde: Not an issue with the docs, but during `vagrant up`, Ansible failed on the `app : Add hosts file` with an `"unsupported parameter for module: path"` error message.

README.md
```diff
@@ -4,31 +4,49 @@
...
+used in development to run the application. When started for the first time, the Vagrant box will be provisioned using
+Ansible.
+
+Note: if you don't use the Vagrant Hostsupdater plugin, you have to manually add
+`192.168.33.19  dev.support.surfconext.nl` to your hosts file so that requests will be routed to the virtual machine.
```
MKodde: While checking my `hosts` file, I noticed this IP is used for another SurfConext project. `serviceregistry.dev.support.surfconext.nl`. It might be a good idea to use another IP for this project. Not really an issue with the documentation, just saying.

README.md
nicwortel: Ah. To be honest, the Ansible scripts were set up by Bart (@quartje). It seems that the `path` parameter of the `lineinfile` module was introduced very recently, in Ansible 2.3. We should probably either replace it with one of its aliases (`dest`, `destfile`, `name`) or document that Ansible 2.3+ is required.

@quartje what do you think?

README.md
nicwortel: Are you using the Hostsupdater plugin?

That domain is [listed as an alias in the Vagrantfile](https://github.com/SURFnet/sp-dashboard/blob/03de5d22c0c139254b5d213aa1681c532181eb58/Vagrantfile#L7), that's why it is added to the hosts file as well. As far as I know, there is no other project using the same IP address.

MKodde: The Service-provider-registration project uses the same address in it's Vagrantfile. Which might prove to be problematic when working on both projects at the same time (during reconstruction). The severity of this issue is so low that we can resolve it (if need be) during development.

README.md
quartje: I'd say keeping it backwards compatible seems the most efficient

MKodde: I'd like to change this parameter to `dest`. 

This worked for me and should be compatible with higher `Ansible` versions. I'll commit this change to this branch. The warning in the documentation [Nic mentioned](http://docs.ansible.com/ansible/lineinfile_module.html#notes) does not look line a future release deprication warning.

Are you OK with my proposal @quartje ?

README.md
DRvanR: my 0.02: would just let it be. If the problem arises, the SP-registration VM can be moved to a different IP as that project is/will be deprecated - this is the replacement and should aim at being stable.

README.md
quartje: Sounds good to me!

PR #7, opened 2017-09-19
Perform base installation and configuration

MKodde:
### Work in progress
For a checklist of the features to be installed, see [Pivotal](https://www.pivotaltracker.com/story/show/150963310).

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -23,7 +23,6 @@ framework:
         engines: ['twig']
     default_locale: '%locale%'
     trusted_hosts: ~
-    trusted_proxies: ~
     session:
         # http://symfony.com/doc/current/reference/configuration/framework.html#handler-id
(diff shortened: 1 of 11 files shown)


app/config/config.yml
```diff
@@ -68,3 +75,9 @@ swiftmailer:
     username: '%mailer_user%'
     password: '%mailer_password%'
     spool: { type: memory }
+
+assetic:
+    debug: '%kernel.debug%'
+    use_controller: '%kernel.debug%'
+    filters:
+        cssrewrite: ~
```
Tjeerd: Sure about adding assetic? This could/should have some consequences to deployments etc as well. But maybe you simply don't use it (yet)?

MKodde: Most SURFnet Symfony projects use Assetic. Using it should not pose any serious problems as these have been solved before.

PR #14, opened 2017-09-28
Implement initial page theming based on Teams-NG

jorissteyn:


app/scss/base/base.scss
```diff
@@ -0,0 +1,31 @@
+body {
+  font-family: "Roboto", Sans-serif;
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 20px;
+  color: $dark-grey;
+  -webkit-font-smoothing: antialiased;
+  background-color: $background;
+
+  &.ReactModal__Body--open {
```
MKodde: Rename or remove this? 

Same goes for the React mentions in and around: `app/scss/components/confirmation_dialog.scss:67`

MKodde:
@quartje can we ask you to include the installation of npm and yarn to the `sp-dashboard` task? I've created a task for this in Pivotal.

app/scss/base/base.scss
jorissteyn: I've removed all react-specific rules.

jorissteyn:
I don't think there's a common place to store non-public (scss) assets outside of bundles (and I'd like to keep them outside of the bundle). As discussed: we keep them in app/js and app/scss.

PR #16, opened 2017-09-29
Add webtest for the entity list action

jorissteyn:
This is just an exapmle testcase since there is nothing much to test
yet.


jorissteyn:
@MKodde shouldn' we call this "service list"?

MKodde:
@jorissteyn Yes, it should. As discussed, we'll rename the `EntityListController` next time when one of us need to touch it.

PR #13, opened 2017-09-27
Add the Supplier form

MKodde:
In this PR, the following tasks have been performed

- [x] Install Tactician
- [x] Create the Supplier form
- [x] Build Command & CommandHandler

Todo
- [x] Tests
- [x] Form styling
- [x] Include changes from #10 
- [x] ~~~JS validation?~~~
- [x] Validation in CommandHandler

JS validation has not been implemented yet

PR #17, opened 2017-09-29
 Scan source files for missing license headers 

jorissteyn:
The static-analysis target now scans all src and test sources for
files with missing license headers.


jorissteyn:
Cherry-picked 428e3a0  from #13 to make the new target pass.

PR #26, opened 2017-10-03
Feature/ansible install dashboard

quartje:
This PR will allow the installation of the sp dashboard application itself on different environments.


ansible/roles/spdashboard/tasks/main.yml
```diff
@@ -43,14 +43,6 @@
   notify:
     - reload httpd
 
-- name: Create SP Dashboard database
```
MKodde: This might prove problematic when setting up a development vm. The sp dashobard database was not created in the OpenConext-deploy mysql role. Thats why I've added it here explicitly. 

Would it be possible to add it to the OpenConext-deploy mysql role? I was reluctant to do so as we might not want to clutter the OpenConext-deploy repo with to many SP Dashboard configuration.

quartje: I've added the database in the group_vars of the VM in OpenConext-deploy (https://github.com/OpenConext/OpenConext-deploy/commit/c3083557f38e84ce685a509dc35f4f8596e25e62#diff-43ade091ea633bbc1614dd7508f36cbb). This ensures that the database is created.

MKodde: Thanks!

PR #50, opened 2017-10-20
Allow regular users to use the service switcher

jorissteyn:
The switcher is no longer exclusively for administrators. Instead of
showing all services in the database (administrator behaviour) the
switcher is shown when users have access to more than one services.

- [x] Drop the migration in favor of a schema drop/create solution
- [x] The Guid field for a service is no longer required
- [x] Don't forget to mention to Bart, the schema needs to be dropped & created to fix schema changes for this PR.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/AuthorizationService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/AuthorizationService.php
@@ -18,11 +18,17 @@
 namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service;
 
 use RuntimeException;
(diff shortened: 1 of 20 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/AuthorizationService.php
```diff
@@ -34,14 +40,97 @@ class AuthorizationService
...
+    /**
+     * Get the service the user is currently working on.
+     *
+     * The service can be active in two ways:
+     *
+     * - the user has access to only one service, the switcher is not shown
+     *    and the service is always "active"
+     *
+     * - the user has access to multiple services and the user selected one
+         service from the switcher
```
MKodde: There is a missing asterisk, I'll fix this for you.

PR #51, opened 2017-10-23
Update parameters template with admin team name

MKodde:
I've added the administrator_team to the parameters.yml.j2 ninja template file.


MKodde:
@quartje I've introduced an Ansible variable in `ansible/group_vars/all.yml` is this the desired location for these kind of application config parameters?

quartje:
That'll work, the value for other environments can be changed in the relevant environment variable repositories.

PR #54, opened 2017-10-30
Show error message when user has no access to service

jorissteyn:
Instead of showing a generic 'something went wrong' message, we show a
speficic error page for these two situations:

 - the user is not member of any teams (empty isMemberOf attribute)
 - the user is member of one or more teams, but no service in the
   dashboard matches any of the teams

Additionally, the templates for the 'authnfaild' and
'preconditionnotmet' exceptions were rendered using the wrong path,
resulting in a twig error. This has been fixed as well.

### DIFF
--- a/app/scss/components/service_switcher.scss
+++ b/app/scss/components/service_switcher.scss
@@ -9,3 +9,8 @@ li.service-switcher {
     display: none;
   }
 }
+
+ul.team-names {
+  list-style: inside;
+}
+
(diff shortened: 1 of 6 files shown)


MKodde:
Covering this behaviour in a web test would be a nice addition to this PR.

jorissteyn:
As discussed: this can only be tested in a system-test/end-to-end which we don't currently have.

PR #58, opened 2017-10-30
Show services from manage in the entity list

jorissteyn:
**Checklist**
- [x] Tests are written
- [x] Travis build is passing
- [x] Git history is clean
- [x] Issue in Pivotal is updated

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
@@ -70,7 +70,7 @@ class Entity
         $formattedContact = '';
 
         if ($contact) {
-            $formattedContact = self::formatContact($contact);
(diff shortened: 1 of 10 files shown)


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
```diff
@@ -82,10 +82,46 @@ public static function fromEntity(DomainEntity $entity)
...
+            'connect'
+        );
+    }
+
+    /**
+     * @return string
+     */
+    private static function formatManageContact(array $metadata)
+    {
+        for ($i=0; $i<=2; $i++) {
```
MKodde: This will work for implementation where we can save three contacts. 

In manage an additional fourth contact can be set. In theory this could result in not showing administrative contact information even though it is set.

PR #65, opened 2017-11-01
Set the MetadataUrl by default on import

MKodde:
**Review notes**
This will set the metadata url field (in the entity edit form) with the entity id. But only when the metadata url field has not been set before.

Reviewers, is this what we want? Or am I solving a non issue?

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/LoadMetadataCommandHandler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/LoadMetadataCommandHandler.php
@@ -89,9 +89,9 @@ class LoadMetadataCommandHandler implements CommandHandler
         $this->mapContacts($entity, $metadata);
         $this->mapAttributes($entity, $metadata);
 
-        // By default set the entityId as the metadataUrl but only when the metadataUrl is not set yet.
-        if (!empty($entity->getMetadataUrl())) {
-            $entity->setMetadataUrl($entity->getEntityId());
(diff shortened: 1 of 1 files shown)


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/LoadMetadataCommandHandler.php
```diff
@@ -89,6 +89,11 @@ public function handle(LoadMetadataCommand $command)
         $this->mapContacts($entity, $metadata);
         $this->mapAttributes($entity, $metadata);
 
+        // By default set the entityId as the metadataUrl but only when the metadataUrl is not set yet.
+        if (!empty($entity->getMetadataUrl())) {
```
jorissteyn: !empty -> empty

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/LoadMetadataCommandHandler.php
```diff
@@ -90,7 +90,7 @@ public function handle(LoadMetadataCommand $command)
         $this->mapAttributes($entity, $metadata);
 
         // By default set the entityId as the metadataUrl but only when the metadataUrl is not set yet.
-        if (!empty($entity->getMetadataUrl())) {
+        if (empty($entity->getMetadataUrl())) {
             $entity->setMetadataUrl($entity->getEntityId());
```
jorissteyn: Wijzigen naar import URL!

PR #93, opened 2017-12-08
Remember logo width and height - send to manage

jorissteyn:


MKodde:
I think you missed one use case. When editing an entity and you update the logo URL, shouldn't the width and height be updated? In the current changeset, this is not the case.

jorissteyn:
Actual fix for missing logo (always send width/height) is already merged. See https://www.pivotaltracker.com/story/show/153558582 for remaining issues.

PR #99, opened 2017-12-19
Export entity metadata to JSON instead of XML

jorissteyn:
Not all entity data can be expressed in SAML metadata, so we use the
manage JSON format to export entities. This simplifies publishing to
the test environment since we never generator XML. The only XML
handling left is when importing the original metadata, or importing
from manage.


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -0,0 +1,239 @@
...
+    {
+        $metadata = [
+            'active'          => true,
+            'allowedEntities' => [],
+            'allowedall'      => true,
+            'arp'             => $this->arpMetadataGenerator->build($entity),
+            'entityid'        => $entity->getEntityId(),
+            'metaDataFields'  => $this->generateMetadataFields($entity),
+            'metadataurl'     => $entity->getMetadataUrl(),
+            'state'           => 'testaccepted',
```
MKodde: Will the JSON attachement in the 'copy to prod' mail message have this state? This would be incorrect in my opinion.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
```diff
@@ -0,0 +1,71 @@
...
+            // Skip attributes we know about but don't have registered.
+            if (!method_exists($entity, $getterName)) {
+                continue;
+            }
+
+            $attr = $entity->$getterName();
+
+            if ($attr instanceof Attribute && $attr->isRequested()) {
+                $urn = reset($definition->urns);
+                $attributes[$urn] = [
```
MKodde: Maybe accompany this hardcoded section of code with an explanation.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
jorissteyn: As discussed, this is desired behaviour.

jorissteyn:
We could configure the metadata endpoint to require authentication. Not doing that until it's decided how to expose the metadata.

The missing organization fields is a bug/feature in manage, Okke will fix that -> no changes in the dashboard required.

PR #102, opened 2018-01-11
Save or publish the form when pressing enter

jorissteyn:
Browsers send the first submit button in the form when the for is
submitted using the enter key. In the case of the entity form, this is
'import'. But this is not user-friendly: when pressing enter we want
the default action to occur:

 - enter should save the form when there is a save button
 - enter should publish the form if there is only a publish button

This behaviour is implemented by adding a hidden 'default' submit
button on the top of the form.

This commit also fixes a bug where published entities could still be
saved as draft after using the import button.


jorissteyn:
https://www.pivotaltracker.com/n/projects/1400064

PR #103, opened 2018-01-11
Allow privileged users to create draft entities for production

jorissteyn:
SURF-administrators can enable service administrators to create
entities on the production environment without going to test
first.

When the checkbox 'allow production entities' is enabled, users are
allowed to create entities directly for the production environment. To
distinguish creating entities for test and production, the navigation
link 'Add entity' is removed and the 'Add' actions are shown above the
entity list.

### DIFF
--- a/app/DoctrineMigrations/Version20180110141503.php
+++ b/app/DoctrineMigrations/Version20180110141503.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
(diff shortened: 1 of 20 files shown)


jorissteyn:
https://www.pivotaltracker.com/story/show/150962957

MKodde:
When users are able to directly publish to production, you can change the environment by modifying the environment in the uri.

For example adding an entity on this uri: `https://spdashboard.dev.support.surfconext.nl/entity/create/foobar` will result in an entity with environment: `foobar`.

jorissteyn:
Fixed: environment must always be either test or production.

PR #107, opened 2018-01-16
Allow copy to production action

MKodde:
Allow copy to production action while revoking illegal production create actions.

See [Pivotal](https://www.pivotaltracker.com/story/show/154357906) story for steps to reproduce in revisions before the bugfix.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
@@ -75,7 +75,10 @@ class EntityCreateController extends Controller
 
         $service = $this->getService();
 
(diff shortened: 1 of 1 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
```diff
@@ -75,7 +75,10 @@ public function createAction(Request $request, $manageId, $environment)
 
         $service = $this->getService();
 
-        if (!$service->isProductionEntitiesEnabled() && $environment !== Entity::ENVIRONMENT_TEST) {
+        if (!$service->isProductionEntitiesEnabled() &&
+            $environment !== Entity::ENVIRONMENT_TEST &&
+            !$this->isCopyRoute($request)
```
jorissteyn: Makes sense: everyone is allowed to copy to production!

PR #112, opened 2018-04-23
Loosen UUID validation to allow GUID's

MKodde:
The Symfony UUID validator is configured very strict by default. GUIDs
from the MS CRM suite generate GUID (UUID's) not compliant to the
RFC 4122 standard.

By loosening the strictness of the validator, the MS GUID's are now
also considered valid.


jorissteyn:
Works for create-service, but missing for edit!

MKodde:
Thanks, I fixed this. Didn't add an explicit web test to cover this situation.

PR #114, opened 2018-04-23
Show XML parser errors on metadata import

MKodde:
Before this change, xml parser errors where only logged to the app log.
Leaving end users in the dark as to what went wrong during import of their metadata.

This commit changes that, adding a flash message for each parse error.

To prevent display of unwanted xml parser messages (in test), the error
buffer is cleared before validating the XML schema.

To test this feature, simply try to import a malformed SAML SP metadata document and hit the import button. The resulting error messages should help you to fix the violations in the schema.

### DIFF
--- a/app/scss/excludes/forms.scss
+++ b/app/scss/excludes/forms.scss
@@ -29,6 +29,21 @@ em, div.message {
       margin-bottom: 15px;
     }
   }
+
+  &.preformatted {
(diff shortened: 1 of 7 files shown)


MKodde:
@jorissteyn, I'm merging this to develop as I need it for another feature. Please still review this PR and any requests for change will be taken care of.

PR #120, opened 2018-04-30
Client side validation

MKodde:
Parsley is used to perform the client side validation tasks. For now parsley is only used to validat url and email fields.

A home made warning script was written to inform users to leave a detailed attribute motivation.

See [Pivotal](https://www.pivotaltracker.com/n/projects/1400064) for an overview of the required validation fields.


jorissteyn:
Works fine, one remark: URLs are currently allowed without scheme (example.com instead of https://exampe.com).

PR #122, opened 2018-05-01
Upgrade to Symfony 3.4

MKodde:
Upgrade to Symfony 3.4
 - Some effort was put into getting rid of as many deprecation notices as possible. Not all have been resolved.
 - Lexik Translation bundle was updated to support symfony 3.4

When upgrading to version 4, we might need to take a closer look at the way web tests use the DI container. As this results in the most deprecation warnings (reading of private services)

For the translation bundle to work as intended please force update the database schema:
`bin/console doctrine:schema:update --force` as an additional field was added to the `lexik_trans_unit` table.

### DIFF
--- a/tests/webtests/WebTestCase.php
+++ b/tests/webtests/WebTestCase.php
@@ -25,16 +25,15 @@ use Doctrine\ORM\EntityManager;
 use GuzzleHttp\Handler\MockHandler;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Contact;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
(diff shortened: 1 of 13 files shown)


tests/webtests/WebTestCase.php
```diff
@@ -50,10 +49,13 @@ class WebTestCase extends SymfonyWebTestCase
 
     public function setUp()
     {
+        // Disable notices, strict and deprecated warnings. Many Symfony 3.4 deprecation warnings are still to be fixed.
+        Debug::enable(E_RECOVERABLE_ERROR & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED, false);
```
jorissteyn: Let's try to enable at least E_NOTICE.

PR #128, opened 2018-05-16
Ansible change to provision multiple teams

quartje:


ansible/roles/spdashboard/templates/parameters.yml.j2
```diff
@@ -13,7 +13,10 @@ parameters:
     mail_receiver: {{ spdashboard_support_email }}
     mail_no_reply: {{ noreply_email }}
     secret: {{ spdashboard_symfony_secret }}
-    administrator_teams: {{ spdashboard_administrator_team }}
+    administrator_teams: 
+    {% for team in spdashboard.adminteams %}
```
MKodde: Small nitpick: Is there a way in Ansible to ascertain at least one admin team is provided? As this is a requirement for the application to work correctly. So maybe error out of provisioning when `spdashboard.adminteams` are empty.

quartje: I've added a default admin team to defaults/main.yml. This ensures that at least the default team is provisioned as admin team

PR #132, opened 2018-05-18
Require at least one attribute before publication

jorissteyn:
See [pivotal](https://www.pivotaltracker.com/story/show/156912587).


MKodde:
Changes to the code look allright. The display of the error message is displayed differently from the other validation messages. 

Why didn't you put the red background over the message?

screenshot-spdashboard dev support surfconext nl-2018-05-22-09-20-28

PR #134, opened 2018-05-22
Make all entity fields optional

jorissteyn:
When a value is entered, it must be valid but the form can always be
saved with empty fields. Requiredness is checked on publication.


MKodde:
As discussed, we might want to fix this problem in a less verbose way. For this we created a TD ticket in pivotal. Otherwise this fixes our problem.

PR #135, opened 2018-05-24
Format dates as RFC3339 dates

jorissteyn:
See [pivotal](https://www.pivotaltracker.com/story/show/157842831).


jorissteyn:
Tested with latest manage. Merging!

PR #136, opened 2018-05-25
This commit makes sure that Manage is used in stead of Serviceregistry

quartje:


jorissteyn:
As discussed on slack, this works fine but requires to manual actions:

 * add spdashboard.dev.support.surfconext.nl manage.dev.support.surfconext.nl to /etc/hosts
 * import dashboard metadata into manage and push to EB

jorissteyn:
Very workable, but I expected the dashboard metadata to be imported into manage automatically which is not the case (so everything works out-of-the-box). And in order to import the metadata manually I still need to add `spdashboard.dev.support.surfconext.nl` to /etc/hosts of the guest (altough I could just copy/paste the XML).

PR #137, opened 2018-05-28
Various validation improvements for the entity form

jorissteyn:
See [pivotal](https://www.pivotaltracker.com/story/show/152807745).

### DIFF
--- a/app/js/components/validation.js
+++ b/app/js/components/validation.js
@@ -107,7 +107,7 @@ window.Parsley.addValidator('urlstrict', function (value, requirement) {
 // must be empty or be a valid URN  or URL (with protocol).
 window.Parsley.addValidator('uri', function (value, requirement) {
     return validateEmpty(value) || validateUrl(value) || validateUrn(value);
-}, 32).addMessage('en', 'uri', 'This value should be a valid URI.');
+}, 32).addMessage('en', 'uri', 'This value should be a valid URL or URN.');
 
 function validateEmpty(value) {
     return value === '';
--- a/app/js/components/validation.js
+++ b/app/js/components/validation.js
@@ -25,9 +25,10 @@ if (form.length) {
         const parent = $(this).closest('.form-row');
         const errorContainer = parent.parents('.form-row').find('.parsley-errors').first();
         const translatedValidationText = $(this).data('motivation-keep-talking');
-        const words = $(this).val().split(' ').filter(v => v !== "");
+        const val = $(this).val();
+        const words = val.split(' ').filter(v => v !== "");
 
-        if (words.length < 10) {
+        if (val !== '' && words.length < 10) {
             errorContainer.parent('.form-row').addClass('warning');
             errorContainer.text(translatedValidationText);
         } else {
--- a/app/js/components/validation.js
+++ b/app/js/components/validation.js
@@ -25,8 +25,9 @@ if (form.length) {
         const parent = $(this).closest('.form-row');
         const errorContainer = parent.parents('.form-row').find('.parsley-errors').first();
         const translatedValidationText = $(this).data('motivation-keep-talking');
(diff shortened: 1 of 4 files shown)


app/js/components/validation.js
```diff
@@ -25,8 +25,9 @@ if (form.length) {
         const parent = $(this).closest('.form-row');
         const errorContainer = parent.parents('.form-row').find('.parsley-errors').first();
         const translatedValidationText = $(this).data('motivation-keep-talking');
+        const words = $(this).val().split(' ').filter(v => v !== "");
```
MKodde: The filter is a nice and elegant way of filtering away empty entries. Using `v` as the name for the function input is not very descriptive. Using `word` instead of `v` might be a bit more self-explaining.

app/js/components/validation.js
```diff
@@ -25,9 +25,10 @@ if (form.length) {
         const parent = $(this).closest('.form-row');
         const errorContainer = parent.parents('.form-row').find('.parsley-errors').first();
         const translatedValidationText = $(this).data('motivation-keep-talking');
-        const words = $(this).val().split(' ').filter(v => v !== "");
+        const val = $(this).val();
+        const words = val.split(' ').filter(v => v !== "");
 
-        if (words.length < 10) {
+        if (val !== '' && words.length < 10) {
```
MKodde: Just nitpicking here.. Does jQuery val also trim the input of whitespaces? Using `words.length > 0 && words.length < 10` might be a bit better expression?

MKodde:
Codewise this looks kind of good to me. As we discussed a couple of days ago, the validation.js file starts to become more of a maintenance problem. 

Would it be an idea to register this as a technical debt issue in Pivotal? One of my first thoughts is to utilize TypeScript, enabling us to restructure this more easily OO style.

app/js/components/validation.js
```diff
@@ -85,8 +85,30 @@ if (form.length) {
...
+// allowed to be empty, or a URL with protocol.
 window.Parsley.addValidator('urlstrict', function (value, requirement) {
+    return validateEmpty(value) || validateUrl(value);
+}, 32).addMessage('en', 'urlstrict', 'This value should be a valid URL.');
+
+// Add URI validator (must be URN or URL), fields validated with urn
+// must be empty or be a valid URN  or URL (with protocol).
+window.Parsley.addValidator('uri', function (value, requirement) {
+    return validateEmpty(value) || validateUrl(value) || validateUrn(value);
+}, 32).addMessage('en', 'uri', 'This value should be a valid URI.');
```
MKodde: My opinion is that using URI in this context is possibly confusing as you are not allowed to fill any URI but just URL's and URN's. 

For example, I'm not allowed to fill this valid URI into the Entity ID field: `mailto:someone@example.com`

MKodde: My suggestion is to substitute uri with urn.

app/js/components/validation.js
```diff
@@ -85,8 +85,30 @@ if (form.length) {
     });
 }
 
-// Add a stricter url validator, this one ensures the protocol is set on the input value
+// Add a stricter url validator, fields validated with urlstrict are
+// allowed to be empty, or a URL with protocol.
 window.Parsley.addValidator('urlstrict', function (value, requirement) {
+    return validateEmpty(value) || validateUrl(value);
+}, 32).addMessage('en', 'urlstrict', 'This value should be a valid URL.');
```
MKodde: The validation message for the URL is probably good enough. When looking at it in a very puristic manner. The message implies I should be able to fill all sorts of URL's like: `file:///home/someuser/somefile.txt` or `ftp://foo@example.com`. Adding a hint on the allowed schemes (http/https) would take away this very unlikely confusion.

PR #140, opened 2018-06-06
Allow basic HTTP authentication for SP dashboard vhost

jorissteyn:
PHP under fcgi doesn't parse the Authorization header into PHP_AUTH_*
$_SERVER variables. Symfony requires this to work. This commit changes
the vhost configuration to copy the contents of Authorization into an
environment variable HTTP_AUTHORIZATION.

EngineBlock API already uses this rule in its configuration.


quartje:
Looks good to me!

PR #145, opened 2018-11-05
Set curl timeout on logo downloads

MKodde:
The timeout is now set at ten seconds for both the connect timeout
and the entire running time of the curl request.

:warning: The build fails on the security checker. The recommended updates will be installed in another PR. Either merge this PR with the failing build or rebase onto afore mentioned PR.

See: https://www.pivotaltracker.com/story/show/158383467
See: https://github.com/SURFnet/sp-dashboard/pull/147 (composer update)


MKodde:
@quartje if this problem was still present. You could now bring `develop` to test, and check if this timeout works out for us.

PR #149, opened 2018-11-06
Show the service status options on add and edit forms

MKodde:
First, merge #148 and rebase this branch onto develop.

:stop_sign: Do not yet run the migration or provide a means to roll back. As the migration will be squashed in a successive PR. One field will be removed (privacy_questions_anwered) and default values will be set.


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/CreateServiceCommandHandler.php
```diff
@@ -51,6 +51,12 @@ public function handle(CreateServiceCommand $command)
         $service->setTeamName($command->getTeamName());
         $service->setProductionEntitiesEnabled($command->isProductionEntitiesEnabled());
         $service->setPrivacyQuestionsEnabled($command->isPrivacyQuestionsEnabled());
+        $service->setServiceType($command->getServiceType());
```
pablothedude: If all these values are set in the constructor I think it will  make the code better and less prone to errors. I don't think it's an issue now but if we're tackling technical debt we could consider to refactor this.

MKodde: :+1:

MKodde:
Merged PR's 153 through 150 into this branch. :crossed_fingers: for the travis build to be green :)

PR #150, opened 2018-11-07
Disable privacy questions service status

MKodde:
:warning: Review and merge #149 first!

The privacy questions service status radio group is disabled when
privacy questions are not enabled for the service.

### DIFF
--- a/app/js/components/service_status_interactions.js
+++ b/app/js/components/service_status_interactions.js
@@ -1,0 +1,17 @@
+'use strict';
+
+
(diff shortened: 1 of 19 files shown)


app/js/components/service_status_interactions.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
pablothedude: Why mot use typescript?

MKodde: It has not yet been made available in the project. When it is, this little routine might be a good candidate for a first proof of concept.

PR #153, opened 2018-11-07
 Verify privacy questions are displayed correctly 

MKodde:
First review #148 .. #152 
This adds a web test to the test suite that verifies the correct determination of the privacy questions filled service status option.


pablothedude:
If I save the privacy questions without answering a question the privacy questions show up as answered. Is that expected behavior?

PR #154, opened 2018-11-08
Provide TypeScript support in the project

MKodde:
In addition to just adding typescript support, the following other changes have been made:

- Ansible provisioning was adjusted slightly (node version pinning was added) @quartje would you be so kind to test this part of the PR? 903fea8
- Webpack encore was updated to the latest version (also to get webpack 4)
- Typescript was installed along with other QA tooling to support TypeScript developing
    - TypeScript
    - Jest (testing framework)
    - Static analysis tools (tslint, stylelint, cpd)
- Build.xml was updated to make sure the new QA tooling is also part of the build process.


ansible/roles/spdashboard/tasks/main.yml
```diff
@@ -12,6 +12,14 @@
     - npm 
     - yarn
 
+- name: Install n in order to upgrade node
+  command: yarn global add n --prefix /usr/local
+  become: yes
```
quartje: become: yes is not needed in the role. That has already been taken care of in the playbook

MKodde: My understanding is that adding become: yes runs the operation with super user privileges. But this is not the case? Or is the entire role run with sudo?

quartje: The entire role is run with sudo, see https://github.com/SURFnet/sp-dashboard/blob/develop/ansible/vagrant.yml#L16

MKodde: :eyes:

MKodde:
Thanks @pablothedude for the for the scss fix!

PR #156, opened 2018-11-09
Add contract signed toggle based on service type

pablothedude:
The contract signed field should only be filled when a service is
for an institution.


app/js/components/service_form.test.js
```diff
@@ -0,0 +1,39 @@
+jest
+    .dontMock('fs')
+    .dontMock('jquery');
+
+
+window.$ = require('jquery');
+
+let html = require('fs').readFileSync('./app/js/components/mock/serviceForm.html').toString();
```
MKodde: Cool solution for integration testing. My only concern with this kind of test is that they are very fragile. If the form changes in the slightest, the test becomes unpredictable/unstable. 

Some thoughts, and feel free to discuss them IRL

* Can we further abstract the DOM from these tests? In a way that only mockable elements are used in the javascript and we only have to provide mock objects for these elements?
* Do we need to update our test plan? 
* Do we need to update our DoD with a trigger to update test files when certain changes have been applied?

app/js/components/service_form.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+let serviceForm = new function($) {
```
MKodde: Maybe rebase onto develop and reap the power of TypeScript. This might improve testability too?

app/js/components/service_form.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Disable the privacy questions answered field when the privact question enabled checkbox is unchecked
+     */
+    function initTogglePrivacyQuestionsAnsweredStatusField()
+    {
+        let togglePrivacyQuestions = function () {
+            if ($privacyQuestionsEnabledToggle.is(':checked')) {
+                $privacyQuestionsStatusContainer.parent().show();
```
MKodde: You removed the show animation (fadein 200ms). Is this something we really want? In my opinion, some animation is beneficial for a good user experience.

app/js/components/service_form.test.js
pablothedude: I think for now this solution will be sufficient. In order to have a more predictable/stable outcome we should consider js end-to-end testing but for now in my opinion that's too much work. I agree that it's better to use a more component based approach with rigorous  unit testing for the different components. But I think for now we're good off as is. Because it will be very easy to copy-paste the form from a browser.

I think if you're testing something you always have a commitment in keeping the mocks/fixtures/tests etc. updated. Should you then define that in your DOD or testplan? I don't mind however if we put it there. What's your opinion?

app/js/components/service_form.js
pablothedude: I removed it to be able to test the logic without waiting till the animation has ended. I've now added a class, so I could check for that `class` in the tests.

app/js/components/service_form.js
pablothedude: I've unleashed the power of Typescript...

app/js/components/service_form.test.js
MKodde: Yes, the test shouldn't be too hard to maintain. I'm fine with this!

app/js/components/mock/serviceCreateForm.html
```diff
@@ -0,0 +1,74 @@
+<form name="dashboard_bundle_service_type" method="post" novalidate="novalidate">
```
MKodde: The naming of the mock files should in my opinion also be in snake case. Also goes for the 'serviceEdit' counterpart.

app/js/components/service_form.test.js
```diff
@@ -0,0 +1,74 @@
...
+    .dontMock('fs')
+    .dontMock('jquery');
+
+
+window.$ = require('jquery');
+
+
+const loadServiceForm = require('./service_form.ts');
+
+describe('validateToggles serviceEditForm', function() {
```
MKodde: This description doesn't sit well with me. Or is this Jest convention?

Suggestion:
`validate visibility toggling of service status fields on the service edit form`

app/js/components/service_form.ts
```diff
@@ -0,0 +1,104 @@
...
+            $('.contract-signed-toggle'),
+            $('.contract-signed-container'),
+        );
+        serviceForm.registerEventHandlers();
+    }
+}
+
+$(document).ready(loadServiceForm);
+
+module.exports = loadServiceForm;
```
MKodde: nitpick: CRLF

app/js/components/service_form.ts
```diff
@@ -0,0 +1,104 @@
...
+        $element.hide(200);
+
+    }
+
+    /**
+     * Helper method to show an element
+     */
+    private showElement($element: JQuery) {
+        $element.removeClass('hidden');
+        $element.show(200);
```
MKodde: I thought jQuery animations where dropped in favor of CSS transitions?

app/js/components/mock/serviceCreateForm.html
pablothedude: The naming of fixture files doesn't bother me that much, but I've changed it as suggested.

app/js/components/service_form.test.js
pablothedude: Thanks for the good suggestion, I've changed it accordingly.

app/js/components/service_form.ts
pablothedude: Added

app/js/components/service_form.ts
pablothedude: I've looked for a pure css solution but you would require a `hidden` and a `show` state and for each state a `enter` and a `leave`. Because if you add a `display: none` there is no animation. I however didn't want to be dependent of the animations for testing so I've added a `hidden` class to the element to test the state for testing. I've also removed the css for the `hidden` class, because that isn't used anymore.

MKodde:
Rebased the changes onto develop, and upgraded the yarn.lock file.

PR #158, opened 2018-11-12
Prevent ROLE_USER from using service switcher

MKodde:
This role will be using the more user friendly overview page that is
yet to be created. This change rolls back the previous adjustment where
contacts with multiple services leveraged the admin service switcher.

See: https://www.pivotaltracker.com/story/show/161719597
See: Task 1 of #161719597

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/ServiceSwitcherExtension.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/ServiceSwitcherExtension.php
@@ -58,7 +58,7 @@ class ServiceSwitcherExtension extends Twig_Extension
     public function render(Twig_Environment $environment)
     {
         $token = $this->tokenStorage->getToken();
-        if (!$token) {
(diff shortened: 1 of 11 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/ServiceSwitcherExtension.php
```diff
@@ -58,7 +58,7 @@ public function getFunctions()
     public function render(Twig_Environment $environment)
     {
         $token = $this->tokenStorage->getToken();
-        if (!$token) {
+        if (!$token || !$token->hasRole('ROLE_ADMINISTRATOR')) {
```
pablothedude: I think there shouldn't be authentication logic in templates. I however don't mind if you don't fix it now.

MKodde: I think there should be and even has to be in order to show/hide functionality from certain roles. This would be quite tedious to fix on a higher level. It is a very interesting thought tho! One to consider more often.

MKodde:
Merging #160 into this branch

PR #160, opened 2018-11-12
Introduction of the services overview page

MKodde:
Some decisions have been made which might need adjustment later on:
1. The service names have been made links to enable us to switch to that service (to get to their entities & privacy questions)
2. The service overview has been made the default route at (/) And the previous default (entities overview) is now available at /entities.
3. Service entities are fed into the twig template. A more template friendly object might be required at some point.

:warning: This PR is built on top of #158

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
@@ -188,14 +188,15 @@ class ServiceController extends Controller
     }
 
     /**
-     * @Method("POST")
+     * @Method({"GET", "POST"})
      * @Route("/service/select", name="select_service")
-     * @Security("has_role('ROLE_ADMINISTRATOR')")
+     * @Security("has_role('ROLE_USER')")
      */
     public function selectAction(Request $request)
     {
+        $serviceId = $request->get('service', $request->query->get('service'));
         $command = new SelectServiceCommand(
-            $request->request->get('service')
+            $serviceId
         );
 
         $this->commandBus->handle($command);
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
@@ -87,10 +87,10 @@ class ServiceController extends Controller
      */
     public function overviewAction()
     {
-        $serviceOptions = $this->authorizationService->getAllowedServiceNamesById();
-        $services = $this->serviceService->getServicesByCurrentServiceIds($serviceOptions);
+        $allowedServices = $this->authorizationService->getAllowedServiceNamesById();
+        $services = $this->serviceService->getServicesByCurrentServiceIds($allowedServices);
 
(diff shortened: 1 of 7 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -80,6 +80,25 @@ public function __construct(
...
 
+    /**
+     * @Method({"GET"})
+     * @Route("/", name="service_overview")
+     * @Template()
+     */
+    public function overviewAction()
+    {
+        $serviceOptions = $this->authorizationService->getAllowedServiceNamesById();
+        $services = $this->serviceService->getServicesByCurrentServiceIds($serviceOptions);
```
pablothedude: Isn't the array returned from the `getAllowedServiceNamesById` not an array with names as suggested with the method name? Maybe it's better to use a valueobject or collection to return the required services.

tests/webtests/ServiceOverviewTest.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public function test_users_can_use_the_service_overview_page()
+    {
+        $serviceRepository = $this->getServiceRepository();
+        $surfNet = $serviceRepository->findByName('SURFnet');
+        $this->logIn('ROLE_USER', [$surfNet]);
+
+        $crawler = $this->client->request('GET', '/');
+
+        $nodes = $crawler->filter('.card h1');
```
pablothedude: I personally find it hard to tell what's happening in this code block? I think it would be very valuable to add a small comment.

tests/webtests/ServiceOverviewTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $serviceRepository = $this->getServiceRepository();
+        $surfNet = $serviceRepository->findByName('SURFnet');
+        $ibuildings = $serviceRepository->findByName('Ibuildings B.V.');
+        $this->logIn('ROLE_USER', [$surfNet, $ibuildings]);
+
+        $crawler = $this->client->request('GET', '/');
+
+        $nodes = $crawler->filter('.card h1');
+
+        $this->assertEquals(2, $nodes->count());
```
pablothedude: I personally find it hard to tell what's happening in this code block? I think it would be very valuable to add a small comment.

tests/webtests/ServiceOverviewTest.php
MKodde: See ca4ed06

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
MKodde: See ca4ed06

tests/webtests/ServiceOverviewTest.php
MKodde: See ca4ed06

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
MKodde: Yes this is a just comment! Let me improve this.

PR #161, opened 2018-11-12
Add an endpoint to retrieve the service states

pablothedude:
In order to have the service states visualized we need an api
endpoint to retrieve the data so we could represent it to the
user in a fancy way.

https://www.pivotaltracker.com/story/show/161719597

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
@@ -1,0 +1,160 @@
+<?php
+/**
+ * Copyright 2018 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\Assembler;
+
+use Surfnet\ServiceProviderDashboard\Application\Dto\ServiceStatusDto;
+use Surfnet\ServiceProviderDashboard\Application\Service\ServiceStatusService;
+use Surfnet\ServiceProviderDashboard\Application\ViewObject\EntityList;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
+
+class ServiceStatusAssembler
+{
+    /**
+     * The different status types used
+     */
+    const SERVICE_STATE_INTAKE_CONDUCTED = 'intake-conducted';
+    const SERVICE_STATE_ENTITY_ON_TEST = 'entity-on-test';
+    const SERVICE_STATE_REPRESENTATIVE_APPROVED = 'representative-approved';
+    const SERVICE_STATE_CONTRACT_SIGNED = 'contract-signed';
+    const SERVICE_STATE_PRIVACY_QUESTIONS = 'privacy-questions';
+    const SERVICE_STATE_PRODUCTION_CONNECTION = 'production-connection';
+
+
+    /**
+     * The possible states used in the mapping
+     */
+    const SERVICE_STATUS_SUCCESS = 'success';
+    const SERVICE_STATUS_INFO = 'info';
+    const SERVICE_STATUS_WARNING = 'warning';
+    const SERVICE_STATUS_DANGER = 'danger';
+
+    /**
+     * The mapping of the different states
+     *
+     * @var array
+     */
+    private $stateMapping = [
+        self::SERVICE_STATE_INTAKE_CONDUCTED => [
+            Service::INTAKE_STATUS_YES => self::SERVICE_STATUS_SUCCESS,
+            Service::INTAKE_STATUS_NO => self::SERVICE_STATUS_DANGER,
+        ],
+        self::SERVICE_STATE_ENTITY_ON_TEST => [
+            Service::ENTITY_PUBLISHED_YES => self::SERVICE_STATUS_SUCCESS,
+            Service::ENTITY_PUBLISHED_NO => self::SERVICE_STATUS_DANGER,
+            Service::ENTITY_PUBLISHED_IN_PROGRESS => self::SERVICE_STATUS_WARNING,
+        ],
+        self::SERVICE_STATE_REPRESENTATIVE_APPROVED => [
+            Service::SURFCONEXT_APPROVED_YES => self::SERVICE_STATUS_SUCCESS,
+            Service::SURFCONEXT_APPROVED_NO => self::SERVICE_STATUS_DANGER,
+        ],
+        self::SERVICE_STATE_CONTRACT_SIGNED => [
+            Service::CONTRACT_SIGNED_YES => self::SERVICE_STATUS_SUCCESS,
+            Service::CONTRACT_SIGNED_NO => self::SERVICE_STATUS_DANGER,
+        ],
+        self::SERVICE_STATE_PRIVACY_QUESTIONS => [
+            true => self::SERVICE_STATUS_SUCCESS,
+            false => self::SERVICE_STATUS_DANGER,
+        ],
+        self::SERVICE_STATE_PRODUCTION_CONNECTION => [
+            Service::CONNECTION_STATUS_NOT_REQUESTED => self::SERVICE_STATUS_DANGER,
+            Service::CONNECTION_STATUS_REQUESTED => self::SERVICE_STATUS_INFO,
+            Service::CONNECTION_STATUS_SURFCONEXT_INFORMED => self::SERVICE_STATUS_INFO,
+            Service::CONNECTION_STATUS_ACTIVE => self::SERVICE_STATUS_SUCCESS,
+        ],
+    ];
+
+    /**
+     * @var ServiceStatusDto
+     */
+    private $serviceStatusDto;
+
+    /**
+     * ServiceStatusAssembler constructor.
+     * @param Service $service
+     * @param string $serviceLink
+     * @param ServiceStatusService $serviceStatusService
+     * @param EntityList $entityList
+     */
+    public function __construct(Service $service, $serviceLink, ServiceStatusService $serviceStatusService, EntityList $entityList)
+    {
+        $states = $this->getStates($service, $serviceStatusService);
+        $mappedStates = $this->mapStates($states);
+
+        $this->serviceStatusDto = new ServiceStatusDto(
+            $service->getName(),
+            $serviceLink,
+            $entityList,
+            $mappedStates
+        );
+    }
+
+    /**
+     * @return ServiceStatusDto
+     */
+    public function getDto()
+    {
+        return $this->serviceStatusDto;
+    }
+
+    /**
+     * @param Service $service
+     * @param ServiceStatusService $serviceStatusService
+     * @return array
+     */
+    private function getStates(Service $service, ServiceStatusService $serviceStatusService)
+    {
+        $states = [];
+        $type = $service->getServiceType();
+
+        $intakeStatus = $service->getIntakeStatus();
+        if ($intakeStatus != Service::INTAKE_STATUS_NOT_APPLICABLE) {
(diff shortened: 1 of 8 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -0,0 +1,195 @@
...
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\Assembler;
+
+
+use Surfnet\ServiceProviderDashboard\Application\Dto\ServiceStatusDto;
+use Surfnet\ServiceProviderDashboard\Application\Service\ServiceStatusService;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
+use Symfony\Component\Routing\RouterInterface;
```
MKodde: You certainly are not the first to do this, but we should aim to not couple to the infrastructure layer in the application layer. Without going overboard, can you think of a way to remove this coupling?

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -0,0 +1,195 @@
...
+{
+    /**
+     * The different status types used
+     */
+    const SERVICE_STATE_INTAKE_CONDUCTED = 'intake-conducted';
+    const SERVICE_STATE_ENTITY_ON_TEST = 'entity-on-test';
+    const SERVICE_STATE_REPRESENTATIVE_APPROVED = 'representative-approved';
+    const SERVICE_STATE_CONTRACT_SIGNED = 'contract-signed';
+    const SERVICE_STATE_PRIVACY_QUESTIONS = 'privacy-questions';
+    const SERVICE_STATE_PRODUCTION_CONNECTION = 'production-connection';
```
MKodde: In retrospect to my own naming in previous commits. This state could have been named:

`entity on production` to be more in line with the entity on test status. It's up to you what you want to do with this.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
MKodde: For example you can probably create the routes in the Twig template.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        if (isset($this->stateMapping[$name])) {
+            if (isset($this->stateMapping[$name][$value])) {
+                return $this->stateMapping[$name][$value];
+            }
+        }
+
+        throw new \InvalidArgumentException(sprintf('invalid status requested: %s %s', $name, $value));
+    }
+}
```
MKodde: Nitpick: CRLF

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet B.V.
```
MKodde: ~~2017~~ 2018

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
```diff
@@ -0,0 +1,117 @@
...
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Controller\Api;
+
+use League\Tactician\CommandBus;
+use Psr\Log\LoggerInterface;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
+use Surfnet\ServiceProviderDashboard\Application\Assembler\ServiceStatusAssembler;
+use Surfnet\ServiceProviderDashboard\Application\Command\Service\CreateServiceCommand;
```
MKodde: 8 unused imports can be removed

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @Method({"GET", "POST"})
+     * @Route("/api/service/status", name="api_service_status")
+     * @Security("has_role('ROLE_ADMINISTRATOR')")
+     * @Template()
+     *
+     * @param Request $request
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Symfony\Component\HttpFoundation\Response
+     */
+    public function statusAction(Request $request)
```
MKodde: Request isn't used

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+        $service = $this->serviceService->getServiceById(
```
MKodde: The idea is that a user (known as a Contact) can have 1..n Services. The services of the user can be retrieved from the authorizationservice::getAllowedServiceNamesById.

Do we want to make this a singular get endpoint where you pass an identifier of the service?
Or do we want to search all available services for the user and return them all at once?

My personal preference would be the last option.

tests/unit/Application/Assembler/ServiceStatusAssemblerTest.php
```diff
@@ -0,0 +1,265 @@
...
+                    'type' => Service::SERVICE_TYPE_INSTITUTE,
+                    'intakeStatus' => Service::INTAKE_STATUS_YES,
+                    'representativeApproved' => Service::SURFCONEXT_APPROVED_YES,
+                    'contractSigned' => null,
+                    'isPrivacyQuestionsEnabled' => true,
+                    'connectionStatus' => Service::CONNECTION_STATUS_ACTIVE,
+                    'entities' => [],
+                ],
+                'statusService' => [
+                    'entityStatus' => Service::ENTITY_PUBLISHED_YES,
```
MKodde: Maybe add an entity for good measure? Or set the entity status to no. Just for the sake of correctness.

tests/unit/Application/Assembler/ServiceStatusAssemblerTest.php
```diff
@@ -0,0 +1,265 @@
...
+use Mockery as m;
+use Mockery\Adapter\Phpunit\MockeryTestCase;
+use Mockery\MockInterface;
+use Surfnet\ServiceProviderDashboard\Application\Assembler\ServiceStatusAssembler;
+use Surfnet\ServiceProviderDashboard\Application\Service\ServiceStatusService;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity;
+use Symfony\Component\Routing\RouterInterface;
+
+class ServiceStatusAssemblerTest extends MockeryTestCase
```
MKodde: An idea for an extra test: the mapState method of the assembler can throw an InvalidArgumentException. This behavior is currently untested.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    private function mapState($name, $value)
+    {
+        if (isset($this->stateMapping[$name])) {
+            if (isset($this->stateMapping[$name][$value])) {
+                return $this->stateMapping[$name][$value];
+            }
+        }
+
+        throw new \InvalidArgumentException(sprintf('invalid status requested: %s %s', $name, $value));
```
MKodde: Please use or create an exception from the application namespace. And secondly in this project we alway import all classnames. I've seen other instance in this PR where you use them without importing them first. Please fix this.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * @param Service $service
+     * @param RouterInterface $router
+     * @return array
+     */
+    private function getEntities(Service $service, RouterInterface $router)
+    {
+        $entityLinks = [];
```
MKodde: The entities need to be grouped on their connection status (production / test).
Idea: make an entity VO and collect them in this DTO

MKodde:
Hi @pablothedude, I've addressed some of the more minor feedback for you, and rebased this branch onto head of develop. 

I'd appreciate it if you do the rest of the feedback resolving. Because it's up to you how you'd like to address my feedback.

tests/unit/Application/Assembler/ServiceStatusAssemblerTest.php
pablothedude: Removed the exception, this should never occur

tests/unit/Application/Assembler/ServiceStatusAssemblerTest.php
pablothedude: Added an entity

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
pablothedude: I think it would not be an issue to determine this when building the 'donut'. If it's changed this would be a very small change.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/Api/ServiceController.php
pablothedude: Done

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
pablothedude: I've left it this way

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
pablothedude: Moved the logic to the service

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
pablothedude: Used the EntityList VO to accomplish this and grouped them on environment.

pablothedude:
I've  addressed the outstanding issues.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
MKodde: Why not utilize the EntityList itself and stop keeping a entityLinks in the DTO? You could add a method to the entity list to retrieve prod/test entities.

pablothedude: I've removed the separation and only return a list with VO's with the environment as parameter, then it's up to the frontend how to display the data.

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
```diff
@@ -187,4 +196,12 @@ public function getEnvironment()
     {
         return $this->environment;
     }
+
+    /**
+     * @return string
+     */
+    public function getLink()
+    {
+        return $this->router->generate('entity_edit', ['id' => $this->getId()]);
```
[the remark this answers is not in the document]
pablothedude: If this should work differently we could change this later on and for example introduce a decorator

pablothedude:
Regarding the separate commits, my intention was in the end to squash all commits in a single commit. Because there was no real business logic involved and this change only addressed presentation logic I found that a single commit would be more useful on the longterm.

Now I've squashed them all into a single commit.

PR #162, opened 2018-11-13
Make sure migrations can always be run

MKodde:
The current setup did not allow us to run migrations when starting with a new project. The solution used to be to do a schema create and then tell doctrine that all migrations have been executed. 

To make this a bit more user-friendly, the initial schema was put into a migration.

For installation details see [Pivotal](https://www.pivotaltracker.com/story/show/161877201/comments/196410528)


app/DoctrineMigrations/Version20180101010000.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf(
+            $this->connection->getDatabasePlatform()->getName() !== 'mysql',
+            'Migration can only be executed safely on \'mysql\'.'
+        );
+
```
pablothedude: Maybe it's an idea to cleanup the migrations table to prevent warnings when running the doctrine commands.
   ```
     // If this migration is run on an existing DB just clear the version table and assume we are up-to-date
        if (count($this->version->getConfiguration()->getMigratedVersions())) {
            $this->addSql("TRUNCATE migration_versions");
            return;
        }
```

MKodde: A good suggestion and thanks for the discussion IRL.

PR #166, opened 2018-11-14
Read production entities from Manage

MKodde:


MKodde:
The four last commits where required to get the build to pass. The yarn upgrade and Travis log catting can be dropped from this PR if you think it is out of line! But my personal feeling is that they are somewhat useful.

pablothedude:
Good test coverage and thanks for testing the Entity VO methods.

1. If I edit a 'production/requested' entity and publish it I see the following message: `Successfully published the entity to test` is that correct?
2. If I copy a 'published/production' entity and copy it, it still has the state `production/published` is that correct?

MKodde:
@pablothedude 
1. That was an error. The wrong published endpoint was visited after publishing to production. This is fixed in 73d91ed
2. Yes it is, see the pivotal story.
3. I've found another nasty issue myself: copy to production did not consolidate with the correct manage instance. This is fixed now.

MKodde:
The copy command handler is a number one candidate to be refactored into smaller command handlers. The distinction between test and production copy actions should be considered here.

PR #169, opened 2018-11-20
Display service form in two columns

MKodde:
The first column contains the 'general' form fields like the name,
team, crm id and some config flags. The second column is filled
with the service status radio groups.

This improves the form visually and makes it easier to perform
tasks on.

In addition to this change, the global setup of the form was updated
sligthly. The functionality that was reused by the edit and create
forms has been abstracted into a form type of its own. This typ is
then extended by the CreateServiceType and EditServiceType.


pablothedude:
Personally I don't think the two column setup is a major improvement. If the column on the left is the full width and the column on the rights is separated over two columns I think it would be better. But that's a personal preference and don't want to reject this PR on a personal preference.

Or if it's easier, swapping the columns would be a better. But that is also a personal preference.

pablothedude:
I don't like it that we lost the git history of the editservice- and createservice- types.

MKodde:
Regarding your first comment, that might work! Feel free to add this as a suggestion on the ticket!
And the second comment: Yes that was the consequence of this change. And we did not actually lose any history. It was obscured by moving the files to the correct namespace.

PR #173, opened 2018-11-26
Enable deleting of services

MKodde:
Delete a service and its entities.

Note that requesting the deletion of a published production entity is yet to be built. And awaits details in regards to the required Jira workflow.

For functional details, see the pivotal story.
https://www.pivotaltracker.com/story/show/154841957

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
@@ -18,10 +18,14 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\CommandHandler\Service;
 
+use Exception;
+use League\Tactician\CommandBus;
 use Psr\Log\LoggerInterface;
+use Surfnet\ServiceProviderDashboard\Application\Command\Entity\DeleteCommandFactory;
 use Surfnet\ServiceProviderDashboard\Application\Command\Service\DeleteServiceCommand;
 use Surfnet\ServiceProviderDashboard\Application\CommandHandler\CommandHandler;
 use Surfnet\ServiceProviderDashboard\Application\Service\EntityServiceInterface;
+use Surfnet\ServiceProviderDashboard\Application\ViewObject\EntityList;
 use Surfnet\ServiceProviderDashboard\Domain\Repository\ServiceRepository;
 
 class DeleteServiceCommandHandler implements CommandHandler
@@ -37,6 +41,16 @@ class DeleteServiceCommandHandler implements CommandHandler
     private $entityService;
 
     /**
+     * @var DeleteCommandFactory
+     */
+    private $deleteCommandFactory;
+
+    /**
+     * @var CommandBus
+     */
+    private $commandBus;
+
+    /**
      * @var LoggerInterface
      */
     private $logger;
@@ -44,10 +58,14 @@ class DeleteServiceCommandHandler implements CommandHandler
     public function __construct(
         ServiceRepository $serviceRepository,
         EntityServiceInterface $entityService,
+        DeleteCommandFactory $deleteCommandFactory,
+        CommandBus $commandBus,
         LoggerInterface $logger
     ) {
         $this->serviceRepository = $serviceRepository;
         $this->entityService = $entityService;
+        $this->deleteCommandFactory = $deleteCommandFactory;
+        $this->commandBus = $commandBus;
         $this->logger = $logger;
     }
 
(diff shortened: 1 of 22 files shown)


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
```diff
@@ -0,0 +1,118 @@
...
+     */
+    public function handle(DeleteServiceCommand $command)
+    {
+        $serviceId = $command->getId();
+        $service = $this->serviceRepository->findById($serviceId);
+
+        $this->logger->info(sprintf('Removing "%s" and all its entities.', $service->getName()));
+
+        // Remove the entities of the service
+        $entityList = $this->entityService->getEntityListForService($service);
```
pablothedude: Maybe a bit odd to use a ViewObject to not present data. Maybe that's because of the naming of the list.

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
```diff
@@ -0,0 +1,118 @@
...
+
+        // Finally delete the service
+        $this->serviceRepository->delete($service);
+    }
+
+    /**
+     * Using the deleteCommandFactory, entity delete commands are created
+     * that will remove them from the appropriate environment.
+     */
+    private function removeEntitiesFrom(EntityList $entityList)
```
pablothedude: Param typehint missing.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Repository/ServiceRepository.php
```diff
@@ -119,4 +119,10 @@ public function findByTeamNames($teamNames)
             'teamName' => $teamNames,
         ]);
     }
+
+    public function delete(Service $service)
```
pablothedude: Missing typehint

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
```diff
@@ -0,0 +1,118 @@
...
+    {
+        foreach ($entityList->getEntities() as $entity) {
+            try {
+                $command = $this->deleteCommandFactory->from($entity);
+                $this->commandBus->handle($command);
```
pablothedude: I like the use of CommandHandlers very much because of it's clear structure. To inject a new Command from a CommandHandler into the bus again seems a bad practice because you don't expect at all that if you're working on the EntityDelete commands/handlers that they are initiated here.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Repository/ServiceRepository.php
MKodde: The type hint is present, I did not add a doc block on purpose as the type hint is explicit enough in my opinion.

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
MKodde: Yes, see my feedback on this below. The @param definition is not really adding any value. As the type hint is perfectly explicit.

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
MKodde: We discussed this IRL. I'm leaving it as is. The conclusion was: It would be best if the CommandHandlers would only perform tasks on a high level and use services to delegate the modifications on the domain. Now the command handlers are Jack of all trades (controller pattern ;) ) and it might be a good thing to refactor this in a future sprint.

I created a TD ticket to fix this in the future.

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Service/DeleteServiceCommandHandler.php
MKodde: Maybe the name of that object no longer reflects its intent and purpose. Maybe we should have decided at an earlier state to separate them into individual VO's.

MKodde:
Also from this review, I've created yet another [TD ticket](https://www.pivotaltracker.com/story/show/162298797) that I'll be addressing next, in a new PR.

Merging these changes!

PR #174, opened 2018-11-27
Add the entity type dialog screen

MKodde:
In this dialog, the user can select the option to add either a SAML2 or OpenID connect entity.

The scope of this PR is to add the screen and integrate it into the current workflow untill the rendering of the entity create form. This will be taken care of in an upcoming PR.

Details: https://www.pivotaltracker.com/story/show/161720049

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
@@ -48,6 +48,9 @@ class Entity
     const STATE_DRAFT = 'draft';
     const STATE_PUBLISHED = 'published';
 
+    const TYPE_SAML = 'saml20';
+    const TYPE_OPENID_CONNECT = 'oidc';
+
     /**
      * @var string
      *
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Command/Entity/ChooseEntityTypeCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Command/Entity/ChooseEntityTypeCommand.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Copyright 2018 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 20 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Command/Entity/ChooseEntityTypeCommand.php
```diff
@@ -0,0 +1,49 @@
...
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity;
+
+class ChooseEntityTypeCommand
+{
+    /**
+     * @var string
+     */
+    private $type;
+
+    public function __construct($type = Entity::TYPE_SAML)
```
pablothedude: Why wouldn't you omit the default?

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
```diff
@@ -48,6 +48,9 @@ class Entity
     const STATE_DRAFT = 'draft';
     const STATE_PUBLISHED = 'published';
 
+    const TYPE_SAML = 'saml20';
```
pablothedude: Shouldn't it be better if a ValueObject is used for re-usability on the long term? For example an EntityTypeObject which could be used throughout the project? Typehinting support would be much better and you could prevent adding logic in concrete implementations using the type of an entity.

MKodde: Yes, that would be much better, and during the design stage of the entities, we also considered this. The biggest reason we decided against this is the way Doctrine works with custom types.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Command/Entity/ChooseEntityTypeCommand.php
MKodde: This is pure convenience. This way the SAML option is checked by default without the need for setting a default in the controller. What would you suggest otherwise?

PR #176, opened 2018-11-28
Set the required Jira ticket fields

MKodde:
Nine fields are set on the ticket. This is still a hypothetical change
which has not yet been tested on an actual Jira installation.

The custom fields for the APPLICANT_NAME|EMAIL are still to be set.

See:
https://www.pivotaltracker.com/story/show/162124271/comments/196880347


pablothedude:
It seems to do the job fine but it's hard at the moment to test the functionality.

PR #177, opened 2018-11-28
Optimize to the service overview 'donut' page

pablothedude:
- thicker donut
- legend based on available state
- label next to state part
- add tooltip (with html, link support)
- add percentage

https://www.pivotaltracker.com/story/show/162112946


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -199,15 +199,34 @@ mail.confirmation.publish_production.service_name: Service name
...
+service.overview.progress.label.intake-conducted: "Intake\ndone"
+service.overview.progress.label.entity-on-test: "Entity\non\ntest"
+service.overview.progress.label.contract-signed: "Contract\nsigned"
+service.overview.progress.label.representative-approved: "Approved\nby\nrepresentative"
+service.overview.progress.label.privacy-questions: "Privacy\nquestions\nanswered"
+service.overview.progress.label.production-connection: "Production\nconnection\naquired"
+service.overview.progress.tooltip.intake-conducted.success: "Thank you\nthe intake was completed"
+service.overview.progress.tooltip.intake-conducted.info: "Is the intake done?"
+service.overview.progress.tooltip.entity-on-test.success: 'Thank you<br> there is a entity on the test environment<br><a href="/">more info</a>'
+service.overview.progress.tooltip.entity-on-test.info: "Is there an entity on the test environment?"
```
MKodde: In order to render wysiwyg editors in the translation interface, please append a .html section to the key. Otherwise html tags will be stripped.

MKodde: To test this: first import the translations

`bin/console lex:trans:import`

Then search for the appropriate translation in the web translator and update it.

Finally invalidat the cahces with the magic blue button.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
```diff
@@ -4,7 +4,85 @@
...
+            <div class="service-status-graph" data-service-id="{{ service.id }}"></div>
+            <div class="service-status-entities">
+
+                <h2>{{ 'service.overview.entitylist.title'|trans({'%name%': service.name}) }}</h2>
+
+                <table>
+                    {# Production entities #}
+                    <tr>
+                        <td colspan="4">
+                            <h3>{{ 'service.overview.entitylist.production.title'|trans }}</h3>
```
MKodde: Maybe render this in the light blue upper case style the form field labels also use. This will make the table less harsh.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
```diff
@@ -4,7 +4,85 @@
...
+                                <td><a href="{{ entity.link }}">{{ entity.name }}</a></td>
+                                <td><a href="{{ entity.link }}">{{ entity.entityId }}</a></td>
+                                <td><a href="{{ entity.link }}">{{ entity.protocol }}</a></td>
+                                <td><a href="{{ entity.link }}">{{ entity.state }}</a></td>
+                            </tr>
+                        {%  endif %}
+                    {% endfor %}
+                    {% if hasEntities == false %}
+                        <tr>
+                            <td colspan="4">{{ 'service.overview.entitylist.empty'|trans }}</td>
```
MKodde: Did you consider removing the table entirely if no entities are present? Or is that not in accordance with @teunfransen s design?

tests/webtests/ServiceOverviewTest.php
```diff
@@ -111,4 +172,23 @@ public function test_service_overview_shows_message_when_no_service_selected()
         $this->assertContains('No service selected', $response->getContent());
         $this->assertContains('Please select a service', $response->getContent());
     }
+
+
+    private function rowsToArray(Crawler $crawler)
```
MKodde: Nice touch, maybe add this to the abstract WebTest class? might be usefull for other similar tests!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -95,8 +111,16 @@ public function overviewAction()
             return $this->redirectToRoute('service_add');
         }
 
+
+        $serviceObjects = [];
+        foreach ($services as $service) {
+            $entityList = $this->entityService->getEntityListForService($service);
```
MKodde: Please read [this](https://www.pivotaltracker.com/story/show/162298797/comments/197068977)

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
```diff
@@ -0,0 +1,103 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace Surfnet\ServiceProviderDashboard\Application\ViewObject;
+
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Service as DomainService;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * @SuppressWarnings(PHPMD.TooManyPublicMethods)
```
MKodde: Copy pasta?

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
```diff
@@ -227,6 +226,14 @@ public function getEnvironment()
         return $this->environment;
     }
 
+    /**
+     * @return string
+     */
+    public function getProtocol()
+    {
+        return 'TODO';
```
MKodde: As in my functional review; maybe default SAML instead of TODO? If only for a smoother demo.

app/js/components/service_status.ts
```diff
@@ -102,14 +146,20 @@ export class ServiceStatus {
         }],
       },
       options: {
-        cutoutPercentage: 85,
+        cutoutPercentage: 75,
```
MKodde: Is the donut fat enough now? Did not see verification from @teunfransen

app/js/components/service_status.ts
```diff
@@ -92,6 +86,56 @@ export class ServiceStatus {
       tooltips.push(states.service.tooltips[key]);
     }
 
+    const customTooltip = function (this: any, tooltipModel: any) {
```
MKodde: Could the tooltip part be levered to an individual file? Making the service_status script more OO.

pablothedude:
**Functional test review**
Thanks for your review. Personally I don't want to make any decisions the design of the service overview page. Because it's purely a gut feeling and very subjective. Therefore I don't want to do these changes and I want to consult @quartje and @teunfransen so they could make the decisions.

Some remarks:
1. @quartje what should happen with the entity links?
2. I've done this on purpose to allow the user to click on links. Also it's very easy to move the tooltip.
3. There isn't a decision made to do so.
4. I did consult @quartje 
5. That's a personal opinion, I think it's looking fine, but I don't want to make the decision myself.
6. I haven't done it to prevent I would forget it. For now I've changed it and added this to the OPENID-connect ticket.
7. @teunfransen what should be done with the coloring of the entity state in the table?

MKodde:
2. When a larger text is in the tooltip, it will overlap the entity list. That tooltip will not hide. Again, I think both the form and the donut tooltips should behave similarly. 
6. Maybe set it to SAML and add a Todo annotation in the comment or somewhere else? Again for us, it does not matter, but for people attending a demo, it might raise unnecessary questions.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
pablothedude: FYI: the html tags will not be stripped you only don't get the fancy editor. Addee `.html` to the key to show the WYSIWYG-editor.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
pablothedude: This seems to me a personal preference.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
pablothedude: There was nothing in the design, and this was the most easy thing to do at the moment.

tests/webtests/ServiceOverviewTest.php
pablothedude: I don't like one god webtestcase, but feel free to reuse

app/js/components/service_status.ts
pablothedude: I do like if this is in the same file as the component. I agree it's not very good quality code but I don't like putting it far away either.

pablothedude:
2: It depends on how much text will be put in the tooltip/popover. Maybe @teunfransen could test this on test2 if it's at the moment sufficient?
6: I did already add it and put a remark on the openid-connect ticket or do you mean something else?

PR #179, opened 2018-11-29
Stop using the Entity view object everywhere 

MKodde:
For all the details see:
https://www.pivotaltracker.com/story/show/162298797


pablothedude:
Thanks!

PR #183, opened 2018-12-03
Improve Jira error reporting to end users

MKodde:
The Jira service might throw an exception whenever it was unable to
create/connect/.. these exceptions where not yet caught by the command
handler. They now are, and a flash message is shown to the user. The
original log message is saved with the logger.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/RequestDeletePublishedEntityCommandHandler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/RequestDeletePublishedEntityCommandHandler.php
@@ -18,20 +18,17 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\CommandHandler\Entity;
 
+use JiraRestApi\JiraException;
 use Psr\Log\LoggerInterface;
 use Surfnet\ServiceProviderDashboard\Application\Command\Entity\RequestDeletePublishedEntityCommand;
 use Surfnet\ServiceProviderDashboard\Application\CommandHandler\CommandHandler;
 use Surfnet\ServiceProviderDashboard\Application\Service\TicketService;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Ticket;
 use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Client\QueryClient as ManageQueryClient;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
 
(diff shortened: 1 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/RequestDeletePublishedEntityCommandHandler.php
```diff
@@ -62,9 +71,13 @@ public function handle(RequestDeletePublishedEntityCommand $command)
         );
         $entity = $this->queryClient->findByManageId($command->getManageId());
         $ticket = Ticket::fromManageResponse($entity, $command->getApplicant());
-        $issue = $this->ticketService->createIssueFrom($ticket);
-
-        $this->logger->info(sprintf('Created Jira issue with key: %s', $issue->key));
-        $this->ticketService->storeTicket($issue->key, $command->getManageId());
+        try {
+            $issue = $this->ticketService->createIssueFrom($ticket);
+            $this->logger->info(sprintf('Created Jira issue with key: %s', $issue->key));
```
pablothedude: Don't you want to put the logging above the try-catch-block or at least at the top of the block?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityList/list.html.twig
```diff
@@ -24,6 +24,12 @@
     </div>
     {% endif %}
 
+    {% for type, messages in app.session.flashbag.all %}
```
pablothedude: Maybe add a partial we could also use on the edit entity view?

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/RequestDeletePublishedEntityCommandHandler.php
MKodde: No, this log entry reports on successfully adding the Jira ticket. Awaiting a successful Jira response seems pivotal in logging this information.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityList/list.html.twig
MKodde: https://www.pivotaltracker.com/story/show/162378323

PR #184, opened 2018-12-03
Translate Jira issue summary and description

MKodde:
Both the summary and description have been made translatable. This
allows the application admin to set the translations for these texts
in the translation web  interface.

The description can be set with three parameters:
 - %applicant_name%
 - %applicant_email%
 - %entity_name%

And the summary can use:
 - %entity_name%

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Ticket.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Ticket.php
@@ -21,22 +21,26 @@ namespace Surfnet\ServiceProviderDashboard\Domain\ValueObject;
 /**
  * See https://bugs.php.net/bug.php?id=66773
  */
+
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Contact as Applicant;
 
 class Ticket
 {
     /** @var string */
-    private $description;
-    /** @var string */
     private $entityId;
     /** @var string */
-    private $summary;
+    private $entityName;
+    /** @var string */
+    private $applicantName;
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Ticket.php
```diff
@@ -38,31 +37,26 @@ class Ticket
...
-        $this->description = $description;
         $this->entityId = $entityId;
+        $this->entityName = $nameEn;
+        $this->applicantName = $applicantName;
+        $this->applicantEmail = $applicantEmail;
     }
 
     public static function fromManageResponse($entity, Applicant $applicant)
     {
         $entityId = $entity['data']['entityid'];
```
pablothedude: The name of the entity doesn't have to be translated?
I also don't like the conversion from manage arrays here, because later on it will take a lot of work if this needs to be refactored later on. For example if the names of the entities should be translated later on. I think it will be prone to bugs because you don't know what this method really will expect. I don't think we have to change it now but we have to take this into account while building further on it.

MKodde: 1. No, it already is translated in Manage / SP Dashboard. I simply choose to display the English translation as English is the main language of the application.
2. Yes, I'm with you on this one. https://www.pivotaltracker.com/story/show/162378398

pablothedude: :+1:

PR #185, opened 2018-12-03
Improve entity delete authorization

MKodde:
By using a voter, the authorization can be tested against the actual
manage entity.


MKodde:
In your use case you are logged in as an admin user? The regular user is not allowed to create a service. I don't understand this part of your question:

> Will you now be able to delete this entity, or should you first authorize again to update the services through the contacts? Could this be problematic, or is this not possible in real life?

Please verify!

pablothedude:
If a Service is created after you logged in the teams are not updated because that'll only happen when logging in. So the voter will check this against 'old' data. And you would be not allowed to the route while you supposed to get there. I think the change this would occur however is small but could lead to user unfriendly situations.

PR #192, opened 2018-12-10
Optimize to the service overview 'donut' page part II

pablothedude:
The links on the service overview page needed some TLC. This commit
will  address this and will link entities to their respective detail
page. Also per service in th elist a link to the entity list and a link
to the privacy questions is added.

https://www.pivotaltracker.com/story/show/162412857

### DIFF
--- a/app/js/components/service_status.test.ts
+++ b/app/js/components/service_status.test.ts
@@ -63,7 +63,7 @@ jest
 describe('validate donut status graph', function() {
 
     let stateHtml = `
-        <div id="service-states">
+        <div class="fieldset card">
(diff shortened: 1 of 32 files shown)


MKodde:
I've taken over this feature from @pablothedude. The initial work was included in my commits, but I've structured the commits differently.

pablothedude:
The `back to overview ` text on the detail page doesn't seem valid if your coming from the service overview. But I don't find it annoying enough to not approve this PR. Maybe we could ask @teunfransen if we should address this later on.

pablothedude:
Approved

PR #188, opened 2018-12-05
Display entity details

MKodde:
Work in progress

PR #193, opened 2018-12-11
Implement the entity detail view part II

MKodde:
See https://www.pivotaltracker.com/story/show/162412987 for further details. The commit shows more about the adjustments that have been made from #188


MKodde:
The build fails on the sec-checker. I'll add a ticket on the backlog to address this and in the process, update the security checker to > 4.0

PR #198, opened 2018-12-12
Show entities on service delete confirmation page

MKodde:
The entity list (as shown on the entity list page) is also shown on the
delete confirmation page. Excluding the actions.

https://www.pivotaltracker.com/story/show/162439601


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -260,6 +260,8 @@ public function deleteAction(Request $request, $serviceId)
             $this->authorizationService->getActiveServiceId()
         );
 
+        $entityList = $this->entityService->getEntityListForService($service);
```
pablothedude: It should be more efficient to fetch the entities after the form is handled.

MKodde: Good point! Moved it

PR #207, opened 2019-01-03
Split the entity overview page on environment

pablothedude:
The entities on the entity overview page should be split based
on the environment of the entity. So production and test entities are
in a separate table. This should make more clear to the sp were
the entity is residing.


pablothedude:
#208 uses this  branch because the added table needed translations as well for the states.

PR #208, opened 2019-01-03
Change the state translations

pablothedude:
The state translations of the entities should be changed to reflect the
actual state much better for sp's. Therefore translations are added.

The changes from #207 are used to prevent merge conflicts because
the entityties were split based on environment.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
@@ -247,7 +247,7 @@ class Entity
      */
     public function getState()
     {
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Entity.php
```diff
@@ -247,7 +247,7 @@ public function getContact()
      */
     public function getState()
     {
-        return $this->state;
+        return 'entity.state.'.$this->environment . '.' . str_replace(' ','_', $this->state);
```
mroest: sprintf() FTW! 😂

PR #210, opened 2019-01-04
Add option to reset oidc secret

pablothedude:
Oidc secrets needed to be reset, this change will make sure that
resetting oidc secrets through manage are possible.


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/ResetOidcSecretCommandHandler.php
```diff
@@ -0,0 +1,96 @@
...
+        $id = (string) Uuid::uuid1();
+        if (!$this->repository->isUnique($id)) {
+            throw new InvalidArgumentException(
+                'The id that was generated for the entity was not unique, please try again'
+            );
+        }
+
+        $secret = new Secret(Entity::OIDC_SECRET_LENGTH);
+
+        $entity->setId($id);
```
mroest: This violates CQRS. You should provide the ID with the command. And maybe the uuid1() should return a unique id that you don't have to worry about (as in not checking it again)

mroest: This should go for line 94

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/ResetOidcSecretCommand.php
```diff
@@ -0,0 +1,135 @@
...
+     */
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    /**
+     * @param string $id
+     */
+    public function setId($id)
```
mroest: Try to make the command readonly and inject the ID in the constructor

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/ResetOidcSecretCommandHandler.php
pablothedude: Fixed

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/ResetOidcSecretCommand.php
pablothedude: Fixed

PR #219, opened 2019-01-14
Fix playground redirect uri

pablothedude:
The playground checkbox didn't work as expected and should add a
playground url when checked and should tick the checkbox and remove
the redirect uri from the form when the redirect uri gets returned from
manage.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
@@ -407,8 +407,8 @@ class Entity
         ManageEntity $manageEntity,
         $environment,
         Service $service,
-        $playGroundUriTest = '',
-        $playGroundUriProd = ''
+        $playGroundUriTest,
+        $playGroundUriProd
     ) {
         $metaData = $manageEntity->getMetaData();
         $coin = $metaData->getCoin();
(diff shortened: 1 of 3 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/163120130

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
```diff
@@ -502,6 +495,25 @@ public static function fromManageResponse(
         return $entity;
     }
 
+    private static function setRedirectUrisFromManageResponse(
+        Entity $entity,
+        ManageEntity $manageEntity,
+        $environment,
+        $playGroundUriTest = '',
```
MKodde: If you add input validation in `\Surfnet\ServiceProviderDashboard\Application\Service\LoadEntityService::__construct` using `Assert::stringNotEmpty` you can kind of prevent the `$playGroundUriTest = ''` in this method. Although it's not guaranteed that the LoadEntityService only uses this static constructor.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
```diff
@@ -502,6 +495,25 @@ public static function fromManageResponse(
...
+        $playGroundUriProd = ''
+    ) {
+        $redirectUris = $manageEntity->getOidcClient()->getRedirectUris();
+        $playGroundUri = ($environment == self::ENVIRONMENT_PRODUCTION ? $playGroundUriProd : $playGroundUriTest);
```
MKodde: If the playgroundUriProd is set to '', we now default to playgroundTest. Is that what we want? In other words, what's the meaning of the default values for the playGroundUriProd and playGroundUriTest?

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
pablothedude: When I first created the optional parameters in the entity domain entity it was not possible to make them mandatory. Because the oidc logic is now completed I'm able to make them mandatory. Thanks for pointing that out.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
pablothedude: See comment above.

PR #220, opened 2019-01-14
Fix oidc clientId when publishing to production

pablothedude:
The clientId when publishing the entity to production had a : replaced
with @. This shouldn't be done, and this commit will fix that behaviour.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
@@ -74,7 +74,7 @@ class SaveOidcEntityCommand implements Command
      * @Assert\Url()
      * @SpDashboardAssert\ValidClientId()
      */
-    private $clientId;
+    private $entityId;
 
     /**
      * @var string
@@ -372,7 +372,7 @@ class SaveOidcEntityCommand implements Command
         $command->service = $entity->getService();
         $command->archived = $entity->isArchived();
         $command->environment = $entity->getEnvironment();
-        $command->clientId = $entity->getEntityId();
+        $command->entityId = $entity->getEntityId();
         $command->clientSecret = $entity->getClientSecret();
         $command->redirectUris = $entity->getRedirectUris();
         $command->grantType = $entity->getGrantType();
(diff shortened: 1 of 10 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/163117541

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
```diff
@@ -482,17 +482,25 @@ public function setEnvironment($environment)
...
-        return $this->clientId;
+        $this->entityId = $entityId;
     }
 
     /**
-     * @param string $clientId
+     * @return string
      */
-    public function setClientId($clientId)
+    public function getClientId()
```
MKodde: Is this getter actually used? I could not find any usages.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -265,7 +265,7 @@ private function generateSecurityMetadata(Entity $entity)
      */
     private function generateOidcClient(Entity $entity)
     {
-        $metadata['clientId'] = $entity->getEntityId();
+        $metadata['clientId'] = str_replace('://', '@//', $entity->getEntityId());
```
MKodde: Perhaps you could use `ViewObject/Entity::getEntityId` here?

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
pablothedude: `\Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Form\Entity\OidcEntityType`

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
pablothedude: No, that is not the responsibility of a viewobject. The replacement is only for Manage to understand the clientId so I find this a much better place.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
pablothedude: The clientId is only for the presentation of the clientId.

MKodde: Ahh, yes

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
MKodde: I now see that the entity in this method is a Domain entity and not a View object. I agree with you it would be a hassle to convert that domain entity to a view object just to get the client id.

The best solution might be to introduce a client id conversion helper class that encapsulates the str_replace logic. But lets not go down that route now. This suffices for now!

PR #221, opened 2019-01-14
Create Jira issue after publishing to manage prod

MKodde:
The mail logic was substituted by the creation of a Jira ticket. The possibility to set a customized issue type was added to the existing setup.

This change utilizes the same Jira ticket that is created upon deletion request. The one difference is that the issue type: `spd-request-production-entity` is used.

This issue type can be configured in the app/parameters.yml under the `jira_issue_type_publication_request` moniker. More information in commit: 2b94d1b4

The previous mail solution was removed entirely.


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PublishEntityProductionCommandHandler.php
```diff
@@ -109,26 +105,30 @@ public function handle(PublishEntityProductionCommand $command)
...
             if (array_key_exists('id', $publishResponse)) {
                 // Send the confirmation mail
                 $this->logger->info(
-                    sprintf('Sending publish request mail to service desk for "%s".', $entity->getNameNl())
-                );
-                $mailCommand = new PublishToProductionMailCommand(
-                    $this->buildMailMessage($entity)
+                    sprintf('Sending publish request mail to service desk for "%s".', $entity->getNameEn())
                 );
-                $this->commandBus->handle($mailCommand);
```
pablothedude: Thanks for getting rid of the recursion for handling commands in the commandbus. I thuink this is very valuable on the long run.

src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PublishEntityProductionCommandHandler.php
```diff
@@ -48,48 +42,50 @@ class PublishEntityProductionCommandHandler implements CommandHandler
...
         EntityRepository $entityRepository,
         PublishEntityRepository $publishClient,
-        CommandBus $commandBus,
-        MailMessageFactory $mailMessageFactory,
-        TokenStorageInterface $tokenStorage,
+        TicketService $ticketService,
         FlashBagInterface $flashBag,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        $issueType
```
pablothedude: I think it would be better if this parameter is mandatory to prevent a lot of if statements later on. Or perhaps remove it here if it's not the responsibility of the ticket.

MKodde: The parameter is required, but you might mean the parameters.yml parameter? That too is an application requirement. I think bootstrapping will crash if the parameter is not set, the assertion is a mere DiD & clarification means to ensure you know where that parameter should come from.

pablothedude: Sorry, I meant the Ticket::issueType

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Ticket.php
```diff
@@ -37,14 +38,17 @@ class Ticket
     private $applicantName;
     /** @var string */
     private $applicantEmail;
+    /** @var string|null */
+    private $issueType;
 
-    public function __construct($entityId, $manageId, $nameEn, $applicantName, $applicantEmail)
+    public function __construct($entityId, $manageId, $nameEn, $applicantName, $applicantEmail, $issueType)
```
pablothedude: I think it would be better if this parameter is mandatory to prevent a lot of if statements later on. Or perhaps remove it here if it's not the responsibility of the ticket.

MKodde: Still, I do not fully understand what you mean with it having to be mandatory. It is mandatory, right? No default value is set. And furthermore, an additional assertion is added to verify a non-empty string is set in the parameters.yml.

Please enlighten me :)

pablothedude: It's suggested with [typehinting](https://github.com/SURFnet/sp-dashboard/blob/bbd1b4d9301a64dc59894ac1160e6bbf2c217440/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Ticket.php#L41) and later on the issueType is [checked](https://github.com/SURFnet/sp-dashboard/blob/bbd1b4d9301a64dc59894ac1160e6bbf2c217440/src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Factory/IssueFieldFactory.php#L117) if it's set.

[](url)

MKodde: Thanks, I've refactored some of the code. See the comment below for more information.

MKodde:
Added to the previous changeset:
1. The ticket is now set with the issue type regardless of which issue you are creating (entity removal or publishing). Each ticket should describe which purpose it intends. The `IssueFieldFactory` is no longer DI'ed with a default entity type.
2. The ticket now is set with the description and summary of the ticket. For the same reason as point 1.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/RequestDeletePublishedEntityCommand.php
```diff
@@ -36,18 +36,42 @@ class RequestDeletePublishedEntityCommand implements Command
...
+     * @param string $issueType
+     * @param string $issueSummaryTranslationKey
+     * @param string $issueDescriptionTranslationKey
      */
-    public function __construct($manageId, Contact $applicatant)
-    {
+    public function __construct(
+        $manageId,
+        Contact $applicant,
+        $issueType,
```
pablothedude: I don't think it should not be the responsibility of the command to have these fields but it should be the responsibility of the handler. Because this prevents coupling in the command and the controller.  Therefore I think they could be better set with DI in the Handler and not through the command. 
```
issueType
issueSummaryTranslationKey
issueDescriptionTranslationKey
```

pablothedude:
I think it is a very valuable addition and could prevent bugs in the future. thanks!

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/RequestDeletePublishedEntityCommand.php
MKodde: Thanks for the feedback. I'm merging this PR and will include your feedback in the next PR that addresses this logic.

PR #222, opened 2019-01-14
Uninstall Mailcatcher

MKodde:
As we no longer send mail messages from SP Dashboard. The mail catcher
is removed from provisioning.


MKodde:
It turns out we still want to be able to send mail messages from SP Dashboard. Closing this PR without merging.

PR #223, opened 2019-01-15
Fix routes missing services

pablothedude:
The given service of an entity is not set for certain entity routes
which will throw an exception because a service is needed. This was not
an issue before because the active service was set in the session. But
because we now have overview pages with multiple services. We should
add the service to the route to know the active service on forehand.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
@@ -23,9 +23,7 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Template;
 use Surfnet\ServiceProviderDashboard\Application\Exception\InvalidArgumentException;
-use Surfnet\ServiceProviderDashboard\Application\Exception\ServiceNotFoundException;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity;
-use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Command\Entity\ChooseEntityTypeCommand;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Form\Entity\ChooseEntityTypeType;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
@@ -44,7 +42,7 @@ class EntityCreateController extends Controller
     /**
      * @Method({"GET", "POST"})
      * @Route(
-     *     "/entity/create/type/{targetEnvironment}",
+     *     "/entity/create/type/{serviceId}/{targetEnvironment}",
      *     defaults={
      *          "targetEnvironment" = "test",
      *     },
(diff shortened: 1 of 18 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/163117411

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
```diff
@@ -55,31 +53,40 @@ class EntityCreateController extends Controller
...
             'form' => $form->createView(),
+            'serviceId' => $service->getId(),
             'environment' => $targetEnvironment,
         ];
     }
 
     /**
      * @Method({"GET", "POST"})
-     * @Route("/entity/create/{type}/{targetEnvironment}", name="entity_add")
+     * @Route("/entity/create/{type}/{serviceId}/{targetEnvironment}", name="entity_add")
```
MKodde: Is it possible to keep the order: /entity/intent/{serviceId}/{other params}. This makes the routes somewhat more predictable in my opinion. Other routes that include the service id also stick to this 'convention'

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityDeleteController.php
```diff
@@ -55,20 +56,31 @@ class EntityDeleteController extends Controller
...
     private $authorizationService;
 
+    /**
+     * @var ServiceService
+     */
+    private $serviceService;
+
     /**
      * @param CommandBus $commandBus
+     * @param EntityService $entityService
```
MKodde: The doc block does not add any context as all parameters are typehinted explicitly

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/AuthorizationService.php
```diff
@@ -226,6 +227,22 @@ function ($id) use ($contact) {
         );
     }
 
+    /**
+     * @param $serviceId
+     * @return \Surfnet\ServiceProviderDashboard\Domain\Entity\Service|null
+     * @throws ServiceNotFoundException
+     */
+    public function getServiceById($serviceId)
```
MKodde: A nice addition! Addidng additional verification logic is very usefull here.

Some thoughts:

1. The service can still hypothetically (very slim chance, as hasAccessToService already verified this. It is bad practice to count on this as that method might change or be replaced/removed) be null, as `return $this->serviceService->getServiceById($serviceId);` still uses the 'findById' method on the `ServiceService`. A null response here should also result in the `ServiceNotFoundException`. 
2. Calling `$this->setSelectedServiceId($serviceId);` should than be performed after getting the actual service.
3. The Exception message here can be made more verbose. I suggest somthing like: `Unable to find service with id %d`

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
pablothedude: I don't think this is going to have much impact either way. So changed it.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityDeleteController.php
pablothedude: They were already there and were incomplete so I fixed them, I agree we could remove them completely.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/AuthorizationService.php
pablothedude: 1. Changed
2. Changed
3. Made more verbose but without the parameter

MKodde:
You merged this with failing web tests. You probably need to add some mock responses to the guzzle queues.

PR #225, opened 2019-01-16
Update the entity publication workflow

MKodde:
1. The Jira ticket is created first. After which the publication action is performed. When one or the other fails, the user is informed that publication failed. And the Jira ticket is removed if it was created.
2. The previously removed mailing logic is reverted in part
3. When Jira was down, send an e-mail to the service desk

More information: https://www.pivotaltracker.com/story/show/163217521


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PublishEntityProductionCommandHandler.php
```diff
@@ -103,32 +133,49 @@ public function handle(PublishEntityProductionCommand $command)
...
+
+        // 2. On success, publish the entity to production
         try {
             $this->logger->info(
                 sprintf('Publishing entity "%s" to Manage in production environment', $entity->getNameEn())
             );
-
             $publishResponse = $this->publishClient->publish($entity);
-
             if (array_key_exists('id', $publishResponse)) {
```
pablothedude: If manage doesn't return an id, and therefore the entity isn't published it doesn't seems handle this situation properly. And in the end no one would get informed.

MKodde: I don't think Manage will ever return a response without an ID except for error situations, which are handled as Exceptions. The Manage PublishEntityClient only returns an array response if the status of the response was OK. Still it's a small task to add an additional error catch.

MKodde: See 3bfe0e5

PR #226, opened 2019-01-18
Fix the dump method in twig template for prod

pablothedude:
In the mail template to notify the support desk when a publication
has failed there is a `dump` method. However this method is not
supported in the prod environment. So this is removed.


MKodde:
The more you know :)

Keeping the trace was not a requirement?

pablothedude:
For now the message seems good enough and else we should have the logs themselves.

PR #227, opened 2019-01-22
Fix manage certificate metadata

pablothedude:
When a certificate is fetched from manage the start- and end-tag are
stripped. This results in an error when submitting a form again because
the SSL validator thinks it isn't a valid certificate. Therefore
the tags are added in the metadata dto to always return a certificate
with those tags.

https://www.pivotaltracker.com/story/show/163385303


src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/MetaData.php
```diff
@@ -194,7 +194,10 @@ public function getSignatureMethod()
 
     public function getCertData()
     {
-        return $this->certData;
+        if ($this->certData) {
```
MKodde: Judging the way EngineBlock deals with certificates from manage, we might need to revise this a little bit. For example, EB is aware that multiple certificates might be in place. It is knowledgable on how to deal with them. I don't say we need to copy this logic. For now, using the primary certificate should do the trick!

What I do think is a valuable addition is to add a certificate validator/parser. Maybe the certificate of some other entity is set with the BEGIN/END 'envelope', this change would break that certificate.

For inspiration you might want to check: `\OpenConext\EngineBlock\Metadata\Entity\Assembler\PushMetadataAssembler::assembleCertificates`

PR #229, opened 2019-01-23
Fix oidc client id on detail page

pablothedude:
The : in an oidc entityId should be replaced with a @. This was done
on other pages except on the entity detail page were it was missing in
the corresponding viewobject.

https://www.pivotaltracker.com/story/show/163417371

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
@@ -323,7 +323,10 @@ class EntityDetail
      */
     public function getEntityId()
     {
-        return $this->entityId;
+        if ($this->getProtocol() !== DomainEntity::TYPE_OPENID_CONNECT) {
+            return $this->entityId;
+        }
+        return str_replace('://', '@//', $this->entityId);
     }
 
     /**


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
```diff
@@ -323,7 +323,10 @@ public function getAcsLocation()
      */
     public function getEntityId()
     {
-        return $this->entityId;
+        if ($this->getProtocol() !== DomainEntity::TYPE_OPENID_CONNECT) {
```
MKodde: Still not completely sold on the str_replace '://', '@//' occurrences popping up in the code in several places. What bothers me the most is that this is not really semantically sound. You are not getting an entity id, but a client id.

As this is just cosmetically, we could think about making a Twig filter that displays the entity id as a client id.

Feel free to postpone this change for now, but it just doesn't sit right with me.

pablothedude: I agree this is not the best solution. However it is a entityId because an EntityId of an OIDC entity is the client id with the column replaced. I think the best option would be to use a value object for the EntityId but because this is a temporary solution I assume the current solution will be sufficient. but I didn't want to introduce value objects only for this use case for a temporary solution.

PR #231, opened 2019-01-23
Makerelease: Make sure the node_modules dir is removed

quartje:
This will save 35.000 files for the release :)


MKodde:
Nice proposal! In my opinion, the proof is in the pudding. Does it chooch?

PR #233, opened 2019-02-04
Correctly publish OIDC redirectUris to Manage

MKodde:
The array that is passed along to manage should be a list, not a map.
If the php indexes are updated along the way, by adding/removing items
from the form. The index order might not be sequentially pristine. This
causes JSON encode to generate a map instead of a list.

By resetting the list before publishing, this problem should be remedied

https://www.pivotaltracker.com/story/show/163646662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Next release
 
+**Bugfix**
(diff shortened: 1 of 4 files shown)


MKodde:
Yes, resetting the data while we are setting it on entity is more elegant.

PR #236, opened 2019-02-19
Change service switcher behavior

MKodde:
The service switcher should now redirect to the service overview (donut) overview page for the selected service. This is basically a filtered view on the existing donut overview page which showed all available services for the admin user.

See commits for details about the changes that are introduced.

https://www.pivotaltracker.com/story/show/163752485

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
@@ -281,7 +281,7 @@ class ServiceController extends Controller
 
         $this->commandBus->handle($command);
 
-        return $this->redirectToRoute('entity_list', ['serviceId' => $serviceId]);
+        return $this->redirectToRoute('service_admin_overview', ['serviceId' => $serviceId]);
     }
 
     /**
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
@@ -296,19 +296,9 @@ class ServiceController extends Controller
      */
     public function adminOverviewAction($serviceId)
     {
-        $allowedServices = $this->authorizationService->getAllowedServiceNamesById();
-        if (!isset($allowedServices[$serviceId])) {
-            throw $this->createNotFoundException(sprintf('Service with id "%s" cannot be found', $serviceId));
-        }
-        $allowedServices = [$serviceId => $allowedServices[$serviceId]];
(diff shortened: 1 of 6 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -281,7 +281,39 @@ public function selectAction(Request $request)
...
+
+    /**
+     * @Method({"GET"})
+     * @Route("/service/{serviceId}", name="service_admin_overview")
+     * @Security("has_role('ROLE_ADMINISTRATOR')")
+     * @Template("@Dashboard/Service/overview.html.twig")
+     */
+    public function adminOverviewAction($serviceId)
+    {
+        $allowedServices = $this->authorizationService->getAllowedServiceNamesById();
```
pablothedude: Just a suggestion: you could use the `\Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service\AuthorizationService::getServiceById` and the `\Surfnet\ServiceProviderDashboard\Application\Service\ServiceService::getServiceById`. To keep the responsibility for fetching and validating services at the designated service. This will also keep your code flexible and easy to refactor later on. And less lines of code are less prone to errors. :wink:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Menu/Builder.php
```diff
@@ -37,8 +44,14 @@ public function mainMenu(array $options)
             return $menu;
         }
 
-        $menu->addChild('Services', ['route' => 'service_overview']);
-
+        if ($this->authorizationService->isAdministrator() && $this->authorizationService->getActiveServiceId()) {
+            $menu->addChild('Service overview', [
+                'route' => 'service_admin_overview',
+                'routeParameters' => ['serviceId' => $this->authorizationService->getActiveServiceId()],
```
pablothedude: What happens with the activeServiceId if your session is timing out?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
MKodde: You ara absolutely right, I was in a tunnel vision to get things done. I've greatly simplified the `adminOverviewAction` method.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Menu/Builder.php
MKodde: It is lost just like I would expect it to be. Do you see a problem I do not?

PR #238, opened 2019-02-19
Configure the workflow state

MKodde:
In order to be more flexible towards changes in workflow states in Manage. We are now able to configure the desired workflow states in SP Dashboard. 

In order to make this configuration more manageable in the application, the config.yml is updated to include a section for the dashboard bundle. Here the manage related config is now specified. Parameters.yml is still the place to configure the actual param values.

Manage config VOs have been created to reflect the configuration set in the config.yml. For details, see the individual commit messages.

https://www.pivotaltracker.com/story/show/164005623


tests/unit/Application/Metadata/JsonGeneratorTest.php
```diff
@@ -433,13 +410,13 @@ public function test_it_can_build_oidc_entity_data_for_new_entities()
                             'arp' => 'arp',
                         ),
                     'type' => 'saml20-sp',
+                    'state' => 'testaccpeted',
```
pablothedude: typo?

MKodde: In the test this value is arbitrary, but a typo nonetheless.

PR #240, opened 2019-02-21
Update the production connection status on publication

MKodde:
The status was not updated on the Service entity, The first time when
the status is not yet set to either requested or active, the status
is updated.

https://www.pivotaltracker.com/story/show/163732191


pablothedude:
I think a test should be a valuable addition. Also we should maybe look into a state engine of some sort later on to keep the PublishEntity*CommandHandlers maintainable. The latter is a suggestion we could very well address later on.

MKodde:
Yes, some handlers are growing and growing. I'm putting a todo on my list to create a TD ticket for this.

I also fixed the test.

PR #241, opened 2019-02-21
Change detailed route on overview page to prevent 403

pablothedude:
Change route on overview page for non admins to the entity detail
page. This should be changed because the currently used route should
not be accessible for non admins. This in order to prevent an exception
when trying to access the edit page for non admins when not allowed.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Service/overview.html.twig
@@ -49,7 +49,7 @@
             <div class="service-status-container">
 
                 <div class="service-status-title">
-                    <a href="{{ path('select_service', {service: service.id }) }}">{{ service.name }}</a>
+                    <a href="{{ path('entity_list', {serviceId: service.id }) }}">{{ service.name }}</a>
                 </div>
                 <div class="service-status-graph" data-service-id="{{ service.id }}"></div>
                 <div class="service-status-entities">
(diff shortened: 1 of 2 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/164135669

app/config/parameters.yml.dist
```diff
@@ -52,7 +52,7 @@ parameters:
     manage_prod_host: 'https://manage-prod.dev.support.surfconext.nl'
     manage_prod_username: sp-dashboard
     manage_prod_password: secret
-    manage_prod_publication_status_create: prodaccepted
+    manage_prod_publication_status_create: notprodaccepted
```
MKodde: Accidental commit?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -300,6 +300,8 @@ public function adminOverviewAction($serviceId)
         $entityList = $this->entityService->getEntityListForService($service);
         $serviceList = new ServiceList([Service::fromService($service, $entityList, $this->router)]);
 
+        echo $service->getTeamName();
```
MKodde: Debug leftover?

pablothedude: Removed

app/config/parameters.yml.dist
pablothedude: Removed

pablothedude:
No didn't forget to commit

MKodde:
Thanks for the fix, to prevent future regression, I've added a web test covering this behavior,

PR #242, opened 2019-02-22
Fix manage query state for prod

pablothedude:
All entities were fetched with the 'testaccepted' state. This resulted
in entities not being fetched after they where published with
'prodaccepted'. This fix will get the entities based on the state
configured in the parameters.yml.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
@@ -255,6 +255,8 @@ services:
             - '@logger'
             - '%playground_uri_test%'
             - '%playground_uri_prod%'
+            - '%manage_test_publication_status_update%'
+            - '%manage_prod_publication_status_update%'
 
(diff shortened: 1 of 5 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/164163949

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
```diff
@@ -255,6 +255,8 @@ services:
             - '@logger'
             - '%playground_uri_test%'
             - '%playground_uri_prod%'
+            - '%manage_prod_publication_status_update%'
```
MKodde: Instead of injecting the params directly, you could have used the Manage Config VO ®.

MKodde: More importantly, the param values are the other way aroud, the prod value is set for the test entities and vise versa.

MKodde:
As discussed, I've taken over this PR and fixed the feedback I gave this morning. 

Merging

PR #244, opened 2019-03-13
Add grantType to the translations

MKodde:


MKodde:
Merging, and opening a PR to fix the security check that breaks the build. See #245

PR #246, opened 2019-03-14
Ansible cleanup

quartje:
Removed redundant translation extractions and cleaned up some yum install commands.

### DIFF
--- a/ansible/roles/spdashboard/tasks/spdashboardvm.yml
+++ b/ansible/roles/spdashboard/tasks/spdashboardvm.yml
@@ -1,11 +1,12 @@
 ---
 - name: Install some packages needed for development
-  yum: pkg={{ item }} state=present
-  with_items:
-    - ruby-devel
-    - gcc-c++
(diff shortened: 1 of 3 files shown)


ansible/roles/spdashboard/tasks/spdashboardvm.yml
```diff
@@ -1,11 +1,12 @@
 ---
 - name: Install some packages needed for development
-  yum: pkg={{ item }} state=present
-  with_items:
-    - ruby-devel
-    - gcc-c++
-    - sqlite-devel
-    - ant
+  yum:
```
MKodde: Nice, this notation is much more readable than the previous 'with_items' solution! I assume we do not have to support prehistoric (pre 2.0) ansible versions?

quartje: Indeed!

MKodde:
Merging!

PR #247, opened 2019-05-10
Fix the playground url after copy to production

pablothedude:
The entity was copied but used the wrong environment so the playground
url from test was added to the redirect uri's and the checkbox was
unchecked.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/LoadEntityService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/LoadEntityService.php
@@ -132,7 +132,7 @@ class LoadEntityService
         // Convert manage entity to domain entity
         $domainEntity = Entity::fromManageResponse(
             $manageEntity,
-            $environment,
+            $sourceEnvironment,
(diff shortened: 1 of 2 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/163634932

src/Surfnet/ServiceProviderDashboard/Application/Service/LoadEntityService.php
```diff
@@ -132,7 +132,7 @@ public function load($dashboardId, $manageId, Service $service, $sourceEnvironme
         // Convert manage entity to domain entity
         $domainEntity = Entity::fromManageResponse(
             $manageEntity,
-            $environment,
+            $sourceEnvironment,
```
MKodde: Can this somehow be verified in the `LoadEntityServiceTest`? Or alternatively, in a functional test?

pablothedude: Done

PR #248, opened 2019-05-10
Prevent a space in the secret when copying

pablothedude:
When clicking on the secret to copy it a space was selected before the
secret and this is very helpful. Therefore removed the spacing.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityDetail/detailTextField.html.twig
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityDetail/detailTextField.html.twig
@@ -5,7 +5,6 @@
             <i class="fa fa-question-circle"></i>
         </span>
     {% endif %}
-    <label>{{ label }}</label>
-    <span>{{ value }}</span>
+    <label>{{ label }}</label><span>{{ value }}</span>
 </div>
 {% endif %}


pablothedude:
https://www.pivotaltracker.com/story/show/163634841

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityDetail/detailTextField.html.twig
```diff
@@ -5,7 +5,6 @@
             <i class="fa fa-question-circle"></i>
         </span>
     {% endif %}
-    <label>{{ label }}</label>
-    <span>{{ value }}</span>
+    <label>{{ label }}</label><span>{{ value }}</span>
```
MKodde: I personally prefer using 'spaceless', keeping the Twig markup slightly more readable.

https://twig.symfony.com/doc/2.x/tags/spaceless.html

pablothedude: To me the solution looks equal, and if there were more fields I would certainly go for a spaceless filter, but for now removing the newline seems sufficient enough.

Be aware you're referring to a deprecated function, see:
https://twig.symfony.com/doc/2.x/filters/spaceless.html

PR #250, opened 2019-05-13
Add transactions middleware to commandbus

pablothedude:
Add transactions to the command bus. This, to prevent duplicate entities when storing entities and an unexpected exception occurs.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
@@ -140,7 +140,7 @@ class Service
     /**
      * @var \Doctrine\Common\Collections\ArrayCollection
      *
-     * @ORM\OneToMany(targetEntity="Entity", mappedBy="service", cascade={"persist"}, orphanRemoval=true)
+     * @ORM\OneToMany(targetEntity="Entity", mappedBy="service", cascade={"persist"})
      */
     private $entities;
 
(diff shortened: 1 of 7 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
```diff
@@ -140,7 +140,7 @@ class Service
     /**
      * @var \Doctrine\Common\Collections\ArrayCollection
      *
-     * @ORM\OneToMany(targetEntity="Entity", mappedBy="service", cascade={"persist"}, orphanRemoval=true)
+     * @ORM\OneToMany(targetEntity="Entity", mappedBy="service", cascade={"persist"})
```
MKodde: Did you remove this because the removal logic is also explicitly added in code in the service remove command handler?

I think this should not be part of this PR, but maybe I'm overlooking something here..

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
```diff
@@ -10,6 +10,10 @@ services:
         resource: '../../Security/Voter'
         tags: ['security.voter']
 
+    surfnet.dashboard.middleware.doctrine_transaction:
+        class: Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Middleware\DoctrineTransactionMiddleware
+        public: true
```
MKodde: Should this service be public?

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
MKodde: @pablothedude explained orphan removal is no longer required when using the transactions. Orphan removal caused cleanup of orphaned entities that later where tried to be removed by the cascading delete option.

MKodde:
The web test would entail adding a mock handler that throws some kind of exception. Not that hard, but might interfere with other future tests. For now I refrained from adding that web test. The added unit test should provide enough test coverage.

PR #251, opened 2019-05-13
Fix urn validator regex

pablothedude:
The regex for the URN validator was faulty and a slash was missing
escaping. This resulted in a false negative when validating URN's.

### DIFF
--- a/app/js/components/validation.js
+++ b/app/js/components/validation.js
@@ -127,7 +127,7 @@ function validateUrl(value) {
 }
 
 function validateUrn(value) {
-    var regExp = /^urn:[a-z0-9][a-z0-9-]{0,31}:[a-z0-9()+,\-.:=@;$_!*'%/?#]+$/i;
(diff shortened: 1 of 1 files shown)


app/js/components/validation.js
```diff
@@ -127,7 +127,7 @@ function validateUrl(value) {
 }
 
 function validateUrn(value) {
-    var regExp = /^urn:[a-z0-9][a-z0-9-]{0,31}:[a-z0-9()+,\-.:=@;$_!*'%/?#]+$/i;
+    var regExp = /^urn:[a-z0-9][a-z0-9-]{0,31}:[a-z0-9()+,\-.:=@;$_!*'%\/?#]+$/i;
```
MKodde: This can easily be covered with a Jest unit test right?

pablothedude: I've looked into this and didn't want to take the effort to refactor the js code to ts. And because this was just a third party regex which was copied without taking valid escaping into account I don't want to test it now.

PR #252, opened 2019-05-13
Change the donutstate with empty service

pablothedude:
If a service has no entities the donut should reflect that. So it
must not look if any progess so far has been made.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
@@ -194,8 +194,6 @@ class ServiceStatusAssembler
         return $states;
     }
 
-
-
     /**
      * @param array $states
      * @return array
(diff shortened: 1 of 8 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/164241210

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -194,6 +194,8 @@ private function getStates(Service $service, ServiceStatusService $serviceStatus
         return $states;
     }
 
+
```
MKodde: Nitpick: Remove exessive whitespace

pablothedude: Done

pablothedude:
Because there isn't much logic in the api controller I think I could do without a webtest. I think the tests in the `ServiceStatusAssemblerTest` will be sufficient for now.

PR #253, opened 2019-05-13
Add translatable footer links

MKodde:
The `contact` and `privacy` links were set to a hard '/'. This is addressed by setting adding them to the translations. Allowing overriding these URL's from the web translation interface. Some dummy defaults are set in the messages.en.yml.

See: https://www.pivotaltracker.com/story/show/165965469

### DIFF
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -51,8 +51,8 @@
         </div>
         <div class="footer">
             <div class="footer-inner">
-                <span><a href="{{ 'page.footer.privacy_link'|trans }}" target="_blank">{{ 'page.header.privacy'|trans }}</a></span>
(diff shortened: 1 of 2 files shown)


app/Resources/views/base.html.twig
```diff
@@ -51,8 +51,8 @@
         </div>
         <div class="footer">
             <div class="footer-inner">
-                <span><a href="/" target="_blank">{{ 'page.header.privacy'|trans }}</a></span>
-                <span><a href="/" target="_blank">{{ 'page.header.contact'|trans }}</a></span>
+                <span><a href="{{ 'page.footer.privacy_link'|trans }}" target="_blank">{{ 'page.header.privacy'|trans }}</a></span>
```
pablothedude: Is the header/footer mismatch in the name deliberate?

pablothedude: Added the translation to prevent regressions in the future

PR #254, opened 2019-05-14
Cascade delete privacy questions on service removal

MKodde:
The relationship between service and privacy question is now configured to apply a cascading delete on privacy question entities when a service is removed. In addition, Doctrine is instructed to remove orphaned entities.

**Bugreport:** https://www.pivotaltracker.com/story/show/165237921
**See:** https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-associations.html#transitive-persistence-cascade-operations
**See:** https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-associations.html#orphan-removal

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
@@ -147,7 +147,7 @@ class Service
     /**
      * @var PrivacyQuestions
      *
-     * @ORM\OneToOne(targetEntity="PrivacyQuestions", mappedBy="service")
+     * @ORM\OneToOne(targetEntity="PrivacyQuestions", mappedBy="service", cascade={"remove"}, orphanRemoval=true)
      */
     private $privacyQuestions;
 
--- a/tests/webtests/ServiceDeleteTest.php
+++ b/tests/webtests/ServiceDeleteTest.php
@@ -98,7 +98,7 @@ class ServiceDeleteTest extends WebTestCase
         $this->assertEquals(
             '/',
             $this->client->getRequest()->getRequestUri(),
-            "Expected to be on the service overview page after succesfully removing the service"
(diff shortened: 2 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
```diff
@@ -147,7 +147,7 @@ class Service
     /**
      * @var PrivacyQuestions
      *
-     * @ORM\OneToOne(targetEntity="PrivacyQuestions", mappedBy="service")
+     * @ORM\OneToOne(targetEntity="PrivacyQuestions", mappedBy="service", cascade={"remove"}, orphanRemoval=true)
```
pablothedude: I don't like this solution because it goes against the application architecture as defined in the documentation of the project. Also this behavior is entirely different then the removal of other related entities (see the removal of an `Entity`) in a `Service`. So I would suggest to remove this responsibility away from the entity and remove it to the command handler.

MKodde: Yes, this would make the application better. For example, we would then also support table engines that do not support foreign key relations. On the other hand, during design time we accepted the cluttering of the domain layer with some specific Doctrine (infra) annotations. Because it worked really well for us. Alternatively, we could have used YML config for doctrine entity configuration.

My suggestion is to put this in a technical debt ticket. For now, I don't see good reason to change this. And when we would, we should fix it throughout the lines.

pablothedude: Added this thread to the 'remove drafts' story:
https://www.pivotaltracker.com/story/show/166004131/comments/203097098

MKodde:
Thanks for adding this to the upcoming draft removal story! For now, I'm merging this feature as is.

PR #258, opened 2019-05-15
Rename 'team name' to 'team identifier'

pablothedude:
The name of the field on the srevie form needs to be changed to be
describe more logically what the purpose of the field is.


pablothedude:
https://www.pivotaltracker.com/story/show/157389551

PR #259, opened 2019-05-16
Add translatable contact email placeholder

pablothedude:
The placeholders of the contact email address fields should be
translatable. So a placeholder is added and a translation is added.


pablothedude:
https://www.pivotaltracker.com/story/show/165716972

PR #261, opened 2019-05-21
Add issue repository for testing

pablothedude:
In order to allow local development without the need of a jira
server a repository is added to reproduce the behavior of jira.
To prevent coupling with third party dependencies a couple of domain
specific valueobjects were added to return from the repositories.

https://www.pivotaltracker.com/story/show/165993367

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * Copyright 2019 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Domain\ValueObject;
+
+class Issue
+{
+    /**
+     * @var string
+     */
+    private $key;
+    /**
+     * @var string
+     */
+    private $issueType;
+
+    /**
+     * @param string $key
+     * @param string $issueType
+     */
+    public function __construct($key, $issueType)
+    {
+        $this->key = $key;
+        $this->issueType = $issueType;
+    }
+
+    /**
+     * @return string
+     */
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    /**
+     * @return string
+     */
+    public function getIssueType()
+    {
+        return $this->issueType;
+    }
+}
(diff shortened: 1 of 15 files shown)


pablothedude:
@MKodde could you test the jira repository with your local dev stack with `\Surfnet\ServiceProviderDashboard\Tests\Unit\Infrastructure\DashboardBundle\Jira\Repository\IssueRepositoryTest`. It should work for the Jira repository.

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @var string
+     */
+    private $issueType;
+
+    /**
+     * @param string $key
+     * @param string $issueType
+     */
+    public function __construct($key, $issueType)
```
MKodde: You might want to add input validation at this point!

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/IssueCollection.php
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Domain\ValueObject;
+
+class IssueCollection implements \Countable
```
MKodde: Even though this is a very lightweight, not complex class. I would like to see a unit test covering the public interface of this class.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+/**
+ * Copyright 2018 SURFnet B.V.
```
MKodde: ~~2018~~ 2019

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Domain\ValueObject;
+
+class Issue
```
MKodde: This VO should be JSON serializable, when saving this file to file in the issues.json file, it is saved as an empty object. Resulting in fatal errors on the user side.

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/IssueCollection.php
MKodde: Please import the Countable

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
MKodde: The actual issue data was not stored in the JSON file, only the key was. I added jsonSerialize to the Issue type, this did the trick.

pablothedude:
@MKodde thanks for addressing the raised concerns. The changes look good to me, the only concern I have is the functional aspect of the jira connection because I don't have a working Jira instance in my dev stack. So if that is tested it all looks good to me.

MKodde:
I don't have one running either, and I would have to go through a new registration procedure. I suggest testing this on the test environment before releasing this. @quartje can you live with that?

quartje:
Sure, I can live with that @MKodde

PR #264, opened 2019-06-04
Fix idp validation from manage result

pablothedude:
When testing on the test server it occurred that the `name:en` is
mandatory and `name:nl` optional. This will fix this so it would be
possible to edit acl's if idp's don't have a `name:nl` set.

https://www.pivotaltracker.com/story/show/165965483

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/UpdateEntityAclCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/UpdateEntityAclCommand.php
@@ -34,7 +34,7 @@ class UpdateEntityAclCommand implements Command
      * @var IdentityProvider[]
      * @Assert\All({
      *     @Assert\NotBlank(),
-     *     @Assert\Type(type="Surfnet\ServiceProviderDashboard\Domain\ValueObject\IdentityProvider")
+     *     @Assert\Type(type="Surfnet\ServiceProviderDashboard\Domain\Entity\IdentityProvider")
      * })
      */
     private $selected;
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/IdentityProvider.php
```diff
@@ -49,13 +49,13 @@ public function __construct($manageId, $entityId, $nameNl, $nameEn)
     {
         Assert::stringNotEmpty($manageId);
         Assert::stringNotEmpty($entityId);
-        Assert::string($nameNl);
+        Assert::nullOrString($nameNl);
         Assert::string($nameEn);
```
MKodde: Not sure if this is an actual business rule, but the nameEN is always set, so we can use stringNotEmpty on the $nameEn parameter here right?

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/UpdateEntityAclCommand.php
```diff
@@ -34,7 +34,7 @@ class UpdateEntityAclCommand implements Command
      * @var IdentityProvider[]
      * @Assert\All({
      *     @Assert\NotBlank(),
-     *     @Assert\Type(type="Surfnet\ServiceProviderDashboard\Domain\ValueObject\IdentityProvider")
+     *     @Assert\Type(type="Surfnet\ServiceProviderDashboard\Domain\Entity\IdentityProvider")
```
MKodde: Good catch, strange this did not result in issues during my test session.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/IdentityProvider.php
pablothedude: Thanks, double checked it in manage and that seems a valid assumption so I'll change this.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/UpdateEntityAclCommand.php
pablothedude: Probably because of the initially missing `isValid` check.

MKodde:
:ship: - it

PR #267, opened 2019-06-12
Do not save a local duplicate upon client reset

MKodde:
Whenever a published OIDC entity was using the client reset function, a duplicate local entity was saved to the database. This entity could not be removed from the application, as it will attempt to removed from Manage (as the production state is published). To fix this, the manage id had to be set on the local copy that is only temporarily used to carry the new client id secret.

In addition to this bugfix, a visual regression was fixed as well in 40570fc.

See: https://www.pivotaltracker.com/story/show/165673435

### DIFF
--- a/tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
+++ b/tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
@@ -1,0 +1,160 @@
+<?php
+
+/**
+ * Copyright 2019 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Tests\Integration\Application\CommandHandler\Entity;
+
+use Mockery as m;
+use Mockery\Adapter\Phpunit\MockeryTestCase;
+use Mockery\Mock;
+use Surfnet\ServiceProviderDashboard\Application\Command\Entity\ResetOidcSecretCommand;
+use Surfnet\ServiceProviderDashboard\Application\CommandHandler\Entity\ResetOidcSecretCommandHandler;
+use Surfnet\ServiceProviderDashboard\Application\Exception\EntityNotFoundException;
+use Surfnet\ServiceProviderDashboard\Application\Service\LoadEntityService;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
+use Surfnet\ServiceProviderDashboard\Domain\Repository\EntityRepository;
+
+class ResetOidcSecretCommandHandlerTest extends MockeryTestCase
+{
+    /**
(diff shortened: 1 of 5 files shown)


tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
```diff
@@ -0,0 +1,155 @@
...
+            ->with($entity);
+
+        $this->commandHandler->handle($command);
+
+        $this->assertEquals('my-manage-id', $entity->getManageId());
+    }
+
+    public function test_handle_entity_not_found()
+    {
+        $this->expectException(EntityNotFoundException::class);
```
BackEndTea: This should be as close as possible to where the exception occurs. (My guess is that it happens in the `commandHandler::handle` method). 

This makes sure the exception happens when it is expected to happen, and not earlier as a fault during set up for example.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
```diff
@@ -0,0 +1,155 @@
...
+                $command->getEnvironment()
+            )
+            ->andReturn(null);
+
+        $this->commandHandler->handle($command);
+    }
+
+    public function test_handle_entity_invalid_protocol()
+    {
+        $this->expectException(EntityNotFoundException::class);
```
BackEndTea: The previous comment applies here as well.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+        $entity = $this->expectedEntityFrom($command, $status);
+        $entity->setProtocol('oauth');
+
+        $this->commandHandler->handle($command);
+    }
+
+    public function test_handle_entity_invalid_publication_status()
+    {
+        $this->expectException(EntityNotFoundException::class);
```
BackEndTea: The previous comment applies here as well.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @return ResetOidcSecretCommand
+     */
+    private function buildCommand($id, $manageId, $environment)
+    {
+        return new ResetOidcSecretCommand($id, $manageId, $environment, m::mock(Service::class));
+    }
+
+    private function expectedEntityFrom(ResetOidcSecretCommand $command, $status)
```
BackEndTea: This can probably do with a return  typehint in the doc blocks as well.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+        $entity = $this->expectedEntityFrom($command, $status);
+
+        $this->repository
+            ->shouldReceive('save')
+            ->with($entity);
+
+        $this->commandHandler->handle($command);
+
+        $this->assertEquals('my-manage-id', $entity->getManageId());
```
BackEndTea: `assertSame` is safer than `assertEquals` (basically `===` instead of `==`), but i'm not sure if there is a standard for that in this project.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
MKodde: Makes sense, changing this for the test methods that expect an exception.

tests/integration/Application/CommandHandler/Entity/ResetOidcSecretCommandHandlerTest.php
MKodde: We use assetEquals by default. Moving to assertSame might be a good idea. For now, leaving this as is.

PR #270, opened 2019-06-13
Consider IdP workflow state when retrieving IdP's for ACL action

MKodde:
The `prodaccepted` state was used by default when retrieving IdP entities for display on the ACL screen. Now, we use the configured workflow state to filter the IdP results on.

This PR starts off with some housekeeping. The first commit (390213b) updates the way the workflow is configured. Previously the workflow state could be set for different intentions (create/update). This is not the way we use this feature. Instead of using this approach, setting a single workflow state per environment is now implemented.

The second commit then proceeds to read the configured workflow state from the manage configuration. This configuration is injected on the IdPclient.

See: https://www.pivotaltracker.com/story/show/165965483

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
@@ -161,7 +161,7 @@ class EntityService implements EntityServiceInterface
 
         $testEntities = $this->findPublishedTestEntitiesByTeamName(
             $service->getTeamName(),
-            $this->testManageConfig->getPublicationStatus()->getCreateStatus()
+            $this->testManageConfig->getPublicationStatus()->getStatus()
         );
         foreach ($testEntities as $result) {
             $entities[] = ViewObject\Entity::fromManageTestResult($result, $this->router, $service->getId());
@@ -169,7 +169,7 @@ class EntityService implements EntityServiceInterface
 
         $productionEntities = $this->findPublishedProductionEntitiesByTeamName(
             $service->getTeamName(),
-            $this->prodManageConfig->getPublicationStatus()->getCreateStatus()
+            $this->prodManageConfig->getPublicationStatus()->getStatus()
         );
         foreach ($productionEntities as $result) {
             $entities[] = ViewObject\Entity::fromManageProductionResult($result, $this->router, $service->getId());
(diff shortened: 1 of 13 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
```diff
@@ -161,15 +161,15 @@ public function getEntityListForService(Service $service)
 
         $testEntities = $this->findPublishedTestEntitiesByTeamName(
             $service->getTeamName(),
-            $this->testManageConfig->getPublicationStatus()->getCreateStatus()
```
BackEndTea: It would be cleaner to use a more explicit function here, instead of casting to string.
Even a (non magic) `toString` method would be beneficial. (Though i think within the domain `status(): string` may be appropriate.

This may help prevent bugs in case `getPublicationStatus()` may become nullable for example. In that case casting it to string will still 'work' but cause harder to debug issues. But if an explicit method call is done, then it will show exactly where the error is happening.

tests/unit/Infrastructure/Manage/Client/IdentityProviderClientTest.php
```diff
@@ -38,15 +41,22 @@ class IdentityProviderClientTest extends MockeryTestCase
      */
     private $mockHandler;
 
+    /**
+     * @var Config|Mock
```
BackEndTea: Technically this would be `Config&Mock`, but that syntax isn't used very widely yet.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Client/IdentityProviderClient.php
```diff
@@ -44,6 +44,7 @@ class IdentityProviderClient implements IdentityProviderRepository
 
     /**
      * @param HttpClient $client
+     * @param Config $manageConfig
```
BackEndTea: I think this doc block can be removed, as it does not add anything of value

tests/unit/Infrastructure/Manage/Client/IdentityProviderClientTest.php
MKodde: Cool, I certainly was not aware we can use an ampersand here. The more you know.

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
MKodde: Yes, a good suggestion. I'm replacing the __toString with a regular getter.

PR #272, opened 2019-06-19
Import correct invalid argument and throw it

BackEndTea:
During a migration an incorrect invalidArgument was imported, resulting
in a fatal error if a non string was passed into the method.

The error was also never thrown, which is now also fixed


MKodde:
Thanks @BackEndTea 

As you mentioned, phpstan found more problems in this realm. I'm adding a ticket to the backlog to address this. And with a recommendation to add the phpstan tool to the current set of testing tools.

PR #273, opened 2019-06-20
Ensure the issues are indexed on Manage id

MKodde:
Instead of indexing the Jira issues on the corresponding manage id, they were indexed on Jira key. This caused the status not being updated.

This is hard to test in a development environment, as we use a tests stand in for Jira interaction. And the issue did not surface in that solution.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/IssueCollection.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/IssueCollection.php
@@ -32,18 +32,18 @@ class IssueCollection implements Countable
      */
     public function __construct(array $issues)
     {
-        foreach ($issues as $issue) {
-            $this->issues[$issue->getKey()] = $issue;
+        foreach ($issues as $id => $issue) {
+            $this->issues[$id] = $issue;
         }
     }
 
     /**
      * @return Issue|null
      */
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/IssueCollection.php
```diff
@@ -32,18 +32,18 @@ class IssueCollection implements Countable
...
 
     /**
      * @return Issue|null
      */
-    public function getIssueByKey($key)
+    public function getIssueById($id)
     {
-        if (array_key_exists($key, $this->issues)) {
-            return $this->issues[$key];
+        if (array_key_exists($id, $this->issues)) {
```
BackEndTea: Small performance optimization possible here, `isset` is a lot faster then `array_key_exists`, unless its on php 7.4 and uses the FQN (either through use function, or by adding a \ before the function name)

MKodde: Thanks, I'm aware of that, not going to change it in this context. @pablothedude and I prefer the more readable `array_key_exists` over `isset`. And the performance penalty in this context is so irrelevant that I'm keeping it as is.

PR #276, opened 2019-07-03
Fix the entity delete acl

pablothedude:
The voter used on the entity delete endpoint didn't work as expected
for the administrator role because the role was validated against a
string and not a Role.

Also the endpoint didn't take the result into account from the isGranted
method and the isGranted method din't throw an exception either. Both
issues are now fixed.

https://www.pivotaltracker.com/story/show/167062944


pablothedude:
Added an integration test

PR #277, opened 2019-07-03
Fix the translations acl

pablothedude:
The translation endpoint could be reached if a user was logged in
and wasn't restricted to Administrators only.

The access_control is therefore changed and a change in a voter was
needed to allow validation of the endpoint.

https://www.pivotaltracker.com/story/show/167063760

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Security/Voter/ManageEntityAccessGrantedVoter.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Security/Voter/ManageEntityAccessGrantedVoter.php
@@ -38,10 +38,15 @@ class ManageEntityAccessGrantedVoter extends Voter
 
     protected function supports($attribute, $subject)
     {
-        $supportedAttribute = $attribute === self::MANAGE_ENTITY_ACCESS;
-        $subjectFieldsPresent = isset($subject['manageId']) && isset($subject['environment']);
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Security/Voter/ManageEntityAccessGrantedVoter.php
```diff
@@ -38,10 +38,15 @@ public function __construct(EntityServiceInterface $entityService)
 
     protected function supports($attribute, $subject)
     {
-        $supportedAttribute = $attribute === self::MANAGE_ENTITY_ACCESS;
-        $subjectFieldsPresent = isset($subject['manageId']) && isset($subject['environment']);
+        if ($attribute !== self::MANAGE_ENTITY_ACCESS) {
```
MKodde: This change had me confused for a minute. You wrote in your commit message this file needed a change, and it did. I only later figured out that the subject parameter was causing you difficulties. What I personally would have done, was input validate the subject parameter. If it is instance of `Request`, then this voter is not supported. 

But I can live with this solution too.

pablothedude: I've looked into this and considered using a VO for the subject to be able to validate it much more reliable but I thought that would cost me too much effort for now. So I kept the rest as-is.

PR #278, opened 2019-07-03
Add additional secyurity headers to response

pablothedude:
It was suggested to add more security headers to the responses by Cure53.
Those are now added.

https://www.pivotaltracker.com/story/show/167065629


pablothedude:
A good suggestion, I will look into this.

pablothedude:
Added the bundle and used the config from profile, the only thing which was changed was the use of the `Content-Security-Policy` configuration (see extra parameters in `forced_ssl` configuration section).

MKodde:
Thanks for adding the security bundle. One thing caught my eye, and that's the addition of the `spdashboard_saml_remote_idp_host` parameter. This change seems unrelated to this PR?

pablothedude:
The `spdashboard_saml_remote_idp_host` parameter is needed to support external redirects form EB. Else I was unable to log in, the same was done with Profile I think for the same reason.

PR #281, opened 2019-07-08
Escape the wysiwyg translations

pablothedude:
The lexik translation bundle  doesn't input validate the wysiwyg
entries. Because it's third party software it's hard to get it into
the bundle. Therefore we decided to validate the output.

To do so a wysiwyg filter is introduced and the documentation is
updated accordingly.

Also the ServiceStatusAssembler needed to be updated in order to
escape the json returned to render the service status donut.

---
Boyscouted:
Add ajax.googleapis.com to CSP whitelist to load external assets
for the Lexik translation bundle.

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -60,7 +60,8 @@ nelmio_security:
             - '%saml_remote_idp_host%'
     csp:
         report_logger_service: monolog.logger.security
-        hosts: []
+        hosts:
+            - 'ajax.googleapis.com'
         content_types: []
         enforce:
             report-uri: [/csp/report]
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -12,5 +12,5 @@ To import the default translations into the lexik database, run the
 deployment.
 
 Translation tokens ending with `.html` will be loaded in a WYSIWYG editor in the translation interface. These
-translations are also to be outputted unescaped in the template. To do so in twig use the raw filter. For example:
-`{{ 'service.edit.comments.html'|trans|raw }}` 
+translations are stored without input validation or sanitization. Therefore we need to strip the allowed tags in the
+template. To do so in twig use the wysiwyg filter. For example: `{{ 'service.edit.comments.html'|trans|wysiwyg }}` 
--- a/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
@@ -19,7 +19,7 @@ namespace Surfnet\ServiceProviderDashboard\Application\Assembler;
 
 use Surfnet\ServiceProviderDashboard\Application\Dto\ServiceStatusDto;
 use Surfnet\ServiceProviderDashboard\Application\Service\ServiceStatusService;
-use Surfnet\ServiceProviderDashboard\Application\ViewObject\EntityList;
+use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Twig\WysiwygExtension;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
 use Symfony\Component\Translation\TranslatorInterface;
 
@@ -226,7 +226,7 @@ class ServiceStatusAssembler
     {
         $labels = [];
         foreach (self::states() as $state) {
-            $labels[$state] = $this->translator->trans('service.overview.progress.label.' . $state);
+            $labels[$state] = $this->getSanitizedHtmlTranslation('service.overview.progress.label.' . $state);
         }
         return $labels;
     }
@@ -235,7 +235,7 @@ class ServiceStatusAssembler
     {
         $tooltips = [];
         foreach ($mappedStates as $state => $status) {
-            $tooltips[$state] = $this->translator->trans('service.overview.progress.tooltip.' . $state . '.' . $status . '.html');
+            $tooltips[$state] = $this->getSanitizedHtmlTranslation('service.overview.progress.tooltip.' . $state . '.' . $status . '.html');
         }
         return $tooltips;
     }
@@ -251,5 +251,12 @@ class ServiceStatusAssembler
             $total++;
         }
         return \round($done/$total*100);
+    }
+
+    private function getSanitizedHtmlTranslation($key)
+    {
+        $translated = $this->translator->trans($key);
+        $sanitized = WysiwygExtension::sanitizeWysiwyg($translated);
+        return $sanitized;
     }
 }
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityModal/secretResetModal.html.twig
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityModal/secretResetModal.html.twig
@@ -3,8 +3,8 @@
 {% block page_heading %} {{ 'entity.actions.modal.oidc_client_reset.title'|trans }} {% endblock %}
 
 {% block body %}
-    <div class="warning"><i class="fa fa-exclamation-triangle"></i>{{ 'entity.actions.modal.oidc_client_reset.warning'|trans|raw }}</div>
-    <div class="wysiwyg">{{ 'entity.actions.modal.oidc_client_reset.info.html'|trans|raw }}</div>
+    <div class="warning"><i class="fa fa-exclamation-triangle"></i>{{ 'entity.actions.modal.oidc_client_reset.warning'|trans }}</div>
+    <div class="wysiwyg">{{ 'entity.actions.modal.oidc_client_reset.info.html'|trans|wysiwyg }}</div>
 
     <div class="row">
         <div class="button-row">
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
@@ -3,8 +3,8 @@
 {% block page_heading %} {{ 'entity.list.modal.oidc_confirmation.title'|trans }} {% endblock %}
 
 {% block body %}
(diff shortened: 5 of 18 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
```diff
@@ -3,8 +3,8 @@
 {% block page_heading %} {{ 'entity.list.modal.oidc_confirmation.title'|trans }} {% endblock %}
 
 {% block body %}
-    <div class="warning"><i class="fa fa-exclamation-triangle"></i>{{ 'entity.list.modal.oidc_confirmation.warning'|trans|raw }}</div>
-    <div class="wysiwyg">{{ 'entity.list.modal.oidc_confirmation.info.html'|trans|raw }}</div>
+    <div class="warning"><i class="fa fa-exclamation-triangle"></i>{{ 'entity.list.modal.oidc_confirmation.warning'|trans|wysiwyg }}</div>
```
MKodde: I don't think filtering the output is warranted here. The original input is not filled using the wysiwyg editor.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Twig;
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+
+class WysiwygExtension extends AbstractExtension
+{
+
+    private static $allowedTags = '<p><em><strong><span><h1><h2><h3><h4><h5><h6><ul><ol><li><a><sup><sub><code><blockquote>';
```
MKodde: Good idea to make a custom extension! I'd love to see the allowed tags being configurable in the parameters.json or elsewhere. If we keep them here, please update the readme to instruct users of this feature. Must we consider attribute sanitization? As this is another valid attack vector in this scenario.

For example we now allow the use of span tags. They are filled with inline css to create strikethroughts and underlines. A possible attacker can now also run javascript from the attribute as these are not sanitized.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -215,7 +215,7 @@ private function getLegend()
         $legend = [];
         foreach (self::status() as $name) {
             $legend[$name] = [
-                'label' => $this->translator->trans('service.overview.legend.'.$name),
+                'label' => $this->getSanitizedTranslation('service.overview.legend.'.$name),
```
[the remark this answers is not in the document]
pablothedude: I think it will be valuable in terms of defense in depth.

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
```diff
@@ -252,4 +252,11 @@ private function getPercentage($mappedStates)
         }
         return \round($done/$total*100);
     }
+
+    private function getSanitizedTranslation($key)
+    {
+        $translated = $this->translator->trans($key);
+        $sanitized = WysiwygExtension::sanitizeWysiwyg($translated, '<p><em><strong><span><h1><h2><h3><h4><h5><h6><ul><ol><li><a><sup><sub><code><blockquote>');
```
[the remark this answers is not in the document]
pablothedude: The second parameter was already omitted but still had to be removed here. Good catch!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityModal/secretResetModal.html.twig
```diff
@@ -4,7 +4,7 @@
 
 {% block body %}
     <div class="warning"><i class="fa fa-exclamation-triangle"></i>{{ 'entity.actions.modal.oidc_client_reset.warning'|trans|raw }}</div>
```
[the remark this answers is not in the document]
pablothedude: Good suggestion

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
pablothedude: Was indeed not necessary so removed it

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
pablothedude: I didn't found it that valuable because it won't be easily configurable due to the Tinymce configuration itself.

app/config/security.yml
```diff
@@ -60,7 +60,8 @@ nelmio_security:
             - '%saml_remote_idp_host%'
     csp:
         report_logger_service: monolog.logger.security
-        hosts: []
+        hosts:
+            - 'ajax.googleapis.com'
```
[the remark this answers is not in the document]
pablothedude: I left it as-is because that wan't a real advisory so I won't fix it here but maybe we could address the issue to Bart?

docs/translations.md
```diff
@@ -12,5 +12,5 @@ To import the default translations into the lexik database, run the
 deployment.
 
 Translation tokens ending with `.html` will be loaded in a WYSIWYG editor in the translation interface. These
-translations are also to be outputted unescaped in the template. To do so in twig use the raw filter. For example:
-`{{ 'service.edit.comments.html'|trans|raw }}` 
\ No newline at end of file
+translations are stored without input validation. Therefore we need to escape the allowed tags in the template.
```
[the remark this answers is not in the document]
pablothedude: Updated

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
MKodde: FYI: `entity.list.modal.oidc_confirmation.info.html` should have the wysiwyg filtering as this can be loaded with HTML content. I was commenting on the `entity.list.modal.oidc_confirmation.warning` translation. Which is a pure text translation that does not benefit from html output sanitisation.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
```diff
@@ -0,0 +1,74 @@
...
+     * @var HTMLPurifier
+     */
+    private static $purifier;
+
+    public function __construct()
+    {
+        $config = HTMLPurifier_Config::createDefault();
+
+        $config->set('Core.Encoding', 'UTF-8'); // replace with your encoding
+        $config->set('HTML.Doctype', 'XHTML 1.0 Transitional'); // replace with your doctype
```
MKodde: Our doctype is set to `html` (5). That s(c)ould be reflected in this config?!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
```diff
@@ -0,0 +1,74 @@
...
+class WysiwygExtension extends AbstractExtension
+{
+    /**
+     * @var HTMLPurifier
+     */
+    private static $purifier;
+
+    public function __construct()
+    {
+        $config = HTMLPurifier_Config::createDefault();
```
MKodde: Using the HTML purifier :+1:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/oidcConfirmationModal.html.twig
pablothedude: I removed the wrong filter, fixed

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
pablothedude: Unfortunately not, http://htmlpurifier.org/live/configdoc/plain.html#HTML.Doctype

src/Surfnet/ServiceProviderDashboard/Application/Assembler/ServiceStatusAssembler.php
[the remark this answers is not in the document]
pablothedude: I don't think it would have done any damage to kept it as is but I'm fine to remove it from the legends because I did some additional testing if the legends were thoroughly escaped and they were.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/WysiwygExtension.php
pablothedude: I don't think it would do any damage either with the used subset of allowed html.

MKodde: Would not setting it be an option?

MKodde: And maybe remove the `// replace with your ..` comments?

pablothedude: Removed the comments and used a specialized html5 htmlpurifier.

PR #284, opened 2019-07-11
Disable the htmlpurifier cache

pablothedude:
This is done in order to prevent warning because it will try
to write to `/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer`
which will generate a warning on production.

https://www.pivotaltracker.com/story/show/167222617


pablothedude:
Personally I didn't notice degrading performance and I consulted Bart if it would be an issue to disable the cache for now. This was acceptable. If we run into degraded performance we could enable the cache at a later moment.

PR #285, opened 2019-07-11
Prevent error when loading an empty issue mock file

pablothedude:
When loading an empty issue mock file to get a working implementation without the use of a working Jira instance, a mock file was used.

If this file was empty an exception was thrown because the data was then not parsed to an empty array but to null which triggered input validation. This is now fixed.


tests/unit/Infrastructure/DashboardBundle/Jira/Repository/IssueRepositoryTest.php
```diff
@@ -109,6 +109,13 @@ public function test_repository_flow()
         $this->assertSame(4, $found);
     }
 
+    public function test_empty_file_should_not_throw_an_exception()
```
MKodde: :+1: for adding the test. 

Off-topic: The class name of the test had me confused for a second, this should be DevelopmentIssueRepositoryTest. Let's fix this some other time :)

PR #287, opened 2019-07-16
Make Jira test mode configurable

MKodde:
We sometimes do not have a working Jira configuration at the ready. In those cases the test stand-in is a useful fallback measure. Previously the test and development environment by default used this stand-in. This has now been changed to become a feature flag.

This flag can be set to true/false in the parameters.yml

The filePath of the test file has also been made into a configuration option while at it.

https://www.pivotaltracker.com/story/show/167315333

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * Copyright 2019 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\DependencyInjection\Compiler;
+
+use Exception;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Factory\IssueFieldFactory;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Factory\JiraServiceFactory;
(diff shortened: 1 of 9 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
```diff
@@ -0,0 +1,95 @@
...
+        if (!$hasParameter || !$hasDefinition) {
+            return;
+        }
+
+        $isTestModeEnabled = (bool) $container->getParameter(self::ENABLE_TEST_MODE_FEATURE_FLAG);
+
+        $kernelEnv = $container->getParameter('kernel.environment');
+
+        // Web tests, ironically for now do not utilize the test stand-in.
+        if ($kernelEnv === 'test' && $isTestModeEnabled) {
```
pablothedude: I think this is kind of magic to override the feature flag based on environment. Why not set this in `config_test.yml`?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
```diff
@@ -0,0 +1,95 @@
...
+
+use Exception;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Factory\IssueFieldFactory;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Factory\JiraServiceFactory;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Repository\DevelopmentIssueRepository;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Jira\Repository\IssueRepository;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class IssueRepositoryCompilerPass implements CompilerPassInterface
```
pablothedude: Personally I don't like the introduced magic by using a CompilerPass because it's interferes with the regular DI implementation. but in this case it gets the job done.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
MKodde: The compiler pass would then override it again. I've tried this.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DependencyInjection/Compiler/IssueRepositoryCompilerPass.php
MKodde: I'm no big fan of this either. It's the only way I know of, that can provide this feature.

PR #290, opened 2019-08-21
Request EPTI by default for each OIDC TNG entity

MKodde:
The EduPersonTargettedID attribute is always requested for oidcng
entities. The command tasked with saving the entities enables the epti
attribute by default.

See: https://www.pivotaltracker.com/story/show/167511328

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
@@ -370,7 +370,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
 
     /**
      * @param Service $service
-     * @return SaveOidcEntityCommand
+     * @return SaveOidcngEntityCommand
      */
     public static function forCreateAction(Service $service)
     {
@@ -383,7 +383,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
     /**
      * @param Entity $entity
      *
-     * @return SaveOidcEntityCommand
+     * @return SaveOidcngEntityCommand
      */
     public static function fromEntity(Entity $entity)
     {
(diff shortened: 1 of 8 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -363,6 +363,12 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
 
     private function __construct()
     {
+        // Set/enable the EPTI attribute by default as required by: https://www.pivotaltracker.com/story/show/167511328
+        $epti = new Attribute();
+        $epti->setMotivation('OIDC requires EduPersonTargetedID by default');
+        $epti->setRequested(true);
+
+        $this->eduPersonTargetedIDAttribute = $epti;
```
pablothedude: Why not drop the EduPersonTargeId altogether for OIDCng and put this in the infrastructure layer, so this is only used when an OIDCng entity is written to Manage it will clean things up becasue of YAGNI and will prevent the use of something which shouldn't be used at all?

MKodde: Solving business rules in the infra-layer doesn't seem wise to me, as it clutters the handlers/helpers and will require domain logic knowledge to understand why a certain value is added for a given situation. I've considered doing this but it is not something I'm terribly thrilled about. Maybe we should talk about this IRL to make the discussion more efficient.

MKodde: I did add an integration test to cover the logic added to the application layer. Thanks for suggesting that!

pablothedude: Thanks for adding the tests!

Regarding the EPTI I highly doubt it's domain logic. It's not required for OIDCng but only to keep Manage happy. And that's in the context of EB the infrastructure layer. So why support that throughout all layers if it's only required by Manage?

MKodde: I see your point, and mostly agree. I'm now adding the EPTI attribute right before we generate the Manage ARP data that is sent to Manage (prod/test). The test suites have been updated accordingly.

PR #291, opened 2019-08-21
Do not allow duplicate redirect URIs

MKodde:
Both the front- and backend did not check for duplicate entries. This PR adds both. For more information about this feature and some back and forth between me and the PO see:

https://www.pivotaltracker.com/story/show/167511283


MKodde:
> Why not use the assert annotations to validate the command as this would be more intuitive?

In our version of Symfony it's not yet possible to solve this without a callback.

> I think it would be even better to validate this in the entity as this is domain logic.

That would be tricky as we want to verify this rule during validation time. Could be fixed by moving this verification logic into the `domain layer` Entity entity, and call it from the `application layer` command. But that would raise the next question, is performing this validation logic the responsibility of the entity?

pablothedude:
> > Why not use the assert annotations to validate the command as this would be more intuitive?
> 
> In our version of [Symfony](https://symfony.com/doc/current/reference/constraints/Unique.html) it's not yet possible to solve this without a callback.
>
But normally a custom validator is used and those are reusable and better testable: see for example `@SpDashboardAssert`.

> > I think it would be even better to validate this in the entity as this is domain logic.
> 
> That would be tricky as we want to verify this rule during validation time. Could be fixed by moving this verification logic into the `domain layer` Entity entity, and call it from the `application layer` command. But that would raise the next question, is performing this validation logic the responsibility of the entity?

Could you elaborate why validation in your domain would be tricky? I think that the validation in the domain should be there to validate consistency in your domain layer and thus throughout the entire application. If you cannot have duplicates in an entity then it shouldn't be possible. The validation in the command is there to validate input and to inform the end user better so that's a different responsibility.

MKodde:
1. That's a good idea. But moves the validation from the application layer to the infra layer. 
2. Validation takes place when the form is saved for publication. Symfony will reload the form if the validation criteria are not met, and present the user with meaningful feedback. Moving this into the domain layer would complicate things. As you have already passed the validation stage at that point.

pablothedude:
>     2. Validation takes place when the form is saved for publication. Symfony will reload the form if the validation criteria are not met, and present the user with meaningful feedback. Moving this into the domain layer would complicate things. As you have already passed the validation stage at that point.
I don't mean to move the validation away from the form into the domain but add minimal validation to keep the domain state consistent. This would have prevented bugs like these and will continue to prevent them in the future. It's something you could consider.

PR #294, opened 2019-08-22
Refrain from sending a NULL secret to Manage

MKodde:
Sending the client secret a second time (null) will cause Manage to spit out a 500 error. Easily remedied by testing if the client secret is set. If it is not, we no longer pass it to Manage API.

https://www.pivotaltracker.com/story/show/168049212


pablothedude:
Is the issue the same as the one described in the story because the `clientId` is mentioned and not the `secret`?
`The client id causes issues when saving a test oidcng entity. Manage refuses the entity for some reason.`

MKodde:
Thanks for noticing that! At first my assumption was that the client id was the field causing this issue. After closer inspection it turned out it was the secret field.

I've updated the bug report slightly to better reflect the change.

PR #295, opened 2019-08-27
Correctly display ClientID values

MKodde:
The Oidc and Oidc TNG entities have different client id conventions. The parsers should ensure we always create the correct client id. Where currently the conversion logic is implemented in different classes, this solution enables the reuse of the conversion logic.

Both the Oidc and Oidc TNG classes, using or displaying the ClientID now use the correct parser to do so.

See: https://www.pivotaltracker.com/story/show/167510409


MKodde:
Moved the documentation from the method to the class. and added some extra.

PR #299, opened 2019-08-28
Push signature method to manage for OIDC

pablothedude:
When pushing an OIDC entity to manage the signature method
wasn't pushed.

https://www.pivotaltracker.com/n/projects/1400064/stories/166838510

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
@@ -238,8 +238,10 @@ class JsonGenerator implements GeneratorInterface
             $metadata['AssertionConsumerService:0:Binding'] = $entity->getAcsBinding();
             $metadata['AssertionConsumerService:0:Location'] = $entity->getAcsLocation();
             $metadata['NameIDFormat'] = $entity->getNameIdFormat();
-            $metadata = array_merge($metadata, $this->generateSecurityMetadata($entity));
+            $metadata['coin:signature_method'] = 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256';
+            $metadata = array_merge($metadata, $this->generateCertDataMetadata($entity));
         } else if ($entity->getProtocol() == Entity::TYPE_OPENID_CONNECT) {
+            $metadata['coin:signature_method'] = 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256';
             $metadata["coin:oidc_client"] = '1';
         }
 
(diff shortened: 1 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -238,8 +238,10 @@ private function generateMetadataFields(Entity $entity)
             $metadata['AssertionConsumerService:0:Binding'] = $entity->getAcsBinding();
             $metadata['AssertionConsumerService:0:Location'] = $entity->getAcsLocation();
             $metadata['NameIDFormat'] = $entity->getNameIdFormat();
+            $metadata['coin:signature_method'] = 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256';
             $metadata = array_merge($metadata, $this->generateSecurityMetadata($entity));
         } else if ($entity->getProtocol() == Entity::TYPE_OPENID_CONNECT) {
+            $metadata['coin:signature_method'] = 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256';
```
MKodde: I think you can get away with calling the `generateSecurityMetadata` method for both protocol types. Added benefit of that is that the setting of the signature method remains 'responsibility' of that method.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/MetaData.php
```diff
@@ -25,7 +25,6 @@ class MetaData
     /**
      * Supports 1 ACS location, the first entry is used that is passed from Manage
      */
-    private $acsBinding;
```
MKodde: Please explain why you are removing the acsBinding field from the metadata DTO. I see it is not used. But seeing this removed in this PR deserves at least a mention in the PR description or your commit message.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
pablothedude: The `certData` is omitted in OIDC so therefore I don't want to push it to manage. However renamed the method to `generateSecurityMetadata` to reflect it's responsibility better.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/MetaData.php
pablothedude: Mentioned it in the commit message.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
MKodde: This is fine too. By calling the method you would have saved a couple of lines of code.

PR #300, opened 2019-09-03
Only allow publication of OIDC RS entities

MKodde:
Saving drafts will be deprecated in the near future. The OIDC TNG
Resource Server will be the first entity type where this is not allowed.


pablothedude:
Would you be so kind to enable travis on 'feature/oidcng-resource-server'? Tnx!

MKodde:
Why this PR previously passed amazes me! The removal of the `SaveOidcngResourceServerEntityCommandHandler` was not warranted. As a local entity is required to publish an entity to Manage. This will be one of the more challenging changes when we are removing the ORM layer altogether.

For now, I reverted the removal of the Handler, thus still saving a working copy in DB. This is cleaned up directly after publication.

pablothedude:
I've seen it before that the tests were successful when they should be failing. This was then because  the expected response from manage wouldn't result in a failure while being the result for an entirely different request. Maybe we could fix this also in the draft removal sprint?

PR #301, opened 2019-09-04
Allow empty contact type on the contact information from Manage

pablothedude:
The ContactType can be empty when fetching an entity from manage.

https://www.pivotaltracker.com/story/show/166232238


pablothedude:
Unsupported contact types won't be added anymore. When invalid contacts are overwritten during the next push that will be acceptable according to the PM.

Added a test to cover invalid or missing contacttypes.

PR #302, opened 2019-09-04
Allow empty commanname when authenticating

pablothedude:
The commonname could be empty


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
```diff
@@ -93,7 +93,12 @@ public function authenticate(TokenInterface $token)
 
         $nameId = $translatedAssertion->getNameID();
         $email = $this->getSingleStringValue('mail', $translatedAssertion);
-        $commonName = $this->getSingleStringValue('commonName', $translatedAssertion);
+
+        try {
+            $commonName = $this->getSingleStringValue('commonName', $translatedAssertion);
+        } catch (BadCredentialsException $e) {
```
MKodde: The exception thrown does not really reflect the issue at hand. But that should have been noted when this feature was implemented 2 years ago. If you feel like it, maybe add a new exception named something like UnknownAttributeRequestedException.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
```diff
@@ -93,7 +93,12 @@ public function authenticate(TokenInterface $token)
...
+            $commonName = '';
```
MKodde: Please add an info log entry for this situation. We are no longer logging the attribute not found/bad credentials error in the logs. But this situation does deserve a log mention in my opinion.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
pablothedude: Good suggestion, created a new event for this.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
pablothedude: Added logging for this fallback mechanism.

pablothedude:
Added a test, this was a bit harder then I did expect because it seemed there was almost nothing tested. So I've added some tests.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Exception/MissingSamlAttributeException.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet B.V.
```
MKodde: 2019

tests/unit/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProviderTest.php
```diff
@@ -22,13 +22,50 @@
...
+    /**
+     * @var m\MockInterface|AttributeDictionary
+     */
+    private $attributeDictionary;
+    /**
+     * @var m\MockInterface|LoggerInterface
+     */
+    private $logger;
+
+    protected function setUp()
```
MKodde: Thanks for adding the test coverage here!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
MKodde: Nice upgrade!

PR #303, opened 2019-09-06
Prevent form submission on enter

pablothedude:
When pushing enter when filling redirect uris' the form gets submitted
while someone could expect only the redirect uri entry would be added.

https://www.pivotaltracker.com/story/show/168192455


app/js/components/collection_widget.test.ts
```diff
@@ -1,6 +1,7 @@
 import { loadEntityOidcForm } from "./collection_widget";
 
 import * as $ from "jquery";
+import * as jQuery from "jquery";
```
MKodde: Have you considered replacing the `$` usage with `jQuery`? By the way, I have zero tension with your solution but was triggered by the duplicate import under different aliases. Leaving it like this, especially because this is a test, if perfectly fine with me.

By the way, shouldn't the Encore `autoProvidejQuery` option not provide jQuery capabilities in this script in the first place?

app/js/components/collection_widget.ts
```diff
@@ -141,15 +156,15 @@ class CollectionWidget {
   }
 
   private isUnique(newElement: JQuery<HTMLElement>): boolean {
-    let isUniqe = true;
+    let isUnique = true;
```
MKodde: :+1: for fixing the typo

app/js/components/collection_widget.ts
```diff
@@ -125,6 +126,20 @@ class CollectionWidget {
     $addEntryButton.on('click', handleAddClick);
   }
 
+  /**
+   * Add click handler to add entry
+   * @param $addEntryButton
+   */
+  private registerKeyDownEnterHandler($input: JQuery<HTMLElement>) {
+    const handleKeydownEnter = (event: JQuery.Event) => {
+      if (event.keyCode === 13) {
```
MKodde: Use of 'keyCode' property is deprecated. Please update, and use the 'which' prop.

MKodde: Reading into this, the which property is also deprecated from the [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) standard. Might be updated soon in Jquery too :confused:

app/js/components/collection_widget.ts
```diff
@@ -125,6 +126,20 @@ class CollectionWidget {
     $addEntryButton.on('click', handleAddClick);
   }
 
+  /**
+   * Add click handler to add entry
```
MKodde: Technically this is not a click handler

app/js/components/collection_widget.test.ts
pablothedude: Yes I've considered that, but I chose to use the `$` only for selectors and `jQuery` to use for objects in the library itself, in this case for a jQuery Event (`jQuery.Event`).

app/js/components/collection_widget.ts
pablothedude: :+1:

app/js/components/collection_widget.ts
pablothedude: Used the `key` property, hopefully that property will last a bit longer.

app/js/components/collection_widget.ts
pablothedude: Changed the name to describe it's usage better.

pablothedude:
Asked @quartje 's opiionion on the default `onenter` behaviour:
https://www.pivotaltracker.com/story/show/168192455/comments/206460645

pablothedude:
Added a change to prevent submitting a form when pressing 'Enter' on input fields.

PR #304, opened 2019-09-09
Set authorization code as default grant type

MKodde:
As requested in: https://www.pivotaltracker.com/story/show/167511241/comments/206159247


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -106,11 +106,11 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
     private $isPublicClient;
 
     /**
-     * @var OidcGrantType
+     * @var string $grantType defaults to OidcGrantType::GRANT_TYPE_AUTHORIZATION_CODE
```
pablothedude: Maybe not too necessary to document the default, but it's not  troublesome either

PR #305, opened 2019-09-10
Configure relation between OIDC Clients and Resource Servers

MKodde:
The client form now renders the available resource servers in a checkbox group. The selected RS entities are registered on the client entity in Manage. But are also stored in the draft environment of SPD before publishing. So, while testing, do not forget to run the migration that is part of this PR.

https://www.pivotaltracker.com/story/show/167530069


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngResourceServerOptionsFactory.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->productionRepository = $productionRepository;
+        $this->testPublicationState = $testPublicationState;
+        $this->productionPublicationState = $productionPublicationState;
+    }
+
+
+    /**
+     * @param string $teamName
+     * @param string $environment
+     * @return array
```
pablothedude: Why not return a `ResourceServerCollection`?

tests/webtests/EntityCreateOidcngTest.php
```diff
@@ -38,6 +38,8 @@ public function setUp()
 
     public function test_it_renders_the_form()
     {
+        // Manage is queried for resource servers
+        $this->testMockHandler->append(new Response(200, [], '[]'));
```
pablothedude: Is this boyscouting or is this really needed to support OidcRs?

pablothedude: Same for all changes in this file....

MKodde:
Re 3: Got it! Fixing this for the other OIDC entities as I go

MKodde:
Making the flash message more verbose as suggested in #306 will be addressed in a separate PR.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngResourceServerOptionsFactory.php
pablothedude: I think it would be better if that isn't the responsibility of the factory. The consumer would be able to use the `ResourceServerCollection::getResourceServers` method to get an array while the factory will be reusable by other consumer later on. Else you would have to create a factory dependent on used consumers. I don't mind if you leave it for now, but you could consider returning a VO next time.

tests/webtests/EntityCreateOidcngTest.php
pablothedude: I notice that sometimes one request is needed and sometimes multiple. Also some lines are documented and some are not so it's not getting any clearer at the moment. And when an empty array with status 200 is expected in multiple different cases how valuable are the functional tests at the moment? Wouldn't it be better to fix it a.s.a.p. as they are currently not representative enough and would have a negative impact on the quality. I think we should think about when to address this.

pablothedude: I saw that there was already a task on the drafts cleanup story to address this when cleaning the draft logic. I have created a new story for this to keep the changes more atomic.

[Drafts cleanup story](https://www.pivotaltracker.com/story/show/166004131)
[New fix mock requests story](https://www.pivotaltracker.com/story/show/168513312)

MKodde:
As requested IRL, this is not a OIDC resource server specific issue. I proposed to fix this on a separate branch. Thanks for the report and the extended testing!

tests/webtests/EntityCreateOidcngTest.php
MKodde: As for your first question:

> ... how valuable are the functional tests at the moment? ...

At the time we made the decision to NOT test Manage. So everywhere we can, we simply return the response of the least effort, yielding the desired functional response. The dedicated manage client tests should cover the correct working of that data layer. The throwback from this solution is that the web tests are rather fragile, and it's not always clear when/what is expected to be retrieved from Manage.

My proposal is to not change this right away, but make it part of the bigger refactor effort we have planned (removal of the database layer). If you feel we should address this now, we must certainly plan this with the team. As this would be quite an effort. And should be preceded by a well thought out plan.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngResourceServerOptionsFactory.php
MKodde: I don't see this being reused. And if it would be, then it would most certainly be in the context of ChoiceType options generation (as this is the single responsibility of this factory), and that is what it would be able to do just fine. 

Also, introducing an additional complexity for the sake of possible reusability down the road is not something I consider valuable here.

Maybe you are aiming at something else here?

PR #309, opened 2019-09-16
Prepend the https protocol after production copy

MKodde:
Only for OIDC TNG entities (RP & RS) should this logic be applied.

https://www.pivotaltracker.com/story/show/168452265

### DIFF
--- a/tests/unit/Application/Service/LoadEntityServiceTest.php
+++ b/tests/unit/Application/Service/LoadEntityServiceTest.php
@@ -517,7 +517,55 @@ JSON
 
         $this->assertNotEmpty($entity->getOidcngResourceServers()->getResourceServers());
     }
-    
+
+    public function test_it_updates_the_client_id_on_copy_to_production()
+    {
+        $this->entityRepository->shouldReceive('isUnique')
+            ->with('dashboardid')
+            ->andReturn(true)
+            ->twice();
+
+        $manageDto = ManageEntity::fromApiResponse($this->getOidcNgRPMetadata());
+
+        $this->manageTestClient->shouldReceive('findByManageId')
+            ->with('manageid')
+            ->andReturn($manageDto);
+
+        $entity = $this->copyService->load(
+            'dashboardid',
+            'manageid',
+            $this->service,
+            Entity::ENVIRONMENT_TEST,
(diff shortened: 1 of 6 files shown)


pablothedude:
**Functional review**
If I change the uri in manage to 'https@//test.test' I can save the entity in manage but I wont be able to store the entity in SPD.
If I change the uri in manage to 'test' I can save the entity in manage but then the validation will also fail although 'http://' is prepended to the enityid but because it's lacking a tld.

MKodde:
@quartje does SPD need to be aware of the scenarios @pablothedude is drawing? And handle them in a graceful manner? Or are these considered edge cases we do not have to consider at this point?

tests/webtests/EntityCopyTest.php
```diff
@@ -153,13 +153,29 @@ public function test_copy_to_production_does_not_create_new_entity()
         );
         $this->testMockHandler->append(new Response(200, [], json_decode($response)));
 
-        $this->client->request('GET', "/entity/copy/d645ddf7-1246-4224-8e14-0d5c494fd9ad/production");
+        $this->client->request('GET', "/entity/copy/1/d645ddf7-1246-4224-8e14-0d5c494fd9ad/production");
```
pablothedude: I think it would be better to fix the test completely because now the `assertEmpty` is returning true on an empty array and results in a false positive.

MKodde: Yes, I agree

tests/unit/Application/Service/LoadEntityServiceTest.php
```diff
@@ -517,7 +517,56 @@ public function test_it_keeps_resource_servers_on_copy_to_same_environment()
 
         $this->assertNotEmpty($entity->getOidcngResourceServers()->getResourceServers());
     }
-    
+
+    public function test_it_updates_the_client_id_on_copy_to_production()
```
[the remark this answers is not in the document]
MKodde: 1b41ba2

tests/webtests/EntityCopyTest.php
pablothedude: :+1: thanks!

PR #310, opened 2019-09-17
Set default grant type value when empty argument provided

MKodde:
The grant type now converts an empty value to the default value. This is
required for pre transition draft entities. These would cause the
invalid arg exception being raised.

https://www.pivotaltracker.com/story/show/167511241/comments/206739007

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/OidcGrantType.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/OidcGrantType.php
@@ -39,6 +39,12 @@ class OidcGrantType
      */
     public function __construct($grantType = self::GRANT_TYPE_AUTHORIZATION_CODE)
     {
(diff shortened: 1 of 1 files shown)


MKodde:
Your suggestion was my initial solution. But I chose this (IMHO better) solution in favor of that. Problem with your suggested solution: you have to start relying on the invalid argument exception being thrown when an empty grant type is passed to the constructor. And in that case you want to fall back on the default value. But what if an actual invalid argument was passed? Would you then also set the default value? That would completely defeat the existing validation logic.

pablothedude:
> invalid

You shouldn't rely on that exception but only make sure the default is initially set. This can easily be done by setting the default value with DI in the constructor of the Command so no additional logic is needed.

MKodde:
That value comes from the database, so a fix on the infra level would be required to fix that (a migration). 

As it turns out, the OIDC tng feature is not yet landed on production. And so no draft entities there can 
have this problem. 

Closing..

PR #311, opened 2019-09-17
Prevent call to ResourceServerCollection function on null

MKodde:
Draft entities without a Rersource Server collection would error into a 500 status because the Doctrine entity did not have a suitable default value for the `oidcngResourceServers` field. This field should be set with a `ResourceServerCollection` instance. But this would be null. Causing an internal server error when the entity was converted to a save command.

See bug report in:
https://www.pivotaltracker.com/story/show/168532438

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
@@ -1596,6 +1596,10 @@ class Entity
      */
     public function getOidcngResourceServers()
     {
+        // Doctrine does not call the constructor, preventing us from setting a default value there.
+        if (is_null($this->oidcngResourceServers)) {
+            return new ResourceServerCollection([]);
+        }
         return $this->oidcngResourceServers;
     }
 


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity.php
```diff
@@ -1596,6 +1596,10 @@ public function isPublicClient()
      */
     public function getOidcngResourceServers()
     {
+        // Doctrine does not call the constructor, preventing us from setting a default value there.
+        if (is_null($this->oidcngResourceServers)) {
```
pablothedude: I think it would be more solid to set this on initialization and to prevent the usage of `null` this will prevent a lot of additional logic throughout the project.

MKodde: That, sadly was not possible, as Doctrine does not call the constructor of this class.

pablothedude: You could force a collection and ommit `null` add a migration to set a empty `ResourceServerCollection` where it currently is `null`. But feel free not to address this as this is only the case for NG entities and if the drafts are cleaned up this won't be an issue at all.

MKodde: A migration would have been another good solution. But as it turns out, this problem would never be an issue in production, as OIDCng is not yet available there. Thanks for keeping quality high by pointing out these kind of issues!

I'm closing this issue.

PR #313, opened 2019-10-03
Persist IdP whitelist settings

MKodde:
When working with an entity, the whitelist info is loaded from Manage, this data would not be saved to the entity in SPD except for in the explicit IdP whitelist edit screen.

In other situations (entity publish and edit actions) this information was lost. By storing the whitelist information on the form in hidden form fields, we ensure the data is persisted when publishing the metadata changes.

The IdP metadata is serialized to a JSON encoded string.

Providing a meaningful integration/acceptance test for this bug was not possible. As we do not actually test manage interactions.

https://www.pivotaltracker.com/story/show/168800119

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcEntityCommand.php
@@ -343,6 +343,19 @@ class SaveOidcEntityCommand implements SaveEntityCommandInterface
      */
     private $manageId;
 
+    /**
+     * @var bool
+     */
+    private $idpAllowAll = true;
+
+    /**
+     * The array of IdP's that are whitelisted are displayed in a hidden form field, and are rendered
+     * as a string (json encoded).
+     *
+     * @var string
+     */
+    private $idpWhitelist;
+
     private function __construct()
     {
         $this->grantType = new OidcGrantType();
@@ -410,6 +423,9 @@ class SaveOidcEntityCommand implements SaveEntityCommandInterface
         $command->organizationDisplayNameEn = $entity->getOrganizationDisplayNameEn();
         $command->organizationUrlNl = $entity->getOrganizationUrlNl();
         $command->organizationUrlEn = $entity->getOrganizationUrlEn();
+
+        $command->idpWhitelist = json_encode($entity->getIdpWhitelist());
+        $command->idpAllowAll = $entity->isIdpAllowAll();
 
         return $command;
     }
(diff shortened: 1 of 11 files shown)


pablothedude:
I don't understand why it's needed in the commands themselves. Wouldn;t it be sufficient to

MKodde:
Regarding: 

> I don't understand why this logic is needed in the commands themselves because the data isn't needed there at all.

The workflow for editing a published entity is:
1. Entity is loaded from manage
2. Data is loaded into the form
3. User updates data and presses publish button
4. Data is stored in database
5. Data is published to manage based on the entity data from the database

We need the ACL data in the form because step 4 would not contain the ACL data as it is not present on the form/command. And we did not want to store the Manage entity to the database before the user presses the publish button. We do this to prevent local draft entities when a user leaves the edit form. So storing the data on the form, at this point is the way we ensure we are not losing any data in Manage.

And regarding:

> Updating ACL data should be handled by the dedicated UpdateEntityAclCommand/Handler logic

That crossed my mind. And could be a good solution to the problem. It would, however, complicate the publication process further. And increase the number of Manage round trips. The edit process of a publication manage entity would result in two publish and two push actions. And that seems excessive to me. And from a puristic standpoint, reusing the UpdateEntityAclCommand is not tecnically correct. As we are not updating that data at all.

And finally regarding:

> If we now for example need another page to manage something else regarding an entity (for example a IP whitelist) we should have to implement the assitional parameters also on all the Save*CommandHandler's?

Yes that's basically how this works at this point.

To solve this problem we could look into several directions:

1. Do not use the form to carry the data, but retrieve it from manage during the publication process. Which is not much more elegant in my opinion.
2. Investigate if we can use a more fine grained update mechanism in manage. Where we can send PATCH updates.

quartje:
Just for your information regarding point 2: "Investigate if we can use a more fine-grained update mechanism in manage. Where we can send PATCH updates.": Manage does provide a merge endpoint, see the wiki for documentation:
https://github.com/OpenConext/OpenConext-manage/wiki/API

MKodde:
As it turns out we do use the merge endpoint when updating entities. But the JSON generator is not intelligent/aware in regards to adding it or not.

pablothedude:
My point regarding the AclUpdateHandler was not to use it to update the ACL's but to show it shouldn't be the responsibility of the other Commands/Handlers to do something with ACL's and I think they should completely refrain from using ACL logic.

For now we could store the data in the draft database as we are currently still doing before publishing an entity so it would later be easier to refactor the logic to get the data solely from Manage in other words when we refactor the drafts.

Could you share with me why you think option 1 is not much more elegant? Personally I think it's Manage's purpose to contain entities and their state as single source of truth so why should we manage that state ourselves? Also it would be much more easy to later utilize the merge endpoint to only update certain data of an entity.

MKodde:
> My point regarding the AclUpdateHandler was not to use it to update the ACL's but to show it shouldn't be the responsibility of the other Commands/Handlers to do something with ACL's and I think they should completely refrain from using ACL logic.

Okay, that's clear now

> For now we could store the data in the draft database as we are currently still doing before publishing an entity so it would later be easier to refactor the logic to get the data solely from Manage in other words when we refactor the drafts.

We talked about this IRL. Conclusion was that the current edit prod entity workflow does not allow us to save a local draft entity before the user clicks the Publish button. This to prevent abandoned draft entities when a user cancels out of the form, or closes the browser window/tab.

> Could you share with me why you think option 1 is not much more elegant?

I do not like the continuous querying of Manage. But I don't like the current solution either :smile: 

As a better permanent solution, I think I should investigate the possibility of using the Manage merge endpoint as intended (only send actually updated values)

MKodde:
I've opened another PR that solves this problem in a different manner. And I kind of prefer that solution over this one. @pablothedude please be the judge on which of the two we'll use in the end!

pablothedude:
Thanks for addressing my concerns and for our fertile discussion. I think I'll go for the other solution as well because then we are future proof when addressing the draft entities and/or merge endpoint issues.

MKodde:
Closing this PR in favour of #315

PR #314, opened 2019-10-03
Add oidc create feature toggle

MKodde:
The toggle is configured in the parameters.yml config file. And a marshaller is tasked with informing the application if the toggle is enabled or disabled.

https://www.pivotaltracker.com/story/show/168783027


quartje:
Could you add the parameter to the Ansible template as well?

MKodde:
Good catch @quartje ! As a default I've enabled the feature toggle. Is that what you deem to be correct?

quartje:
Yes, that's good default.

PR #315, opened 2019-10-03
Persist ACL data on entity edit publication

MKodde:
The IdP ACL data would be overwritten when a published manage entity was edited. This was the case because the data was still present on the entity (with the default allow all settings) and the entity is used to build the Manage Merge Metadata on.

The solution is to load the manage entity prior to publishing it, and loading the required ACL data on the Entity (that the JSON is based on).

https://www.pivotaltracker.com/story/show/168800119
Alternative to: https://github.com/SURFnet/sp-dashboard/pull/313


pablothedude:
Discussed some things IRL and  functional tested the acl's which seemed to work as expected.

MKodde:
For the record. This PR was in contest with #313 This solution was much cleaner and won.

PR #317, opened 2019-10-24
Improve PHP 7.2 support

MKodde:
Multiple tasks have been addressed in this PR, I advise to review this PR on a commit basis as I tried to bundle separate tasks on commits. 

Some highlights include:

1. PHPUnit was updated to allow writing and testing PHP 7.2 code
1. Ansible roles where updated to use PHP 7.2 by default
1. Configure and install PHP 7.2 XDebug extension by default
1. Changed up the composer file to require PHP 7.2


app/AppKernel.php
```diff
@@ -49,7 +49,12 @@ public function getRootDir()
 
     public function getCacheDir()
     {
-        return dirname(__DIR__).'/var/cache/'.$this->getEnvironment();
+        $env = $this->getEnvironment();
+        // For test and dev, use a cache folder outside the project folder.
+        if ($env === 'dev' || $env === 'test') {
```
pablothedude: :candy:

composer.json
```diff
@@ -22,11 +22,12 @@
...
     },
     "require": {
-        "php": "^5.6||^7.0",
+        "php": "7.2.*",
         "doctrine/doctrine-bundle": "^1.6",
         "doctrine/doctrine-cache-bundle": "^1.2",
         "doctrine/doctrine-migrations-bundle": "^1.0",
-        "doctrine/orm": "^2.5",
+        "doctrine/orm": "~2.5.0",
+        "doctrine/dbal": "~2.5.12",
```
pablothedude: Why us this needed, shouldn't this be a dependency of `doctrine/orm` itself?

package.json
```diff
@@ -31,14 +31,17 @@
     "@types/jquery": "^3.3.22",
     "chart.js": "^2.7.3",
     "chartjs-plugin-datalabels": "^0.4.0",
+    "core-js": "3",
     "jquery": "^3.4.0",
     "jquery-modal": "^0.9.1",
+    "jquery.mousewheel": "^3.1.9",
```
pablothedude: Are you sure this is the right version and that `jquery-mousewheel` isn't needed? This one seems to be abandoned.

composer.json
MKodde: DBAL was pinned to a 2.5 version as newer versions remain incompatible with our test setup.

package.json
MKodde: Yeah, I first went with the `-` variant, which did not work as intended.

PR #334, opened 2020-05-06
Prevent overriding of attributes

MKodde:
Arp attributes not tracked in the SPD forms where overridden. This PR takes care of this by changing the way attributes are filtered.

Most notable is the use of the Manage entity that is now used in the metadata generators.

https://www.pivotaltracker.com/story/show/170868465

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Dto/MetadataConversionDto.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Dto/MetadataConversionDto.php
@@ -1,0 +1,559 @@
+<?php
+
+/**
+ * Copyright 2020 SURFnet B.V.
+ *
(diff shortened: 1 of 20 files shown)


MKodde:
That's a good idea, let me see what impact this would have on the current state of affairs.

pablothedude:
Maybe it was easier to use:
```
class ManageEntityMerger {
    merge(entity Entity, manageEntity Entity): Entity
}
```
But feel free to leave it as-is.

MKodde:
I've considered that, but that would have entailed adding extra ARP attributes to the Entity Entity. I think we should strive towards a Entitiy interface with certain getters defined on it. And let the infrastructure depend on (implementations) of that interface. That would have made this change a lot easier.

PR #335, opened 2020-05-07
Prevent overriding the 'exclude_from_push' attribute

MKodde:
The main purpose of this PR is to ensure we do not override the exclude from push setting in Manage once an entity is included for push in production mode.

In addition, related files that where touched during investigation have been cleaned up.

Finally boyscout work was done, and a ManageQueryService was introduced. This service is not implemented thorugout yet, but is used to prevent the double injection of the manage test/prod query clients.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
@@ -246,11 +246,14 @@ class JsonGenerator implements GeneratorInterface
         }
 
         // When publishing to production, the coin:exclude_from_push must be present and set to '1'. This prevents the
-        // entity from being pushed to engineblock.
+        // entity from being pushed to EngineBlock. Once the entity is checked a final time, the flag is set to 0
+        // by one of the administrators. If the entity was included for push, we make sure it is not overridden.
         if ($entity->isProduction()) {
             $metadata['coin:exclude_from_push'] = '1';
         }
-
+        if ($entity->isManageEntity() && !$entity->isExcludedFromPush()) {
+            $metadata['coin:exclude_from_push'] = '0';
+        }
         if (!empty($entity->getLogoUrl())) {
             $metadata = array_merge($metadata, $this->generateLogoMetadata($entity));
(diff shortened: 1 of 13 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -248,11 +248,14 @@ private function generateMetadataFields(Entity $entity, ManageEntity $manageEnti
         }
 
         // When publishing to production, the coin:exclude_from_push must be present and set to '1'. This prevents the
-        // entity from being pushed to engineblock.
+        // entity from being pushed to EngineBlock. This is only set to '1' if the state was not previously set to be
+        // included for push
```
pablothedude, line 252: Maybe mention that Manage is setting this flag also.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Service/MangeQueryService.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Service;
+
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Client\QueryClient;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Exception\InvalidArgumentException;
+
+class MangeQueryService
+{
+    private $validEnvironments = ['test', 'production'];
```
pablothedude, line 26: Consider using constants or a VO

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityResetSecretController.php
```diff
@@ -66,7 +65,7 @@ public function resetAction(Request $request, $serviceId, $manageId, $environmen
             );
         }
 
-        // fetch entity from managae and reset secret
```
pablothedude, line 69: :medal_military:

PR #337, opened 2020-05-13
Allow loopback addresses for redirect uris

MKodde:
This validation was added to the parsley validator. This should be sufficient validation. As when Javascript is disabled you wouldnt be able to submit the form with redirect uris.

See https://www.pivotaltracker.com/story/show/171620767

### DIFF
--- a/app/js/components/validator_helper.test.ts
+++ b/app/js/components/validator_helper.test.ts
@@ -1,0 +1,91 @@
+import {ValidatorHelper} from "./validator_helper";
+
+describe('validation functions', function() {
+
+    it('can validate empty strings', function () {
+        const helper = new ValidatorHelper;
+        expect(helper.validateEmpty('')).toBeTruthy();
+        expect(helper.validateEmpty(' ')).toBeFalsy();
+        expect(helper.validateEmpty('df')).toBeFalsy();
+    });
+
+    it('should validate urls', function () {
+        let allowedUrls : Array<string> = [
+            'https://www.sp-dashboard.com',
+            'http://www.sp-dashboard.com',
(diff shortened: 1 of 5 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngEntityType.php
```diff
@@ -111,7 +111,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     'entry_options' => [
                         'attr' => [
                             'data-parsley-redirecturis' => null,
-                            'data-parsley-uri' => null,
```
pablothedude, line 114: Should this also be removed in the OidcEntity ?

app/js/components/validator_helper.ts
```diff
@@ -0,0 +1,37 @@
+export class ValidatorHelper{
```
pablothedude, line 1: I'm missing type hinting for return types on all methods

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngEntityType.php
MKodde: Yes, good point!

app/js/components/validator_helper.test.ts
```diff
@@ -0,0 +1,91 @@
+import { ValidatorHelper } from './validator_helper';
```
pablothedude, line 1: Thanks for covering the tests!

PR #339, opened 2020-05-14
Enable content security policy

MKodde:
We already had a security bundle configured, but it stopped working on one of the previous changes. This slipped our attention. Now the ticket mentioned below asked to add CSP headers to SPD. 

In order to get the CSP headers working again, slight configuration changes where required. The current setup should provide a good balance between csp and practicality.

https://www.pivotaltracker.com/story/show/171107339


MKodde:
Good point, I dont think we can drop <= IE11 support. I will verify this with @quartje

app/config/security.yml
```diff
@@ -60,16 +60,29 @@ nelmio_security:
...
+                enabled: false
+            report-uri: "%router.request_context.base_url%/csp/report"
             default-src:
                 - 'self'
-                - 'data:'
             object-src:
                 - 'none'
+            style-src:
+                - 'self'
+                - 'netdna.bootstrapcdn.com'
```
pablothedude, line 76: Is it acceptable to whitelist all those hosts?

MKodde: Our software depends on those dependencies, so yes, I think its acceptable. They are required by the translation bundle. So dropping those dependencies will require quite some work

PR #340, opened 2020-05-18
Prevent overwriting of ARP motivations

MKodde:
The arp motivation would be overwritten when a arp motivations was set in manage for one of the attributes not managed in SPD. This is solved  by checking for the presence of an attribute motivation. If there is a non empty value, it is loaded on the attrbibute. Preventing overwriting
that specific attr. motivation. 

**Merge targets:**
 - `develop`
 - `release/2.5`

See: [Pivoal reject message](https://www.pivotaltracker.com/story/show/170868465/comments/214384755)

### DIFF
--- a/tests/unit/Application/Metadata/JsonGenerator/ArpGeneratorTest.php
+++ b/tests/unit/Application/Metadata/JsonGenerator/ArpGeneratorTest.php
@@ -136,6 +136,8 @@ class ArpGeneratorTest extends MockeryTestCase
             ->andReturn('idp');
         $attribute->shouldReceive('getValue')
             ->andReturn('The Manage attr value');
+        $attribute->shouldReceive('getMotivation')
+            ->andReturn('The Manage motivation');
         return $attribute;
     }
 }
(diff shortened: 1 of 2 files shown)


tests/unit/Application/Metadata/JsonGenerator/ArpGeneratorTest.php
```diff
@@ -136,6 +136,8 @@ private function buildManageAttribute(string $attributeName)
             ->andReturn('idp');
         $attribute->shouldReceive('getValue')
             ->andReturn('The Manage attr value');
+        $attribute->shouldReceive('getMotivation')
```
pablothedude, line 139: Be aware that this doesn't test anything and if the change in the ARPGenerator is removed this won't result in a failing test.

MKodde: Removing that expectation results in:

```
Mockery\Exception\BadMethodCallException : Received Mockery_1_Surfnet_ServiceProviderDashboard_Infrastructure_Manage_Dto_Attribute::getMotivation(), but no expectations were specified
 /vagrant/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php:90
 /vagrant/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php:70
 /vagrant/tests/unit/Application/Metadata/JsonGenerator/ArpGeneratorTest.php:82
```

PR #342, opened 2020-05-18
Only preserve the exclude-from-push flag on client secret reset

MKodde:
Only preserve the exclude from push flag when a client secret reset is performed. In all other cases, the flag is reset to be excluded from push

**Merge targets**
 - release/2.5
 - develop

https://www.pivotaltracker.com/story/show/171313433/comments/214387155


pablothedude:
I think the JSONGenerators are doing way to much, they are not just generating JSON but are doing a lot of other things even infrastructural decisions and I think that shouldn't be up to a JSONGenerator.
The main problem is that for manage no entities are used while Manage is infrastructure from the perspective of SPD. By using a repository for manage the benefits of a layered architecture could be reaped even more.
I think this should be refactored sooner than later to utilize the layered architecture more so the code will become better readable, better maintainable and will ultimately prevent bugs in the future.

Please consider refactoring this because the debt will be increasing by constantly adding new functionality to the JSONGenerators.

PR #343, opened 2020-05-18
Ansible: Add missing var

quartje:


MKodde:
@quartje do you want this in release 2.5 or 2.6?

PR #346, opened 2020-05-18
Make secret generator url encode safe

MKodde:
The special characters have been removed from the client secret generator. And uppercase characters have been added to the list of allowed characters. The list of required chars (which where the special chars) have been removed. Making the constructor slightly less complex.

https://www.pivotaltracker.com/story/show/170892653

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Secret.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Secret.php
@@ -24,8 +24,7 @@ class Secret
     /**
      * @var string
      */
-    private static $allowedChars = '0123456789abcdefghijklmnopqrstuvwxyz~!@#$%^&*_+=';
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Secret.php
```diff
@@ -38,18 +37,16 @@ class Secret
      */
     public function __construct($length)
     {
-        if ($length < 8) {
-            throw new Exception('The secret length should be a value greater then 7');
+        if ($length < 20) {
+            throw new Exception('The secret length should be a value greater then 20');
```
pablothedude, line 41: A minimum of 20 should be valid.

MKodde: Right you are

thijskh: Greater _than_

PR #348, opened 2020-05-20
Interpret missing exclude from push correctly

MKodde:
When resetting a client secret on production and the exclude from push attribute is missing, this indicates we are dealing with a published attribute. As the service desk not unchecks the coin checkbox, but simply removes it.

This additional logic covers that (understandable) 'misbehavior' of that coin.

Merge targets:
- `release/2.5`
- `develop`

https://www.pivotaltracker.com/story/show/171313433/comments/214527925

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/ManageEntity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/ManageEntity.php
@@ -167,10 +167,18 @@ class ManageEntity
         return false;
     }
 
+    public function isExcludedFromPushSet()
+    {
+        if (is_null($this->getMetaData()->getCoin()->getExcludeFromPush())) {
+            return false;
+        }
+        return true;
+    }
+
     public function isExcludedFromPush()
     {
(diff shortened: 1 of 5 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Dto/ManageEntity.php
```diff
@@ -167,10 +167,18 @@ public function isOidcngResourceServer()
...
+            return false;
+        }
+        return true;
+    }
+
     public function isExcludedFromPush()
     {
         if (is_null($this->getMetaData()->getCoin()->getExcludeFromPush())) {
-            return true;
+            return false;
```
MKodde, line 181: Caught a bug, where the non-existence of the coin was misinterpreted to be excluded from push. Where this is the other way around..

pablothedude:
It seems that a lot of logic is moving to the JSONGenerator, maybe we need some additional test coverage for that. Please don't do it now but I think we should address this later on.

MKodde:
For now, I've moved the exclude from push to a private function in the Json generators. This fixes the broken build. And secondly I've opened a ticket to address the growing complexity in the json generators.

https://www.pivotaltracker.com/story/show/172936842

PR #351, opened 2020-05-25
Parse OIDCng clientID correctly on secret modal

MKodde:
The OING TNG variaty was parsed correctly, but the resource server flavour was not parsed correctly showing the @ sign in the client id field where the entire protocol should have been removed.

https://www.pivotaltracker.com/story/show/169390094

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityOidcConfirmation.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityOidcConfirmation.php
@@ -17,6 +17,7 @@
  */
 namespace Surfnet\ServiceProviderDashboard\Application\ViewObject;
 
+use Surfnet\ServiceProviderDashboard\Application\Exception\InvalidArgumentException;
 use Surfnet\ServiceProviderDashboard\Application\Parser\OidcClientIdParser;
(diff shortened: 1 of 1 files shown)


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityOidcConfirmation.php
```diff
@@ -66,7 +66,9 @@ public static function fromEntity(DomainEntity $entity)
      */
     public function getEntityId()
     {
-        if ($this->protocol === DomainEntity::TYPE_OPENID_CONNECT_TNG) {
+        $isOidcng = $this->protocol === DomainEntity::TYPE_OPENID_CONNECT_TNG ||
```
pablothedude, line 69: Please throw an exception when an entity type isn't supported. That way an exception will, be thrown when an id is requested for an unsupported type. So if this is getting extended this will result in an exception during development which could be acted upon.

MKodde: Making the constructor type safe is a good addition. As that is the only moment you could create this view object for the wrong entity type.

pablothedude: Nice!

PR #352, opened 2020-05-25
Validate the privacyPolicyUrl field

MKodde:
The entry should be a valid url. By enabling parsely validation on the privacy questions form this goal could be achieved. All other fields needed updated 'required' attributes to prevent them from becoming
required form fields according to Parsley.

https://www.pivotaltracker.com/story/show/169895852


pablothedude:
It seems that URN's like `urn:example:xmpp:foo` are also valid and the error message suggests likewise.

MKodde:
Good point @pablothedude 

Fixed this by using the built in Parsley url validator in combination with the Symfony UrlType. This should always result in a valid Url (with protocol)

PR #355, opened 2020-05-26
Push metadata on production secret reset

MKodde:
This PR issues a metadata push after a production OIDC client secret reset action.

In order to make this possible I extracted the push logic from the existing publication command handlers. So some refactoring was required to get this task done. The first 5 commits cover the refactoring. And the final commit actually performs the metadata push after the secret reset.

__Warning:__
When functional testing this you will not actually be able to push the metadata from Manage prod to EngineBlock and the OIDCng proxy. The push will fail in Manage halfway. Metadata will have been pushed to Engine, but not to the not existing OIDC server. 

To fix your EngineBlock config after this task run:

`ANSIBLE_TAGS=vm_only_provision_manage_eb vagrant provision`

https://www.pivotaltracker.com/story/show/173009970

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PushMetadataCommandHandler.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PushMetadataCommandHandler.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * Copyright 2020 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\CommandHandler\Entity;
+
+use Psr\Log\LoggerInterface;
+use Surfnet\ServiceProviderDashboard\Application\Command\Entity\PushMetadataCommand;
+use Surfnet\ServiceProviderDashboard\Application\CommandHandler\CommandHandler;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Exception\PushMetadataException;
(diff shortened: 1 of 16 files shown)


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PushMetadataCommandHandler.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Surfnet\ServiceProviderDashboard\Application\CommandHandler\Entity;
+
+use Psr\Log\LoggerInterface;
+use Surfnet\ServiceProviderDashboard\Application\Command\Entity\PushMetadataCommand;
+use Surfnet\ServiceProviderDashboard\Application\CommandHandler\CommandHandler;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Exception\PushMetadataException;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Service\ManagePublishService;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
+
+class PushMetadataCommandHandler implements CommandHandler
```
pablothedude, line 28: It is not safe to have a separate CommandHandler for the metadata push the way how transactions are handled. This way when the metadata push fails the transaction with the internal database metadata is committed before it gets rolled  back. Now we would end up again with duplicate entities and conflicts between Manage and SPD entities like we had before. The way how it was designed is that one command should be able to handle one transaction and that's doesn't seem the case any more when doing this.

MKodde: We called IRL about this. A small summary.

1. Bas suspected the `DoctrineTransactionMiddleware` was tasked with rolling back publication related entities that where copied to the local database. This is not the case. That copy action is performed in another command/handler. And thus will not be rolled back when a publication fails.
2. The point Bas made is valid in another area. Previously the metadata PUSH would not be performed when publication failed. This is no longer the case. However our PM @quartje does not mind that. See quote from slack below:

> Nee dat is niet zo erg, een pushje meer of minder liggen we niet van wakker.

My suggestion is to leave this as is for now. I'm adding a warning to the SSOT ticket (removing the doctrine layer from the entity form flows). Where I will warn for possible introduction of side effects when the entity modification is performed in the publication command handler.

MKodde: In retrospect, I have moved the PUSH command to the `if (!$flashBag->has('error')) {` block. That way the PUSH is only performed when no error messages ended up in the flash bag.

PR #356, opened 2020-05-27
Do not overwrite source & value if set in Manage

MKodde:
The source and value parts of ARP attributes where overidden to their
defaults even when set in Manage.

See: https://www.pivotaltracker.com/story/show/173030675

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
@@ -64,18 +64,11 @@ class ArpGenerator implements MetadataGenerator
                 if ($attr->hasMotivation()) {
                     $attributes[$urn][0]['motivation'] = $attr->getMotivation();
                 }
-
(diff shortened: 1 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
```diff
@@ -64,6 +64,14 @@ public function build(MetadataConversionDto $entity)
                 if ($attr->hasMotivation()) {
                     $attributes[$urn][0]['motivation'] = $attr->getMotivation();
                 }
+
+                if ($entity->isManageEntity() && $entity->getArpAttributes()->findByUrn($urn)) {
```
pablothedude, line 68: Consider using a separate method to keep the code better readable. I think it's hard to understand what is actually happening in the ArpGenerator. You could for example use `mergeWithManageAttributes` and `addManageOnlyAttributes` (for the addExtraFields). It's a suggestion but I think now is the time to refactor this to keep the code readable in the future.

MKodde: Good sugestion, will integrate them!

PR #357, opened 2020-05-28
Make form labels normal casing

MKodde:
Previously they where all uppercased. By request, casing was changed to `normal`. 

Buttons and other UI elements are still uppercased. This was not part of the story to address those. Personally I like them uppercased to stand out a bit more.

https://www.pivotaltracker.com/story/show/169284607


MKodde:
Approved on Slack

PR #358, opened 2020-05-28
Match SPD header styling match that of IdPD

MKodde:
Title was changed:
 - to be blue
 - normal cased
 - text moved closer to logo

https://www.pivotaltracker.com/story/show/165311250


MKodde:
I did not see a difference before and after this change. Can you give more detail on how I can reproduce this?

MKodde:
Approved on Slack

PR #359, opened 2020-05-28
Add composer to Ansible for installation

pablothedude:
This is needed because this is removed from Deploy to
be always installed.


pablothedude:
@quartje I saw that there were some dependencies installed in SPD which shouldn't be needed in prod, but I leave them as-is for now.

PR #362, opened 2020-06-08
Add institution fields to Service

MKodde:
Added the two institution fields (id and guid). Of which the id field will only be visible when the service type is 'institution'.

Be advised, a migration was added in order to persist said fields.

https://www.pivotaltracker.com/story/show/163754504

### DIFF
--- a/app/js/components/mock/service_edit_form.html
+++ b/app/js/components/mock/service_edit_form.html
@@ -1,6 +1,5 @@
 <form name="dashboard_bundle_edit_service_type" method="post" novalidate="novalidate" data-parsley-validate=""
       class="service-form">
-
     <div class="row">
         <div class="fieldset card">
             <h2>General</h2>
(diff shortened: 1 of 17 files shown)


app/js/components/mock/service_edit_form.html
```diff
@@ -22,20 +33,28 @@ <h2>General</h2>
                     entities enabled</label>
                     <p class="parsley-errors"></p>
                     <input type="checkbox" id="dashboard_bundle_edit_service_type_general_productionEntitiesEnabled"
-                           name="dashboard_bundle_edit_service_type[general][productionEntitiesEnabled]" value="1"
-                           checked="checked"/></div>
```
pablothedude, line 26: Was the checked state deliberately removed?

MKodde: No, this represents the way the form is rendered. I tried to change as little as possible in that regard.

PR #363, opened 2020-06-09
Postpone Jira publication request ticket creation

MKodde:
The ticket is now created when publication to manage prod succeeds. If publication fails, we no longer need to remove the already created ticket. This makes the function somewhat cleaner and a test could be removed.

https://www.pivotaltracker.com/story/show/173220740


src/Surfnet/ServiceProviderDashboard/Application/CommandHandler/Entity/PublishEntityProductionCommandHandler.php
```diff
@@ -235,4 +201,32 @@ private function isNewResourceServer(Entity $entity)
         $isNewEntity = empty($entity->getManageId());
         return $isNewEntity && $entity->getProtocol() === Entity::TYPE_OPENID_CONNECT_TNG_RESOURCE_SERVER;
     }
+
+    private function createJiraTicket(Entity $entity, PublishEntityProductionCommand $command)
```
pablothedude, line 205: Good decision to split this a bit up!

PR #364, opened 2020-06-09
Prevent Jira ticket on prod client secret reset

MKodde:
When resetting the client secret, no need to create a new publication
request to the service desk Jira. The item probably is already
published. Adding the ticket twice is not expected behaviour.

https://www.pivotaltracker.com/story/show/173220529

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/PublishEntityProductionAfterClientResetCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/PublishEntityProductionAfterClientResetCommand.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Copyright 2017 SURFnet B.V.
+ * Copyright 2020 SURFnet B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


MKodde:
Good suggesteion @pablothedude. I however would like to refrain from creating a new CommandHandler. 

Instead I'd like to let it support both the `PublishEntityProductionSecretCommand` and `PublishEntityProductionCommand`. This to prevent loads of code duplication. If logic between the two changes further in the future we can always split them up. For now the gain in sticking to the DRY rules is outweighing the potential for having to maintain two almost identical command handlers.

An alternative which might be the best, but even more involved, solution would be to further 'solidify' the `PublishEntityProductionCommandHandler`. Splitting it into different command handlers for every seperate task it currently handles.

pablothedude:
> Good suggesteion @pablothedude. I however would like to refrain from creating a new CommandHandler.
> 
> Instead I'd like to let it support both the `PublishEntityProductionSecretCommand` and `PublishEntityProductionCommand`. This to prevent loads of code duplication. If logic between the two changes further in the future we can always split them up. For now the gain in sticking to the DRY rules is outweighing the potential for having to maintain two almost identical command handlers.
> 
> An alternative which might be the best, but even more involved, solution would be to further 'solidify' the `PublishEntityProductionCommandHandler`. Splitting it into different command handlers for every seperate task it currently handles.

I agree you should prevent code duplication as much as possible. But I think this doesn't have anything to do with the commands themselves because they have a completely different responsibility.  I think it would be better to utilize a command per request or usecase that way it would be much easier to refactor architectural logic, for example implementing middleware for commands. I think the best way is to move the publish logic to a dedicated publish service which could be used in the handlers. That way you don't have coupling and you don't have duplicate code either and ultimately a lot of code could be removed because it won't be needed then. But I know it would be a lot of work for now, but maybe something to take into account when refactoring out draft entities.

PR #365, opened 2020-06-09
Display more detailed resource servers in client detail view

MKodde:
More detailed resource server information (and a link to its detail screen) are added in this PR.

See: https://www.pivotaltracker.com/story/show/171620286

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
@@ -22,7 +22,7 @@ use Surfnet\ServiceProviderDashboard\Application\Parser\OidcClientIdParser;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity as DomainEntity;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Attribute;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Contact;
-use Surfnet\ServiceProviderDashboard\Domain\ValueObject\ResourceServerCollection;
+use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Dto\ManageEntity;
 
 /**
  * @SuppressWarnings(PHPMD.TooManyFields)
@@ -255,7 +255,7 @@ class EntityDetail
     private $isPublicClient;
 
     /**
-     * @var string[]|null
+     * @var ManageEntity[]|null
      */
     private $resourceServers = null;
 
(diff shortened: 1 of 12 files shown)


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
```diff
@@ -253,6 +254,11 @@ class EntityDetail
      */
     private $isPublicClient;
 
+    /**
+     * @var string[]|null
+     */
+    private $resourceServers = null;
```
pablothedude, line 260: Why not use the `ResourceServerCollection` and wouldn't it be better to use a concrete implementation then to use a variable that could be a string array or null.

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
```diff
@@ -684,4 +691,12 @@ public function isPublicClient()
     {
         return $this->isPublicClient;
     }
+
+    /**
+     * @return string[]|null
```
pablothedude, line 696: Same here

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
MKodde: The existing ResourceServerCollection is used in the context of selecting resource servers on the form. In this case we are dealing with fully equipped Manage entities that are used for display purposes.

I agree, it would be best to also pour these into more dedicated view object. But not reusing the existing objects.

MKodde:
Some additional changes where required to get this to work for all view modes.

1. The details view (draft) relies on Entity from the database. Enriched the entity service to include manage entities for the resource servers
2. The edit view also uses doctrine entities in draft mode. This entity is loaded using param conversion. And that sadly prevented me to use the new entity service method (mentioned above). I looked into removing the param converter but that also disturbs the security annotation set on the edit entity action. For now, I think this current change set is the best of both worlds. 

I'm creating a TD ticket to further remove the ParamConverter from the entity controller.

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
pablothedude: I think it woul d be better to don't return null but always a collection or array. This will prevent the necessity of validation when this method is being used. Feel free to leave it as-is for now.

MKodde:
Merging!

PR #366, opened 2020-06-10
Allow reverse redirect URLs

MKodde:
This is a best effort TS validator, to prevent disabling the other JS validation for the redirect URI field, a rudimentary reverse domain name format validator was added. Additional validation will take place in PHP as per: https://www.pivotaltracker.com/story/show/171620689/comments/214758619.

The logic added to verify the reverse redirect URL is not the most elegant of methods. Basically, the protocol and the hostname are swapped, and validation is performed as if it was a regular URL. The PHP validator adds an exception to the protocol requirement (allowing custom protocol (shceme)).

For additional details see: https://www.pivotaltracker.com/story/show/171620689/comments/214758619


MKodde:
Using redirect Url: ` nl.mijntoffedomeinnaam.portal://whatever` when your client id equals `portal.mijntoffedomeinnaam.nl` works in my dev env. I wonder what error you encountered. Only newer manage builds are able to support the reverse format. An upgrade of Manage might be required.

Discussed support for the `:/` with PO @quartje, and he's okay with dropping support for them for now. I updated this in the Pivotal ticket.

pablothedude:
> Using redirect Url: ` nl.mijntoffedomeinnaam.portal://whatever` when your client id equals `portal.mijntoffedomeinnaam.nl` works in my dev env. I wonder what error you encountered. Only newer manage builds are able to support the reverse format. An upgrade of Manage might be required.

When the `flipProtocol` method is being utilized with the url below the result is injected in the Url constructor. This results in the following.
original url: `nl.mijntoffedomeinnaam.portal://whatever`
after flip: `whatever://portal.mijntoffedomeinnaam.nl`
host: `mijntoffedomeinnaam.nl`
protocol: `whatever`
This will pass but it doesn't flip the protocol as you would expect. So in the end you're validating something different. Therefore I would suggest to not flip it yourself, but parse the Url with the native class and then flip it.

> Discussed support for the `:/` with PO @quartje, and he's okay with dropping support for them for now. I updated this in the Pivotal ticket.

Thanks for the confirmation!

MKodde:
Calling `flipProtocol` with `nl.mijntoffedomeinnaam.portal://whatever` results in:
`whatever://portal.mijntoffedomeinnaam.nl`.

This is exactly what I expect. The method name could be considered untrue, as it also reverses the hostname.

Or am I overlooking something here?

pablothedude:
> considered

`nl.mijntoffedomeinnaam.portal` > ~~nl~~`.mijntoffedomeinnaam.portal`
Shouldn't `whatever` be the path after flipping instead of the schema?

MKodde:
In the reverse format, the protocol = host and host = protocol. Those are flipped (and the hostname parts are reversed). After that the Url can be validated as if a normal Ulr. The  custom protocol portion is not verified in the JS validation.

Your finding of a domain portion that is dropped from the domain name is worrying, but I'm unable to reproduce this behavior. Can you elaborate?

MKodde:
No problem!

PR #367, opened 2020-06-16
Use entity service throughout application

MKodde:
By using the `Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php` in favour of directly using the QueryClients. Coupling to the infra layer can be reduced.

Fixed this for: 
 - LoadEntityService
 - PublishEntityClient
 - UniqueEntityIdValidator

In addition, the publish clients are now only used via ManagePublishService. https://github.com/SURFnet/sp-dashboard/pull/367/commits/73b8bb3324e94840dac17402c6e3f8a13e8b6436


MKodde:
This change was included in the 'remove the save button' PR

PR #369, opened 2020-06-17
Increase CSP strictness

MKodde:
Remove unwanted CSP settings that where explicitly added for the translations section of SPD. They are now overloaded in the Apache config. As Nelmio does not provide a means to override CSP config per page/path.

To get this into an existing dev-env. Please update the contents of your `/etc/httpd/conf.d/spdashboard.conf` with the contents of the Jinja template found in this PR.

https://www.pivotaltracker.com/story/show/173371989

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -73,14 +73,11 @@ nelmio_security:
                 - 'none'
             style-src:
                 - 'self'
-                - 'netdna.bootstrapcdn.com'
                 - 'unsafe-inline'
             script-src:
                 - 'self'
-                - 'ajax.googleapis.com'
             font-src:
                 - 'self'
-                - 'netdna.bootstrapcdn.com'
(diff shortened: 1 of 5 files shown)


app/config/security.yml
```diff
@@ -73,14 +73,12 @@ nelmio_security:
                 - 'none'
             style-src:
                 - 'self'
-                - 'netdna.bootstrapcdn.com'
                 - 'unsafe-inline'
             script-src:
                 - 'self'
-                - 'ajax.googleapis.com'
+                - 'unsafe-inline'
```
pablothedude, line 79: This change goes against the concept of defence in depth because this will effectively disable the XSS protection mechanism of CSP. Could the googleapis and bootstrapcdn not be moved to the dev config. That way it'll be secure until the webserver removes the csp headers.

MKodde: I think that one can be removed entirely, might have missed that while creating the exception in the Apache config.

PR #373, opened 2020-06-23
Enable and fix Url validation

MKodde:
The Jest test behaves differently to the browser when it comes to the `flipProtocol` method. The URL parser is able to correctly parse on the Jest test, but fails to do so in a Firefox browser.

@pablothedude any thoughts on this?

For example, try this redirect url `spdashboard.dev.support.surfconext.nl://https/my/path`

**Edit**
@pablothedude suggested to use `parse-url` library. And that fixed the problem with ease.

See:
`https://www.pivotaltracker.com/story/show/172833472`

### DIFF
--- a/app/js/components/validation.js
+++ b/app/js/components/validation.js
@@ -109,18 +109,18 @@ if (form.length) {
 // Add a stricter url validator, fields validated with urlstrict are
 // allowed to be empty, or a URL with protocol.
 window.Parsley.addValidator('urlstrict', function (value, requirement) {
-    let helper = new ValidatorHelper();
+    const helper = new ValidatorHelper();
     return helper.validateEmpty(value) || helper.validateUrl(value);
 }, 32).addMessage('en', 'urlstrict', 'This value should be a valid URL.');
 
 // Add URI validator (must be URN or URL), fields validated with urn
 // must be empty or be a valid URN  or URL (with protocol).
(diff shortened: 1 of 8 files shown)


MKodde:
Thanks for noting that regression. My linting tool removed it while building the PR as it thought it was not used.. I've force pushed the change, so please reset your branch to get the change

MKodde:
Thanks for noting that regression. My linting tool removed it while building the PR as it thought it was not used.. I've force pushed the change, so please reset your branch to get the change

pablothedude:
The proposed change doesn't seem to work, I think you're unable to reproduce this because you're testing with only 1 redirect  url. When you test it with multiple redirect url's you could reproduce the regression.

And with the latest change all redirect url's are considered valid as long as the last one is valid.

I've added a test to show what I mean.
https://github.com/SURFnet/sp-dashboard/pull/373/commits/cbffb4487b6ea1b95028710235bb65ddefa10048

pablothedude:
It seems Symfony is laking an `Or` validator :wink:. Personally I don't like the management of the validation state in the validator itself. Probably it will be cleaner to validate the scheme with parse_url and if it isn't 'http' or 'https' add the reversed clientid hostname to the protocol constraint and do the validation in the parent only once. But feel free to leave it as-is for now because now we have at least test to validate this.

PR #376, opened 2020-06-25
Add PHP Stan to the project adding more robust static analysis

MKodde:


MKodde:
If ever we'd like Stan added to the QA toolchain. This PR might prove helpful. For now, closing a dead PR

PR #377, opened 2020-07-09
Build Docker images on Github Actions

quartje:
This is an initial trial to containerize some of the OpenConext workloads. 

In this PR, I've done several things:
1) Create two dockerfiles, one for php-fpm and one for Apache. Both have been based on the Alpine Linux base image, to keep things as small as possible. The application is installed in these images as well.
2) Create a Github action to build and publish these two docker images using the GitHub package registry. 
3) Some extra work was needed to fix composer. 

This PR goes hand in hand with the one in OpenConext-deploy:
https://github.com/OpenConext/OpenConext-deploy/pull/280

It's meant as a starting point for SPdashboard in Docker


docker/httpd/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM alpine
```
pablothedude, line 1: Consider using nodes own image:
https://hub.docker.com/_/node

docker/httpd/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /var/www/html/
+
+RUN apk add --no-cache yarn && \
+    yarn install && \
+    yarn run encore production
+
+FROM httpd:2.4-alpine
+MAINTAINER Bart Geesink (bart.geesink@surfnet.nl)
+RUN mkdir -p /var/www/html/
+COPY --from=0 /var/www/html/web/. /var/www/html/web/
```
pablothedude, line 13: Personally I would use `--from=node-build` and then use `FROM node as node-build`  in the referenced build stage.

docker/httpd/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM alpine
+RUN mkdir -p /var/www/html/
+COPY . /var/www/html/
+WORKDIR /var/www/html/
+
+RUN apk add --no-cache yarn && \
+    yarn install && \
+    yarn run encore production
+
+FROM httpd:2.4-alpine
```
pablothedude, line 10: Apache? :wink: Consider using nginx ...

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    networks:
+      - spdashboard
+    container_name: fpm
+  web:
+    build:
+      context: 'docker/httpd/'
+    depends_on:
+      - php-fpm
+    expose:
+      - "80"
```
pablothedude, line 18: Consider to use https so when developing https quirks could also be encountered.

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Build the Apache container and push to GitHub Packages
+      uses: docker/build-push-action@v1
+      with:
+        dockerfile: docker/httpd/Dockerfile
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+        registry: docker.pkg.github.com
+        repository: surfnet/sp-dashboard/spdashboard_web
+        tag_with_ref: true
```
pablothedude, line 38: A step could be added to run tests but then it would be better to move the Ant tasks to Composer. This was already done in other projects. This would also be beneficial right now because Java wouldn't be needed for the Travis builds.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    networks:
+      - spdashboard
+    volumes:
+      - type: volume
```
pablothedude, line 22: This could also use the short hand form above but should then become: `spddata:/var/www/html/:nocopy`

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+        target: /var/www/html/
+        volume:
+          nocopy: true
+    depends_on:
+      - php-fpm
+    container_name: web     
+
+networks:
+  spdashboard:
+    driver: bridge
```
pablothedude, line 33: `driver: bridge` could be omitted

pablothedude (commented):
Great work!

You should consider running the containers not as root. Also a `docker-compose-dev.yml` to overrule the volumes which mount the projectroot in `/var/www/html` could become valuable in the future for debugging purposes but if the container is ran as root this could result in local files being owned by root. Which could sometimes be problematic during development.

I would also like to see a Manage container.

Also some testing before pushing the images would be a necessity before taking this into production

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      - "80"
+    networks:
+      - spdashboard
+    volumes:
+      - type: volume
+        source: spddata
+        target: /var/www/html/
+        volume:
+          nocopy: true
+    depends_on:
```
pablothedude, line 27: duplicate?

docker/httpd/Dockerfile
quartje: I did consider it. However, the whole platform (apart from Stepup) uses Apache. I see no reason to change it.

docker/httpd/Dockerfile
quartje: I've added your suggestions

docker-compose.yml
quartje: That's certainly a good idea, but involves some hassles in creating and maintaining certificates. I've added it as a chore on the backlog

docker-compose.yml
quartje: Indeed ..

docker-compose.yml
quartje: Done.

.github/workflows/main.yml
quartje: Let's make it so after merging this branch :)

MKodde (commented):
Nice work, see some feedback in the comments below.

MKodde (approved):
LGTM!

PR #379, opened 2020-07-20
Remove memberOf as managed attribute

pablothedude:
The memberOf attribute was defined as attribute but didn't seem to be
managed by SPD. So when pushing the attribute to Manage it was replaced
with an empty value. So therefore the attribute was removed and a
couple of checks for mapping values to getters/setters could be
removed to make the code more explicit and thus better readable.

https://www.pivotaltracker.com/n/projects/1400064/stories/173508616

### DIFF
--- a/app/Resources/metadata/attributes.json
+++ b/app/Resources/metadata/attributes.json
@@ -148,15 +148,5 @@
       "urn:mace:dir:attribute-def:eduPersonTargetedID",
       "urn:oid:1.3.6.1.4.1.5923.1.1.1.10"
     ]
-  },
-  {
-    "id": "isMemberOf",
-    "getterName": "",
-    "setterName": "",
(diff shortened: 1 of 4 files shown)


Tjeerd:
LGTM, the only thing I can think of, with my limited knowledge, is that actually 3 changes had been made, which are not necessarily directly related to each other. So if some behaviour is changed by no longer skipping a getter or setter, this might be a bit hard to correlate to this change. Other than that 👍

pablothedude:
For your understanding. The check if a getter or setter method was defined in the JSON became unnecessary after this change. Because this implicit logic introduced the bug in the first place I removed it and seemed more a leftover from the old sod form, which you might have known. The logic wasn't needed any more so I refactored it out of the code.

PR #384, opened 2020-10-19
Ensure client secret set on copy to production

MKodde:
https://www.pivotaltracker.com/story/show/175288695

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
@@ -70,6 +70,9 @@ class EntityMergeService
         $this->playGroundUriProd = $oidcPlaygroundUriProd;
     }
 
+    /**
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity)
+     */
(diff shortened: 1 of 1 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
```diff
@@ -101,8 +104,12 @@ public function mergeEntityCommand(SaveEntityCommandInterface $command, ?ManageE
         }
 
         $oidcClient = null;
+        $isNewEntity = !$manageEntity;
+        $isCopyToProduction = $manageEntity && $manageEntity->getId() === null;
+
         $secret = new NullSecret();
-        if (!$manageEntity && $protocol->getProtocol() !== Constants::TYPE_SAML) {
+        // Set the client secret when createing a new entity or when we are copying an entity.
+        if ($protocol->getProtocol() !== Constants::TYPE_SAML && ($isNewEntity || $isCopyToProduction)) {
```
pablothedude, line 112: You could consider lowering the cyclomatic complexity by moving this check to a method on the entity itself. Something like `hasClientSecret` for example. Also when new entity types are introduced you need to implement this check and it would be harder to forget to update this check. But feel free to leave it as is for now.

pablothedude (approved):

PR #385, opened 2020-10-21
Docker: Add content security policy to the docker httpd config

quartje:
With this change, the translations work again


MKodde (approved):
LGTM! :shipit:

PR #386, opened 2020-10-21
Docker php fpm: Clear cache after installation

quartje:
Delete the symfony cache after the installation

### DIFF
--- a/docker/php-fpm/Dockerfile
+++ b/docker/php-fpm/Dockerfile
@@ -26,6 +26,7 @@ RUN apk add --no-cache libxml2-dev \
     SYMFONY_ENV=prod composer run-script symfony-scripts && \
     chown -R www-data /var/www/html/var/cache && \
     chown -R www-data /var/www/html/var/logs && \
-    chown -R www-data /var/www/html/var/sessions
+    chown -R www-data /var/www/html/var/sessions && \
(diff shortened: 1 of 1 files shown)


docker/php-fpm/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN apk add --no-cache libxml2-dev \
     SYMFONY_ENV=prod composer run-script symfony-scripts && \
     chown -R www-data /var/www/html/var/cache && \
     chown -R www-data /var/www/html/var/logs && \
-    chown -R www-data /var/www/html/var/sessions
+    chown -R www-data /var/www/html/var/sessions && \
+    rm -rf /var/www/html/var/cache/prod/
```
MKodde, line 30: Is the Symfony cache warmed back up at a later stage? Otherwise this makes not much sense to me.

quartje: The configuration is applied when the docker image starts up. The first http hit will then create the correct cache right?

MKodde: I'm not fully sure that also works in `prod` mode, but apparently it does.

MKodde (approved):

docker/php-fpm/Dockerfile
quartje: It does!

PR #388, opened 2020-10-22
Github actions: Move to the new GitHub container registry, which allo…

quartje:
…ws anonymous access to the containers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,8 +21,8 @@ jobs:
       with:
         dockerfile: docker/php-fpm/Dockerfile
         username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
-        registry: docker.pkg.github.com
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -21,8 +21,8 @@ jobs:
       with:
         dockerfile: docker/php-fpm/Dockerfile
         username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
-        registry: docker.pkg.github.com
+        password: ${{ secrets.GH_CONTAINER_REGISTRY_PAT }}
```
MKodde, line 24: `PATH` ?

MKodde (commented):
Might I've found a typo?

.github/workflows/main.yml
quartje: PAT = Personal Access Token

MKodde: Ah, the more you know!

MKodde (approved):

PR #389, opened 2020-10-22
Feature/move to one multistage docker file

quartje:
Because most stuff that is being built during Docker build is needed in both Docker images, it makes more sense to do so in a single Dockerfile, using stages to get the correct image.


MKodde (approved):
I'm no docker expert by any means, but this looks to be more efficient than the previous setup. I see no missing actions, looks good to me.

PR #390, opened 2020-11-02
Set correct publication state for prod entities

MKodde:
The detail screen showed illegal entity options for the published entities.

https://www.pivotaltracker.com/story/show/175535045

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ cache:
 
 before_script:
   - phpenv config-add travis.php.ini
+  - composer self-update --1
   - composer install --no-interaction
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH="$HOME/.yarn/bin:$PATH"
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
@@ -159,10 +159,15 @@ class EntityService implements EntityServiceInterface
                 $entity->setService($service);
                 // Entities that are still excluded from push are not really published, but have a publication request
                 // with the service desk.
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ cache:
 
 before_script:
   - phpenv config-add travis.php.ini
+  - composer self-update 1.10.16
```
pablothedude, line 23: Maybe it's better to use: `composer self-update --1`

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
```diff
@@ -352,4 +355,15 @@ private function updateEntityStatusWithJiraTicketStatus(ManageEntity $entity, Is
...
         }
     }
+
+    private function updateStatus(ManageEntity $entity)
+    {
+        $excludeFromPush = $entity->getMetaData()->getCoin()->getExcludeFromPush();
+        if ($excludeFromPush === '1') {
+            $entity->updateStatus(Constants::STATE_PUBLICATION_REQUESTED);
+        }
+        if ($excludeFromPush === '0') {
```
pablothedude, line 365: Consider throwing an exception when no valid excludeFromPush value is met e.g. '-1'.

pablothedude (approved):
Please add a link to the issue in Pivotal if there is one.
the changes themselves are looking good.

.travis.yml
MKodde: Superb advice!

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
MKodde: That is certainly possible, but would mess up test entities who might not have the exclude from push coin. So I either had to add catch clauses around the method call or leave as is. I opted for the latter.

PR #392, opened 2021-01-19
Remove the entity details overview by enhancing the service overview

Badlapje:
This PR deals with [this ticket in pivotal](https://www.pivotaltracker.com/story/show/165965607)

It makes a number of changes:
- removes the entity details overview and moves it's functionality to the service overview
- removes the primary contact column from both overviews
- removes the copy/clone functionality from the actions list & the entity details page
- makes the entity tables on the service overview full width by moving the diagram into it's own section
- ensures the diagram is hidden when all is green, thus enhancing UX.
- moves tests to the service overview where possible and adds them where not
- enhances the accessibility of the service overview page by implementing some quick wins:
   - increase the font-size of the service overview so it becomes more readable, as a side-effect of this the base font-size is also increased.  This may cause overflow issues on other pages, which'll be addressed in the next PR.
   - ensuring all animations on this (and other) pages are only shown when users have not expressed they don't want them.  This allows users of modern browsers to opt out, but does not hide animations from users of older browsers.
   - increase the semantics of the html
   - increase the screenreader support for the page
   - improve the keyboard navigation for the page


app/scss/components/service_status.scss
```diff
@@ -98,20 +102,142 @@
...
+  text-transform: none;
+}
+
+.button[for^="add-for-test"],
+.button[for^="add-for-production"] {
+  font-size: $normalFont;
+  line-height: 26px;
+  margin-right: 0;
+  text-transform: none;
+}
```
MKodde, line 195: I think this is no longer used as the buttons became links, and the for attributes where amended with the service name.

docs/testing-with-different-priviliges.md
```diff
@@ -0,0 +1,24 @@
+# Testing the front-end with different priviliges
```
MKodde, line 1: :heart:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -122,15 +126,28 @@ public function overviewAction()
...
+        $publishedEntity = $this->get('session')->get('published.entity.clone');
+
         return $this->render('DashboardBundle:Service:overview.html.twig', [
             'services' => $serviceList,
-            'isAdmin' => false
+            'isAdmin' => false,
+            'publishedEntity' => $publishedEntity,
+            'showOidcPopup' => $this->showOidcPopup($publishedEntity),
+            'productionEntitiesEnabled' => $productionEntityEnabled,
+            'privacyStatusEntities' => $privacyOK,
```
MKodde, line 150: Not wrapped in array here, but on the next actionMethod, the PrivacyOK var is wrapped in an array. Why the difference between the two?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
```diff
@@ -301,10 +301,14 @@ services:
 
     surfnet.dashboard.repository.issue:
 
+    Surfnet\ServiceProviderDashboard\Application\Service\ServiceService:
+        class: Surfnet\ServiceProviderDashboard\Application\Service\ServiceService
```
MKodde, line 305: Autowire/autoconfigure hell in SP Dashboard :disappointed:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -512,15 +502,19 @@ service.form.label.service_type_institute: Institute
 service.form.label.service_type_non_institute: Not an institute
 service.form.label.institution_id: Institute ID
 service.form.label.institution_guid:  Institute GUID
-service.admin_overview.title: Service overview
+service.admin_overview.title: "%name% overview"
```
MKodde, line 505: :ok_hand:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityDetail/detail.html.twig
```diff
@@ -5,8 +5,9 @@
 
     <h1>{% block page_heading %}{{ 'entity.detail.title'|trans }}{% endblock %}</h1>
 
+    <h2>{{ isAdmin }}</h2>
```
MKodde, line 8: Dev leftover?

MKodde (approved):
Quite a pickle to review this one! I tried my best and came up with some nits. See if they are worth it to fix/investigate. Overall this looks very nice!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
Badlapje: nothing, this is already an array, initialized on line 130.  I changed it so the admin action also has the initialization higher up, making the two similar.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityDetail/detail.html.twig
Badlapje: nice spot! 💯

app/scss/components/service_status.scss
Badlapje: right you are! 👍

PR #393, opened 2021-02-05
Apache: Set the Referer policy header

quartje:
Add  Referrer-Policy "origin-when-cross-origin" to the httpd docker image


Badlapje (approved):
Looks good!

PR #396, opened 2021-02-09
Remove oidcng restriction

MKodde:
OIDCng is open for the public. The previous toggle in the Service edit/create forms have been removed. In addition the similar restriction from the Manage config was also removed.

See: https://www.pivotaltracker.com/story/show/175139827


MKodde:
Review instructions:

1. After applying this branch and it's attached database migration, you will lose the `service oidcng_enabled` field. This field can later be added by running the 'back' version of the migration. You could set all values to be true. Or like I did last time, temporarily remove the database field annotation (`@ORM\Column(type="boolean")`) from the Service entity `private $oidcEnabled` field. I'd like to see this merged as quickly as possible. Mitigating these problems altogether.
2. Sorry about the messed up last commit. I accidentally added some 'get the build to finish' changes to a perfectly atomic commit. Hope you can forgive me, I'll be sure to create a squashed merge commit later on.
3. Running the migrations is simple. From your guest, in the webroot folder run: `bin/console doctrine:migrations:migrate` to apply the migration. And to revert the migration run: `bin/console doc:mig:exec 20210209152257 --down`

Happy reviewing :fox_face:

Badlapje (approved):
The code changes look good.  Couldn't find anything else that needs removing when checking it out locally.

However: i didn't try to run the code yet due to timeconstrainst.  Approving it as is for now.  If you'd rather wait till after i ran the code that should be at around 11:30.

MKodde:
I tested this quite thoroughly, and if issues remain, we will surely find them in our development efforts.

PR #397, opened 2021-02-15
Manage organization names

MKodde:
The organization display, and the regular organization names are now manageble via the Service create and edit forms.

Whenever an entity of the service is saved, the current values of the service's org names are transported to the manage entity.

@quartje is investigating if the Organization URL should also be made configurable in the Service create/edit views. If that task is to be performed. That should land in a new PR, this PR can be used as a template for the tasks that need to be performed.

**Warning**: A database migration is required to test this feature
**See**: https://www.pivotaltracker.com/story/show/175513094

### DIFF
--- a/app/DoctrineMigrations/Version20210211124755.php
+++ b/app/DoctrineMigrations/Version20210211124755.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Application\Migrations;
+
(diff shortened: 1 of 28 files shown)


Badlapje (approved):
Nice work!  Works perfectly when tested: :shipit:

PR #398, opened 2021-02-15
Docker: Add php-fpm override config

quartje:
The zz-docker.conf in the local directory overwrites the default file as
supplied in the Docker image. This allows us to set the secure flag on
the session cookie, and set other php ini settings if needed

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -40,4 +40,6 @@ RUN apk add --no-cache libxml2-dev \
     chown -R www-data /var/www/html/var/sessions && \
     rm -rf /var/www/html/var/cache/prod/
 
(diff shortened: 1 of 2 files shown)


MKodde (approved):
Looks like a very straight forward solution. Do we require other PHP settings for adequate PHP security?

You could diff a phpinfo from your docker vm and another ansible managed application.

quartje:
I've added some other parameters to configure explicitly.

MKodde (approved):
Thanks, this is even better!

It might be a good idea to create a good default docker config template somewhere on the githubs. Or look into using a well used userland default. Maybe the php Docker container is set with safe defaults as a standard?

But I think I'm running ahead on this one. For now this is a great improvement!

PR #399, opened 2021-02-15
Drop attribute constraint

MKodde:
No attributes are required for publication of an entity. This was
previously the case. But recent policy change no longer require this.
That enabled us to remove the Constraint validator.

A new webtest case verifies the ability to not require an Attribute.

See: https://www.pivotaltracker.com/story/show/175511455


Badlapje (approved):
❤️

PR #400, opened 2021-02-17
Remove OrganizationDisplayName from Service

MKodde:
This once again removes the OrganizationDisplayName from the service entity. A miscommunication caused this to happen. Note that:

1. The existing migration was updated. Be careful, and remove the two removed fields by hand.
2. Bonus: the display name and the organization URL have been removed from the entity forms. They are not required there. Not sending them to Manage will not cause them being overridden.

Yay for PR 400
:100: :100: :100: :100:


Badlapje (approved):
Great work, 🚀

PR #401, opened 2021-02-17
Ensure presence of InputId on new entity form

MKodde:
This id ensures we know which of the services was selected to create a new
entity fom.

https://www.pivotaltracker.com/story/show/176976523


quartje (commented):
Seems to be working on the test environment

MKodde:
Yes! As this is blocking development. I'm merging this.

PR #402, opened 2021-02-17
Remove trailing slashes from ClientID

MKodde:
When present, remove them. Also cleaned up the unused OidcClientIdParser as per boy scout rule!

https://www.pivotaltracker.com/story/show/173509427


Badlapje (approved):
Nothing but net! 🏀

PR #404, opened 2021-02-18
Remove default scope from Oidcng RP entities

MKodde:
Stop setting default scope oidc on RP entities. This is no longer required, as the resource servers are migrating to their own entity type.

https://www.pivotaltracker.com/story/show/176961848


Badlapje (approved):
Looks logical & when testing it i don't see anything break.  Nice work!  :shipit:

PR #405, opened 2021-02-18
Remove remaining OIDC crud

MKodde:
Some test oidc remaining support was removed. In a previous commit the
OIDC client id generator was removed. This can be seen as part of the
task mentioned below.

https://www.pivotaltracker.com/story/show/175139821


Badlapje (approved):
🧹 out the dust 💯

PR #406, opened 2021-02-18
Remove all leftover Entity entity references

MKodde:
This is not a task on the 2.8 changelist. But a well deserved boy scout rule task while waiting for my fresh VM to arrive.

This actually removes the Entity entity. Some things to note:

- The console command to clean up unused local entities was removed
- A migration to drop the entity table was created
- Unused handlers where removed
- The service container was cleaned up
- Tests are updated


Badlapje (approved):
Am i right in assuming we are dropping entity because manage is the single source of truth now, so it should be ManageEntity everywhere?

Other than the above question: 🗑️ is full, please :shipit: to empty

MKodde:
Yes, you are absolutely right. On my previous effort to get rid of any Entity entity references I drew the line at some given point. And decided to leave the entity in the project for the time being. That was to prevent an even bigger change set and to keep perspective on things. So now that the local saving of draft entities is no longer used. The rest of this crud is removed as well.

PR #411, opened 2021-02-24
Ensure template has default choice

Badlapje:
Prior to this change, the template option did not always have a default choice.

This change ensures there is always a checked option for the template.

See the story in pivotal for more info: https://www.pivotaltracker.com/story/show/165599582


MKodde (approved):
Good catch!

PR #412, opened 2021-02-25
Allow setting multiple grants

MKodde:
In order to support setting multiple grants, several changes to the code where required.

1. Support for multiple grants in SPD is required. Previously we only handled a single grant. This is now a simple collection of grants
2. Do not overwrite the Manage managed grants. These are being ovewritten as we only kept a single value and posted that back into the manage environment.

Note to reviewer: 
~This was rebased onto the docker PR. Please only review the `Allow setting multiple grants` commit.~
The additional commit was a fixup of a mistake I made. One of the previous stories mentioned that OIDCng entities should not track scopes at all. This was not implemented fully yet.

See: https://www.pivotaltracker.com/story/show/174746007

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,7 +9,7 @@ Vagrant.configure(2) do |config|
     v.customize ["modifyvm", :id, "--ioapic", "on"]
     v.customize ["modifyvm", :id, "--memory", "6072"]
   end
-  
+
   config.vm.provision "ansible" do |ansible|
     ansible.playbook = "ansible/vagrant.yml"
     ansible.groups = {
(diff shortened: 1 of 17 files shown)


Vagrantfile
```diff
@@ -1,5 +1,5 @@
 Vagrant.configure(2) do |config|
-  config.vm.box = "OpenConext-CentOS-7.0"
+  config.vm.box = "sp-dashboard"
```
Badlapje, line 2: is this correct?

Vagrantfile
```diff
@@ -9,7 +9,10 @@ Vagrant.configure(2) do |config|
     v.customize ["modifyvm", :id, "--ioapic", "on"]
     v.customize ["modifyvm", :id, "--memory", "6072"]
   end
-  
+  config.ssh.private_key_path = "/home/michiel/.vagrant.d/insecure_private_key"
```
Badlapje, line 12: this seems like a custom mod that should not be in the repo?

Badlapje (changes requested):
Great work!  Looks good to me, save for the changes to the vagrantfile: looks like an accidental commit?

MKodde:
What an utter pillock I have been! Vagrantfile was reverted. History rewritten..

Badlapje:
> What an utter pillock I have been! Vagrantfile was reverted. History rewritten..

Welcome to my world 😝 

💯 for the use of the word pillock!

Badlapje (approved):

PR #413, opened 2021-02-25
Support attributes with multiple values

MKodde:
These where previously reset on a publication to manage. This resulted in loss of data on Manage side. By tracking the manage data and only updating the motivations, this problem is solved.

https://www.pivotaltracker.com/story/show/174549876


Badlapje (approved):
Seems to work perfectly 👍   Great work!

PR #414, opened 2021-02-25
Ensure typecheck for redirect url doesn't fail

Badlapje:
Prior to this change, the typecheck for the redirect url failed when the redirect url wasn't set.

This change ensures the check doesn't fail by ensuring that if the field isn't set an empty array is used.

Bug found while testing PR #413


MKodde (approved):
Thanks!

PR #415, opened 2021-02-25
Add info fields to create entity titles

Badlapje:
Prior to this change, the create entity modal had no extra info to explain the different fields / titles.

This change ensures that protocol gains an explanation link & template gains an explanation fold-out (aka tooltip modern version).

For more info see https://www.pivotaltracker.com/story/show/165599582

### DIFF
--- a/app/scss/components/tooltip.scss
+++ b/app/scss/components/tooltip.scss
@@ -1,0 +1,44 @@
+.tooltipContent {
+  margin-top: 1rem;
+  max-height: 0;
+  opacity: 0;
+
+  @include motion {
+    transition: max-height .4s, opacity .4s;
+  }
+
+  p + p {
+    margin-top: .5rem;
+  }
+
+  strong {
+    font-weight: 600;
+  }
+}
+
+label.tooltipLabel:not(.joske):not(.is):not(.nen):not(.koolmarchant) {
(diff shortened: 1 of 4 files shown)


quartje:
Nice work! It works for me in Chrome, but not in FireFox: The "what does this mean" is not clickable there

app/scss/components/tooltip.scss
```diff
@@ -0,0 +1,44 @@
+.tooltipContent {
+  margin-top: 1rem;
+  max-height: 0;
+  opacity: 0;
+
+  @include motion {
+    transition: max-height .4s, opacity .4s;
```
MKodde, line 7: Smooth

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityModal/addEntityModal.html.twig
```diff
@@ -30,7 +36,18 @@
...
+                {% set allowedTags = 'allowed.html.tags'|trans %}
+                <input type="checkbox" name="templateTooltip" id="templateTooltip" hidden class="tooltip tooltipCheckbox" />
+                <h5 class="add-entity-question" id="fieldset3">
+                    <span class="title">{{ templateTitle }}</span>
+                    <label for="templateTooltip" class="tooltip tooltipLabel">
+                        {{ 'entity.add.template.title.tooltip.srText'|trans }}
+                    </label>
+                </h5>
+                <div class="tooltip tooltipContent">
+                    {{ 'entity.add.template.title.tooltip.text'|striptags('<br><wbr><p><a><ul><ol><li><dl><dd><dt><blockquote><cite><strong><em><div><span><address><time>')|trans|raw }}
```
MKodde, line 49: Thank you for adding the filter on the allowed tags. I recon this is a sane set of allowed elements!

MKodde (approved):
Personally I do not share @Badlapje's issues in Firefox. Smooth sailing on my machine. Maybe your FF cached some CSS Bart?

Badlapje:
> 
> 
> Nice work! It works for me in Chrome, but not in FireFox: The "what does this mean" is not clickable there

Works in Firefox as well, both locally and on browserstack.  I updated the PR so the cursor now also reflects the clickability of the phrase.

PR #416, opened 2021-02-26
Fix custom redirect url validation

Badlapje:
Prior to this change, custom urls didn't validate correctly.

This change ensures that custom redirect urls are valid.  Custom urls are urls where the hostname is provided in reverse as per https://tools.ietf.org/html/rfc8252#section-7.1.
We also allowed for technically "faulty" custom redirect urls which feature a double //.  Eg: nl.test://redirect.  This ought to be nl.test:/redirect as per te specs (with a single /).  The use case where a user types it with double slash is covered however.

For more info see https://www.pivotaltracker.com/story/show/173624681

### DIFF
--- a/app/js/components/validator_helper.ts
+++ b/app/js/components/validator_helper.ts
@@ -1,6 +1,4 @@
 // tslint:disable-next-line:no-var-requires
-const Url = require('url-parse');
-
 export class ValidatorHelper {
 
   public validateEmpty(value: string): boolean {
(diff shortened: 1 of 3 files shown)


Badlapje:
@MKodde the one failing test with the "whatever" protocol: is that a valid reverse url?  Doesn't seem like it should be as only http(s) is allowed for reverse urls if i read the spec correctly?

MKodde:
@Badlapje I might be mistaken, but I think the protocol can also be 'custom' as in not http/https. Thats what makes this URL scheme so 'nice'. @quartje can you chime in on this?

app/js/components/validator_helper.ts
```diff
@@ -84,17 +93,11 @@ export class ValidatorHelper {
       return false;
     }
     const regex = /^https?:\/{2}/i;
-    if (!input.match(regex)) {
-      return true;
-    }
-    return false;
+    return !input.match(regex);
```
MKodde, line 96: One-liners for the win! :+1:

app/js/components/validator_helper.ts
```diff
@@ -18,6 +16,9 @@ export class ValidatorHelper {
    */
   public validateUrl(value: string): boolean {
     let rawUrlValue = value;
+    const urlThingy = new URL(value);
+    // tslint:disable-next-line:no-console
+    console.log(urlThingy);
```
MKodde, line 21: Development left-over?

MKodde (commented):
Aside from the console.log left-over things are looking good here. Maybe Bart can verify if this behavior is now as intended?

quartje:
> @Badlapje I might be mistaken, but I think the protocol can also be 'custom' as in not http/https. Thats what makes this URL scheme so 'nice'. @quartje can you chime in on this?

The scheme which normally is http:// or https:// is different here:

whatever://redirect would be valid here. However, we apply extra restrictions, and we demand that "whatever" should be the domainname in reverse:

clientid: my.app.nl
valid custom redirect: nl.app.my://redirect

Badlapje:
> 
> 
> > @Badlapje I might be mistaken, but I think the protocol can also be 'custom' as in not http/https. Thats what makes this URL scheme so 'nice'. @quartje can you chime in on this?
> 
> The scheme which normally is http:// or https:// is different here:
> 
> whatever://redirect would be valid here. However, we apply extra restrictions, and we demand that "whatever" should be the domainname in reverse:
> 
> clientid: my.app.nl
> valid custom redirect: nl.app.my://redirect

The currently failing test is one with this url: ```com.example.foobar://whatever```  The whatever part of that is a protocol specification I assume, given that the flipped url (aka normalized url) is: ```whatever://foobar.example.com```.  But unless i missed something, only http / https are allowed.  The default is http, with a specific exception being made for https.


Some notes for posterity after reading the spec:

> When choosing a URI scheme to associate with the app, apps MUST use a
   URI scheme based on a domain name under their control, expressed in
   reverse order.
   ...
   For example, an app that controls the domain name "app.example.com"
   can use "com.example.app" as their scheme.  
   
So the spec mandates the domain name having to be in reverse. 💯  
 
 However, according to the spec there should be only a single / after the scheme:
 
>  Following the requirements of Section 3.2 of [RFC3986], as there is
   no naming authority for private-use URI scheme redirects, only a
   single slash ("/") appears after the scheme component.  A complete
   example of a redirect URI utilizing a private-use URI scheme is:

     com.example.app:/oauth2redirect/example-provider

This isn't much of a problem as it's easy enough to filter out a double // & it'll likely be a common mistake to have two slashes after the scheme.

app/js/components/validator_helper.ts
Badlapje: aye, removed it :-)

PR #417, opened 2021-02-26
Ensure redirect url is correct for copy action

Badlapje:
Prior to this change, when a user tried to copy an entity and it went wrong: the redirect url was to the admin service overview.  This obviously led to an "access denied".

This change ensures that the redirect url is dependant on the user's access level.  Users get redirect to the service overview, admins to the admin service overview.

For more info see https://www.pivotaltracker.com/story/show/165965607


MKodde (approved):
Sweet and simple. You could try to cover this in a web test. I'm fine with this as is as well.

PR #418, opened 2021-03-01
Allow Resource Servers from outside of RP's team

MKodde:
This one's a doozy. The PR is twofold.

1. Allow read-only view of RS outside of team
The detail window of RP enitties now also show RP's from outside their own team. This is possible as there is no team restriction in Manage.

2. Prevent RS overwrites when outside of team 
SPD would overwrite these changes upon push to manage. That is also fixed in this PR.

More details: https://www.pivotaltracker.com/story/show/173371940

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/OidcClientInterface.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/OidcClientInterface.php
@@ -62,5 +62,10 @@ interface OidcClientInterface
 
     public function updateClientSecret(SecretInterface $secret): void;
 
-    public function merge(OidcClientInterface $client): void;
+    /**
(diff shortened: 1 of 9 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/OidcClientInterface.php
```diff
@@ -62,5 +62,10 @@ public function resetResourceServers(): void;
 
     public function updateClientSecret(SecretInterface $secret): void;
 
-    public function merge(OidcClientInterface $client): void;
+    /**
+     * Merges the new Oidc data with the existing data already present on the enitty.
```
Badlapje, line 66: nitpick: spellingsfout in entity

Badlapje (approved):
As always great work.  A small nitpick, but other than that no remarks. :shipit:

PR #419, opened 2021-03-01
Consider closed and resolved status from Jira

MKodde:
When service desk closes or resolves a Jira tikcet, but the matching
entity is still present in Manage, we should no longer consider it as
a request for delete status code.

https://www.pivotaltracker.com/story/show/176244612


Badlapje (approved):
Slick PR, :shipit: 💯

PR #420, opened 2021-03-01
Ensure attribute value is set on publishing entity

MKodde:
A missing attribute value will cause the system to break. Setting the default (a star) fixes this problem. This is implemented in the attribute generator.

In addition the Attribute list is now also dealing with new attributes correctly. It would previously only deal with existing attributes.

Relates to:
 - https://www.pivotaltracker.com/story/show/174549876
 - https://github.com/SURFnet/sp-dashboard/pull/413


Badlapje (approved):
Nice'n easy to review ❤️

PR #421, opened 2021-03-01
Refrain from overwriting Manage tracked attributes

MKodde:
It was believed the manage attributes where no longer overwritten. This was not the case. As the new logic does not actually check if the attribute is either manage or spd tracked. This is now fixed in the Json generator. Here we can use the attribute dictionary (repo) to get an intersection of the spd and manage managed attributes.


Badlapje (approved):
nice work!

PR #422, opened 2021-03-01
Build docker image when merging to develop.

quartje:
I'd like to be able to test the develop branch on a regular basis, but so far, only on release or on a commit in a pull_request a docker image was built. Now on all commits in the develop branch a docker image will be built


MKodde (approved):
LGTM!

PR #424, opened 2021-03-02
Resource server: Fix the type as set in manage

quartje:
Manage identifies a resource server as type: oauth20-rs and not oauth20_rs


MKodde (approved):
Maybe update the php comment above for an extra thumbsup ;) Other then that, looking golden!

PR #426, opened 2021-03-02
Show new production entity link for admins

Badlapje:
Prior to this change, admins could not see the "new production entity" link.  Added to that when the label text was semi-long the hover effect was a tad ugly.

This change ensures that admins can see the link & that the hover effect is ok for a longer label.


MKodde (approved):
Very nice! 
image

And thanks for the bonus bugfix! :deciduous_tree: boy scout rule for the win!
image

PR #427, opened 2021-03-02
Make tooltip for test entity clickable

Badlapje:
Prior to this change, the tooltip in the add test entity modal was not clickable ("what does this mean").

This change ensures that it is by adding a unique id.


MKodde (approved):
Works for me!

PR #428, opened 2021-03-02
Ensure modal redirects correctly when submitted

Badlapje:
Prior to this change, the form was sometimes evaluated as not submitted when it was ($form->isSubmitted() yielding the wrong result).  This resulted in an unstyled form for the modal on a blank page.

This change ensures that when the modal is "posted" (aka submitted), it'll be considered submitted & thus be redirected to the right page.


MKodde (approved):
Not stoked on the solution that was required here. But it gets the job done!

PR #429, opened 2021-03-02
Ensure entity copy results in a new entity

MKodde:
Without resetting the manage id, the action would result in an entity update action.


Badlapje (approved):
Looking good, :shipit:

PR #430, opened 2021-03-02
Hide content of tooltip

Badlapje:
Prior to this change, the content of the tooltip was visually hidden but still used up room.

This change ensures that when the tooltip is hidden the content is as well

### DIFF
--- a/app/scss/components/tooltip.scss
+++ b/app/scss/components/tooltip.scss
@@ -3,6 +3,10 @@
   max-height: 0;
   opacity: 0;
 
+  * {
+    display: none;
+  }
+
   @include motion {
     transition: max-height .4s, opacity .4s;
   }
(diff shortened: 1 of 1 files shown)


MKodde (changes requested):
I think you fixed the problem but might have introduced a new one. It might have already been an issue though.. 

When the `what does this mean` collapsible text is longer, the text wil flow over the form. As demonstrated on screenshot below

image

Badlapje:
> I think you fixed the problem but might have introduced a new one. It might have already been an issue though..

That was already an issue indeed.  The trick with the animation when opening/closing the modal requires a set max-height for the tooltip.  I had put it at 10vh given the text @quartje provided.  I just tweaked it a bit to 25vh, which should be more than enough.  

However, once that value is crossed this'll still be a problem.  On the other hand: if that value is crossed, there's likely way too much text in there.

MKodde (approved):
Ahh, much better. I agree with you on the 'too much text' explanation. If @quartje is okay with this, feel free to go ahead with this one!

PR #432, opened 2021-03-03
Remove form validation from CreateNewEntityType form

Badlapje:
Prior to this change, the validation would always throw an error for the create entity modal when submitting a test entity.  This is because if the form is not "submitted" it is considered invalid.

This change alters the validation to check whether there are actual form errors.  If none are found the form is considered valid.

Following the boyscout rule the ids for the titles were also made unique to the modal so they pass html validation.

**Edit @MKodde**
The solution to this problem is somewhat controversial, and can be considered bad practice. We created a TD ticket to later fix this issue. https://www.pivotaltracker.com/story/show/177183585


MKodde (approved):
Thanks! this should do the job, although I had a hard time reproducing the problem

PR #433, opened 2021-03-04
Do not verify service restraint on RS entities

MKodde:
The RS entities can be viewed from outside of the users team. This is a requirement.  But it was not allowed for production entities.

In addition two boy scout rule actions have been undertaken:
1. The test entities are now testing for explicit access to the other entity types (was only performed for prod entities)
2. Typehints where added to the method that was touched in this PR

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
@@ -33,6 +33,7 @@ use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Issue;
 use Surfnet\ServiceProviderDashboard\Infrastructure\Manage\Exception\QueryServiceProviderException;
 use Symfony\Component\Routing\RouterInterface;
+use function sprintf;
 
 /**
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
```diff
@@ -143,6 +146,14 @@ public function getEntityByIdAndTarget($id, $manageTarget, Service $service)
                 $entity = $this->queryRepositoryProvider
                     ->getManageTestQueryClient()
                     ->findByManageId($id);
+
+                // Allow actions on Resource Servers (viewing them outside of our team)
+                if ($entity->getProtocol()->getProtocol() !== Constants::TYPE_OPENID_CONNECT_TNG_RESOURCE_SERVER) {
```
Badlapje, line 151: not related to this PR, but `->getProtocol()->getProtocol()` seems like a bad name for either the first or the second?

Badlapje (approved):
❤️ the boyscout modifications 👍 

Aside from a nitpick, no comments so approved.  Feel free to ignore the nitpick.

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityService.php
MKodde: Yup, that an eyesore for me last couple of months (years even). It may be updated some time soon, but certainly not in this PR

MKodde:
@Badlapje I changed the code a bit after your initial review on account of a high cyclomatic complexity in a certain method. Can you please re-review?

Badlapje (approved):
that warning resulted in nice code, i like it!

PR #434, opened 2021-03-08
Install and configure opcache

quartje:
Add opcache to the FPM docker image: This speeds up the application considerably


MKodde (approved):
:racing_car:

PR #435, opened 2021-04-07
Merge comment (revision note) upon updating entity

MKodde:
The revision note was lost in the publication action when updating an existing entity.

This PR also addresses the incorrect Xdebug setup as we now get Xdebug3 vs 2.


quartje:
I've tested this and it seems to be working!

Badlapje (approved):
Yet again nice work 💯

quartje:
Seems to work. Xdebug also seems to try to connect to my laptop, which means that it works :)

quartje (approved):

PR #437, opened 2021-04-08
Ensure exclude from push coin value is honored

MKodde:
The exclude from push is set on the manage prod entity. And it was not
merged correctly onto the updated (SPD) entity that is to be published
to manage. This fixes that issue for the Coin value object.

https://www.pivotaltracker.com/story/show/177659443


Badlapje (approved):
Nice work!

PR #438, opened 2021-04-08
Correct publishability of production entities

Badlapje:
Prior to this change, production entities could not be published even if test entities were published.

This change ensures production entities can be published if test entities were published.

Pivotal issue at: https://www.pivotaltracker.com/story/show/177677810

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
@@ -19,6 +19,7 @@ namespace Surfnet\ServiceProviderDashboard\Application\ViewObject;
 
 use ArrayIterator;
 use IteratorAggregate;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
 
 class EntityList implements IteratorAggregate
 {
@@ -46,6 +47,19 @@ class EntityList implements IteratorAggregate
     public function getIterator()
     {
         return new ArrayIterator($this->entities);
+    }
+
+    public function hasTestEntities() : bool
+    {
+        foreach ($this->getEntities() as $entity) {
+            if ($entity->getEnvironment() === Constants::ENVIRONMENT_TEST
+                && $entity->getState() === Constants::STATE_PUBLISHED
+            ) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     public function sortEntitiesByEnvironment()
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
@@ -38,6 +38,11 @@ class Service
     private $privacyQuestionsEnabled;
 
     /**
+     * @var bool
+     */
+    private $productionEntitiesEnabled;
+
+    /**
      * @var EntityList
      */
     private $entityList;
@@ -47,20 +52,21 @@ class Service
      */
     private $router;
 
-    /**
-     * @param string $id
-     * @param string $name
-     * @param bool $privacyQuestionsEnabled
-     * @param EntityList $entityList
-     * @param RouterInterface $router
-     */
-    public function __construct($id, $name, $privacyQuestionsEnabled, EntityList $entityList, RouterInterface $router)
-    {
+    public function __construct(
+        String $id,
+        String $name,
+        bool $privacyQuestionsEnabled,
+        EntityList $entityList,
+        RouterInterface $router,
+        bool $productionEntitiesEnabled = false
+    ) {
+    
         $this->id = $id;
         $this->name = $name;
         $this->privacyQuestionsEnabled = $privacyQuestionsEnabled;
         $this->entityList = $entityList;
         $this->router = $router;
+        $this->productionEntitiesEnabled = $productionEntitiesEnabled;
     }
 
     public static function fromService(DomainService $service, EntityList $entityList, RouterInterface $router)
@@ -70,7 +76,8 @@ class Service
             $service->getName(),
             $service->isPrivacyQuestionsEnabled(),
             $entityList,
-            $router
+            $router,
+            $service->isProductionEntitiesEnabled()
         );
     }
 
@@ -112,5 +119,18 @@ class Service
     public function arePrivacyQuestionsEnabled()
     {
         return $this->privacyQuestionsEnabled;
+    }
+
+    public function hasTestEntities() : bool
+    {
+        return $this->getEntityList()->hasTestEntities();
+    }
+
+    /**
+     * @return bool
+     */
+    public function isProductionEntitiesEnabled()
+    {
+        return $this->productionEntitiesEnabled || $this->hasTestEntities();
     }
 }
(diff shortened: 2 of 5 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -318,8 +319,18 @@ public function adminOverviewAction($serviceId)
...
+        $hasTestEntities = false;
+
+        foreach ($serviceList as $service) {
+            if ($service->hasTestEntities()) {
+                $hasTestEntities = true;
+                break;
+            }
+        }
+
+        $productionEntitiesEnabled = [$productionEntityEnabled || $hasTestEntities];
```
MKodde, line 333: Again, maybe key this with the serviceId on the array. And while looking at this change. You should be able to move all the logic you added in the controllers to the service itself. Cleaning up the controller logic significantly.

MKodde (changes requested):
You are on the right track. I think you should be able to move the controller logic back into the Service ViewObject.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -128,7 +128,8 @@ public function overviewAction()
         $productionEntityEnabled = [];
         $privacyOK = [];
         foreach ($services as $service) {
-            $productionEntityEnabled[] = $service->isProductionEntitiesEnabled();
+            $productionEntityEnabled = $service->isProductionEntitiesEnabled();
+            $productionEntityEnabled[] = $productionEntityEnabled || $service->hasTestEntities();
```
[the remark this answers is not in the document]
Badlapje: It needs to be an array because both a user can have multiple services.  Each service has it's own "is production entities enabled" trigger.  In the twig view we loop over the services to present them, and then take the corresponding boolean from the array to know whether or not to portray the "add a production entity"-button.

The question is moot though, by moving the logic to the service view object the entire array is redundant and removed 💯

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
Badlapje: no longer need this array after moving the logic to the viewobject entirely 👍

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
```diff
@@ -48,6 +48,22 @@ public function getIterator()
...
     }
 
+    /**
+     * @return bool
+     */
+    public function hasTestEntities()
+    {
+        $hasTestEntities = false;
+        foreach($this->getEntities() as $entity) {
+            if ($entity->getEnvironment() === 'test' && $entity->getState() === 'published') {
```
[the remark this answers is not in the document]
Badlapje: fixed, good one :-)

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
```diff
@@ -48,6 +48,22 @@ public function getIterator()
         return new ArrayIterator($this->entities);
     }
 
+    /**
+     * @return bool
+     */
+    public function hasTestEntities()
```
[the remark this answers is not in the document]
Badlapje: fixed :-)

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
```diff
@@ -48,6 +49,21 @@ public function getIterator()
         return new ArrayIterator($this->entities);
     }
 
+    public function hasTestEntities() : bool
+    {
+        $hasTestEntities = false;
+        foreach ($this->getEntities() as $entity) {
+            if ($entity->getEnvironment() === Constants::ENVIRONMENT_TEST
+                && $entity->getState() === Constants::STATE_PUBLISHED
```
MKodde, line 57: :100:

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
```diff
@@ -48,6 +49,21 @@ public function getIterator()
...
+            ) {
+                $hasTestEntities = true;
+                break;
```
MKodde, line 60: Can be reduced to a `return true`, no need to keep a `$hasTestEntities` variable. But this is fine too!

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
```diff
@@ -53,14 +58,23 @@ class Service
...
-    public function __construct($id, $name, $privacyQuestionsEnabled, EntityList $entityList, RouterInterface $router)
-    {
+    public function __construct(
+        $id,
+        $name,
+        $privacyQuestionsEnabled,
+        EntityList $entityList,
+        RouterInterface $router,
+        $productionEntitiesEnabled = false
+    ) {
```
MKodde, line 70: You could have added type definitions, but this is good enough for rock and roll!

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
```diff
@@ -113,4 +128,20 @@ public function arePrivacyQuestionsEnabled()
...
+     */
+    public function hasTestEntities()
+    {
+        return $this->getEntityList()->hasTestEntities();
+    }
+
+    /**
+     * @return bool
+     */
+    public function isProductionEntitiesEnabled()
```
MKodde, line 143: Maybe add some return type definitions for the new boolean methods? /nitpick

MKodde (commented):
Great work! I have some additional nitpicks. But one more important feature is missing, and that is a unit test to test if the new logic is working as intended. But again, given the time constraint we find ourselves in..

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityList.php
Badlapje: good one! 🚀

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
Badlapje: done

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
Badlapje: while at it, good point!

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
```diff
@@ -47,20 +52,21 @@ class Service
...
-    public function __construct($id, $name, $privacyQuestionsEnabled, EntityList $entityList, RouterInterface $router)
-    {
+    public function __construct(
+        String $id,
+        String $name,
+        bool $privacyQuestionsEnabled,
+        EntityList $entityList,
+        RouterInterface $router,
+        bool $productionEntitiesEnabled = false
+    ) {
```
MKodde, line 62: Yourtheman!

MKodde (approved):
Feel free to merge, thanks for working with you today!

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/Service.php
Badlapje: but you're the master ;-)

PR #439, opened 2021-04-08
Ensure privacy link is visible for normal users

Badlapje:
Prior to this change, the privacy link was only visible to admin users.  Meaning that once the questions were answered, you could no longer go to the privacy questions.

This change ensures the link is now also visible for normal users (as was the original intent).  Additionally it also adds a link to the privacy questions in the tooltip for the status summary.

Pivotal ticket at https://www.pivotaltracker.com/story/show/177682438


quartje:
It's now visible for normal users, but not yet in warning mode when it's not yet filled in (no red background etc)

quartje:
I'm unable to reproduce that last issue.

PR #440, opened 2021-04-08
Repair faulty Jira ticket status implementation

MKodde:
Previously the status was added to the Jira Issue solution. This status is used to figure out the entity status.

Problem is that the status id was not yanked from the Jira status field, the entire status object was set on the type restricted Issue constructor.

While at it, a similar development issue was fixed.

https://www.pivotaltracker.com/story/show/177659652

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
@@ -29,6 +29,7 @@ class Issue implements JsonSerializable
 
     const STATUS_CLOSED = 'CLOSED';
     const STATUS_RESOLVED = 'RESOLVED';
+    const STATUS_OPEN = 'To Do';
 
     private $key;
 
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
@@ -22,6 +22,7 @@ use Surfnet\ServiceProviderDashboard\Application\Service\TicketServiceInterface;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Issue;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\IssueCollection;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Ticket;
+use function array_key_exists;
 
 class DevelopmentIssueRepository implements TicketServiceInterface
 {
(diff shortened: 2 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/IssueRepository.php
```diff
@@ -100,7 +100,7 @@ public function findByManageIds(array $manageIds)
         foreach ($issues->issues as $issue) {
             $manageId = $issue->fields->customFields[$this->manageIdFieldName];
             if (in_array($manageId, $manageIds)) {
-                $collection[$manageId] = new Issue($issue->key, $this->issueType, $issue->fields->status);
+                $collection[$manageId] = new Issue($issue->key, $this->issueType, $issue->fields->status->id);
```
MKodde, line 103: @quartje I guessed we need the `id` from the status field. I was not able to verify this as at the moment I have no Jira test environment on my dev machine. Would you be so kind to dump the output of `var_dump($issue); die;` placed on line 102 here if the id does not turn out to be something like 'OPEN', 'RESOLVED', or 'CLOSED'.

Peace out!

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
```diff
@@ -117,6 +118,9 @@ private function loadIssues(array $rawData)
     {
         $output = [];
         foreach ($rawData as $issueData) {
+            if (!array_key_exists(Issue::IDENTIFIER_TICKET_STATUS, $issueData)) {
+                $issueData[Issue::IDENTIFIER_TICKET_STATUS] = 'OPEN';
```
Badlapje, line 122: Should this not be a constant?

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
```diff
@@ -29,6 +29,7 @@ class Issue implements JsonSerializable
 
     const STATUS_CLOSED = 'CLOSED';
     const STATUS_RESOLVED = 'RESOLVED';
+    const STATUS_OPEN = 'To Do';
```
Badlapje, line 32: I'm a tad confused by this: there's a 'OPEN' ticket status in the file below, but here the status_open constant is assigned the value To Do?

Badlapje (approved):
It looks good to me: nice work!

Two small questions, but not sure about them, so i trust your judgement on them.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
MKodde: Sure! that should be the constant!

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/Issue.php
MKodde: @quartje provided an actual jira ticket dump, and turns out they use "To Do" for the open state. So, missed that on the DevIssueRepo.

PR #441, opened 2021-04-08
Ensure add production entity is always clickable

Badlapje:
Prior to this change, production entities could not be added because the link was not clickable in all situations.

This change ensures the link is always clickable.

Pivotal issue at: https://www.pivotaltracker.com/story/show/177677810


MKodde:
Nice! Im unable to functionally test this @quartje can you do this for me? I'm back to the stove. :cook:

MKodde (approved):

PR #442, opened 2021-04-09
Ensure ARP attributes is an object when empty

Badlapje:
Prior to this change, an empty attributes array got converted to an empty JSON array (makes sense).  However, when not empty the array got converted to a JSON object (as at that point it was an associative array before conversion).  The JSON schema for ARP specifies that the attributes property is an object, so the empty array did not get through schema validation.

This change replaces any empty attributes-array by a simple PHP stdClass (empty).  Which does get converted to an empty object in JSON.

Pivotal story at https://www.pivotaltracker.com/story/show/177697851


Badlapje:
After a positive test by @quartje on slack, merging without approval so we can release.

PR #443, opened 2021-04-12
Client secret reset service issue

MKodde:
The Organization value object is loaded from the manage provided metadata. This metadata might or might not include the Organization Name (NL/EN). We track the organization name on the Service entity. And that name is leading in the entities organization names. To ensure we have those org names on every Entity Manage interaction. I now update the OrgNames based on the service entity value. 

This happens to fix the issue where the client secret reset would fail on a missing org name value.

Bugfix for bug reported in: https://www.pivotaltracker.com/story/show/177701148


Badlapje (approved):
Nice work!  

I would add a remark in the PR description about equalizing the conditions for allowing a client reset, as that is not covered currently.

quartje:
almost there  ... When I do a client secret reset of a client that is connected to a resource server, the connection with that resource server disappears

MKodde:
Well spotted @quartje, I found the issue and wrote a fix for it.

quartje (approved):
I've tested it, all flows seem to work correctly now!

PR #445, opened 2021-04-13
Add client credential clients support

MKodde:
This introduces support for the oauth client credentials client entity type in SP dashboard. This entity type consists of a oidc tng entity that has the client credentials grant. This labels it internally as an `oauth20_ccc` entity.

https://www.pivotaltracker.com/story/show/177492575

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/OauthClientCredentialsClientJsonGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/OauthClientCredentialsClientJsonGenerator.php
@@ -1,0 +1,380 @@
+<?php
+
+/**
+ * Copyright 2021 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\Metadata;
+
+use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\PrivacyQuestionsMetadataGenerator;
+use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\SpDashboardMetadataGenerator;
+use Surfnet\ServiceProviderDashboard\Application\Parser\OidcngClientIdParser;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Contact;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\ManageEntity;
+
+/**
+ * The OauthClientCredentialsClientJsonGenerator generates oidc10_rp entity json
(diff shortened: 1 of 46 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/OauthClientCredentialsClientJsonGenerator.php
```diff
@@ -0,0 +1,380 @@
...
+        return array_filter($metadata);
+    }
+
+    private function generateContactMetadata($contactType, $index, Contact $contact): array
+    {
+        $metadata = [
+            sprintf('contacts:%d:contactType', $index) => $contactType,
+        ];
+
+        if (!empty($contact->getGivenName())) {
```
Badlapje, line 266: All these if statements can be abstracted into a method, they always follow the exact same pattern.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -87,15 +89,17 @@ class ManageEntity
...
     {
-        $manageProtocol = isset($data['type']) ? $data['type'] : '';
+        $manageProtocol = self::extractManageProtocol($data);
 
         $attributeList = AttributeList::fromApiResponse($data);
         $metaData = MetaData::fromApiResponse($data);
         $oidcClient = null;
         if ($manageProtocol === Protocol::OAUTH20_RS) {
-            $oidcClient = OidcngResourceServerClient::fromApiResponse($data, $manageProtocol);
+            $oidcClient = OidcngResourceServerClient::fromApiResponse($data);
```
Badlapje, line 98: A switch would be more readable here, as well as faster.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -285,7 +298,10 @@ public function merge(ManageEntity $newEntity)
         $this->metaData->merge($newEntity->getMetaData());
         $this->attributes->merge($newEntity->getAttributes());
         $protocol = $this->protocol->getProtocol();
-        if ($protocol === Constants::TYPE_OPENID_CONNECT_TNG || $protocol === Constants::TYPE_OPENID_CONNECT_TNG_RESOURCE_SERVER) {
+        if ($protocol === Constants::TYPE_OPENID_CONNECT_TNG
```
Badlapje, line 301: This if would be more readable if you made a method for it.

Badlapje (changes requested):
Nice work!  Looks good, save for a couple of smaller points.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
MKodde: Agreed

src/Surfnet/ServiceProviderDashboard/Application/Metadata/OauthClientCredentialsClientJsonGenerator.php
MKodde: I agree, but the method would be rather messy in my opinion. Using the accessors on the contacts would require using something like $contact->$contactGetterName(). That kinda sucks.

Badlapje: hm, a matter of preference whether it's messy or not (i find it a beautiful php feature).  If you dislike it though, it makes little sense to refactor.  I can live with how it is atm.

Badlapje (approved):
Looking good to me, nice work!  :shipit:

PR #450, opened 2021-05-03
Ensure that copied entities have a clear ACL

Badlapje:
Prior to this change, an entity based on a template would keep the ACL from the original entity.  This obviously should not be the case, as it's a new entity with it's own ACL.

This change ensures the new entity receives a fresh ACL.

Pivotal story available at: https://www.pivotaltracker.com/story/show/177884116

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/AllowedIdentityProviders.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/AllowedIdentityProviders.php
@@ -92,4 +92,10 @@ class AllowedIdentityProviders
         $this->providers = is_null($allowedIdPs->getAllowedIdentityProviders()) ? null : $allowedIdPs->getAllowedIdentityProviders();
         $this->allowAll = is_null($allowedIdPs->isAllowAll()) ? null : $allowedIdPs->isAllowAll();
     }
+
+    public function clear()
+    {
+        $this->allowAll = true;
+        $this->providers = [];
+    }
 }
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -180,6 +180,11 @@ public function getAllowedIdentityProviders()
         return $this->allowedIdentityProviders;
     }
 
+    public function clearAllowedIdentityProviders()
+    {
+        $this->allowedIdentityProviders = new AllowedIdentityProviders([], true);
+    }
```
MKodde, line 186: Gets the job done for certain. Have you also considered adding a `clear` method to the `AllowedIdentityProviders` VO? That could unset all data and allows the VO itself to determine what a cleared state really means. That way you do not actually need to add this method, but use the getter and call `clear` on the VO.

MKodde (approved):
I approve of this, If you feel like it you could implement my proposition. Leaving it like this is also A-OK!

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
Badlapje: i had not, but i like the idea :-)

It has been done 🧙

MKodde (approved):
Grandslam!

quartje (approved):
Testing indicates that no acl is present on production entities anymore!

PR #451, opened 2021-05-04
Ensure attributes are validated also on test

Badlapje:
Prior to this change, leaving an arp motivation empty was possible on test, but not on production.  However, an attribute with an empty motivation was not added to the ARP.  Whilst privacyfriendly, this didn't match the expectations.

This change uses the same validation for production as for test when it comes to attributes.  Given that production and test now have the same behaviour things should be clearer.

Pivotal story: https://www.pivotaltracker.com/story/show/177774124


MKodde (approved):
Sweet and simple! I'd like to see the story updated as the actual change here does not reflect the story written there.

Badlapje:
> Sweet and simple! I'd like to see the story updated as the actual change here does not reflect the story written there.

as you request, so has it been done, my liege

quartje (approved):

PR #452, opened 2021-05-05
Prevent calling count on null

MKodde:
By setting the default value of the issues collection to be an array.
Count function can be used safely without warning. As calling count on
null started yielding warnings in PHP7.

Resolves first warning of:
https://www.pivotaltracker.com/story/show/177742668


Badlapje (approved):
Well spotted!  :shipit:

quartje:
The message is gone from the logs!

PR #461, opened 2021-05-25
Hide edit for PE on removal requested

Badlapje:
Prior to this change, the edit option was visible for entities that had the status of removal requested.  Upon clicking the edit button, an access denied screen was shown.  This behaviour was right, but the option to click edit should not be present.

This change removes the edit option.

Pivotal ticket: https://www.pivotaltracker.com/story/show/178034009

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
@@ -83,12 +83,14 @@ class EntityActions
         return $this->environment;
     }
 
-    /**
-     * @return bool
-     */
-    public function allowEditAction()
+    public function allowEditAction(): bool
     {
-        if ($this->isPublishedToProduction() || $this->readOnly) {
+        $notEditable =
+            $this->isPublishedToProduction()
+            || $this->readOnly
+            || $this->status === Constants::STATE_REMOVAL_REQUESTED;
+
+        if ($notEditable) {
             return false;
         }
         return true;
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
@@ -181,7 +181,7 @@ class ManageEntity
 
     public function isPublished()
     {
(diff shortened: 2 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -181,7 +181,7 @@ public function getStatus()
 
     public function isPublished()
     {
-        return ($this->status === 'published');
+        return ($this->status === Constants::STATE_PUBLISHED);
```
MKodde, line 184: Good boyscout!

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
```diff
@@ -83,12 +83,14 @@ public function getEnvironment()
...
-     * @return bool
-     */
-    public function allowEditAction()
+    public function allowEditAction(): bool
     {
-        if ($this->isPublishedToProduction() || $this->readOnly) {
+        $notEditable =
+            $this->isPublishedToProduction()
+            || $this->readOnly
+            || $this->status === Constants::STATE_REMOVAL_REQUESTED;
```
MKodde, line 91: Instead of catching this in a notEditable variable, you could directly return the outcome of this expression.

MKodde (commented):
Nice work, and thanks for saving the SPD environment mister boyscout!

One thing I'd like to see added to this PR is an additional test covering your change in: `tests/integration/Application/ViewObject/EntityActionsTest.php`

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
Badlapje: I actually did that first, but found it far less legible to the point that i wanted to add a comment to it to clarify.  So went with this as a middle ground.

MKodde: I can live with that! And it improves debugabillity as well.

MKodde (approved):
Great! Thanks for expanding tests

PR #462, opened 2021-05-25
Add basis for cypress tests

Badlapje:
Prior to this change, we could not adequately perform end-to-end tests in a maintainable manner.

This change adds the cypress framework, so we can start working on tests.

Pivotal issue: https://www.pivotaltracker.com/story/show/178051955


MKodde (commented):
Now that you are setting up Cypress into the project. May I suggest some thingies?

1. If possible I'd love to see the cypress folder being moved to the `/tests` folder
2. The test strategy needs to be updated. Maybe we should discuss what weight we want to put on web tests compared to cypress tests.
3. That's it for now

Badlapje:
fyi, in case this can be of use: i created a cypress plugin that can handle sso authentication with redirects.: https://github.com/Badlapje/cypress-social-logins.  I still need to clean it up, but using a customLogin function (as in the readme) it should also work for spdashboard

PR #463, opened 2021-05-25
Ensure name in En/Nl is mandatory

Badlapje:
Prior to this change the names of the organization were not mandatory.

This change makes them mandatory.

Pivotal story at: https://www.pivotaltracker.com/story/show/177659336


Badlapje:
> Am I correct in my assumption that the PHP validation did not allow a blank value for the NameNL/EN? If thats not the case I'd like us to check the PHP validation.

You are correct indeed.

MKodde (approved):

PR #464, opened 2021-05-25
Add organization unit attribute

Badlapje:
Prior to this change, the attribute wasn't supported by SPD.

This change adds the attribute.

Pivotal ticket: https://www.pivotaltracker.com/story/show/177557987


src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityDetail.php
```diff
@@ -728,4 +742,12 @@ public function getResourceServers()
     {
         return $this->resourceServers;
     }
+
+    /**
+     * @return Attribute|null
+     */
```
MKodde, line 748: Maybe replace the PHPDoc with actual type hints?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -133,6 +134,7 @@ entity:
         prefered_language: Preferred language attribute
         personal_code: Personal code attribute
         scoped_affiliation: Scoped affiliation attribute
+        organization_unit: Organization unit
```
MKodde, line 137: Given we no longer support `oidc` entitites, this entry can be omitted. And the entire section can be wiped I assume. But feel free to only refrain from adding the organizationion_unit here.

MKodde (commented):
Two small things to look at, other then that, looks great!

And I'm pleased to announce you've won the medal of typehints!

:medal_sports:

MKodde:
Thanks for the changes! I see there is a warning for the unused resource servers in CI. That seems to be a glitch as they are used indeed. Let's get rid of this warning, can you see to it?

MKodde (approved):

PR #467, opened 2021-05-27
Prevent double JQuery installations

MKodde:
Two Jquery versions cause problems when Webpack encore serves off Jquery
to the javascript. Pinning Jquery to the known working JQ version solves
that issue.

https://www.pivotaltracker.com/story/show/178316980


Badlapje (approved):
Quicker than his own shadow to fix a bug: nice work!

quartje (approved):
works!

PR #468, opened 2021-05-27
Ensure logging out works & redirects

Badlapje:
Prior to this change, there was no logout & the redirect did not work.

This change adds the logout behaviour & redirects to the url as specified by the user.

Pivotal story @ https://www.pivotaltracker.com/story/show/173125614

### DIFF
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -29,7 +29,7 @@
                     <li class="border-left">
                         <a href="{{ 'page.header.help_link'|trans }}" target="_blank">{{ 'page.header.help'|trans }}</a>
                     </li>
-                    <li class="border-left"><a>{{ 'page.header.logout'|trans }}</a></li>
+                    <li class="border-left"><a href={{ path('logout') }}>{{ 'page.header.logout'|trans }}</a></li>
                     <li class="service-switcher">
                         {{ service_switcher() }}
                     </li>
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -35,13 +35,6 @@ security:
         saml_based:
             saml: true
             logout_on_user_change: true
-            logout:
-                path: /logout
-                target: "%logout_redirect_url%"
-                invalidate_session: true
-                csrf_token_generator: security.csrf.token_manager
-                csrf_parameter: 't'
-                csrf_token_id: 'dashboard_logout_token'
 
     access_control:
         - { path: ^/saml, roles: IS_AUTHENTICATED_ANONYMOUSLY, requires_channel: https }
@@ -58,6 +51,7 @@ nelmio_security:
         log: true
         whitelist:
             - '%saml_remote_idp_host%'
+            - '%logout_redirect_url%'
     csp:
         report_logger_service: monolog.logger.security
         hosts: []
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -18,3 +18,10 @@ nelmio_security_csp:
     path:     /csp/report
     defaults: { _controller: nelmio_security.csp_reporter_controller:indexAction }
     methods:  [POST]
+
+logout:
+    path: /logout
+    defaults:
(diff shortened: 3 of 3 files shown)


app/Resources/views/base.html.twig
```diff
@@ -29,7 +29,7 @@
                     <li class="border-left">
                         <a href="{{ 'page.header.help_link'|trans }}" target="_blank">{{ 'page.header.help'|trans }}</a>
                     </li>
-                    <li class="border-left"><a>{{ 'page.header.logout'|trans }}</a></li>
+                    <li class="border-left"><a href="/logout">{{ 'page.header.logout'|trans }}</a></li>
```
MKodde, line 32: Why not use the `path` method here for the logout anchor?

Example:
`<a href="{{ path('entity_acl', {id: action.id, serviceId: action.serviceId}) }}">`

app/config/security.yml
```diff
@@ -37,11 +37,8 @@ security:
             logout_on_user_change: true
             logout:
                 path: /logout
-                target: "%logout_redirect_url%"
+                target: logout_redirect
```
MKodde, line 40: This feels like a over-complicated solution. 

1. We are not actually logging out, so using the built in logout feature might not be that useful. When you return to SPD after visiting the page. You have not lost your SSO.
2. Maybe just create a single logout action (like you did) and return a redirect response there.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SecurityController.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Routing\Annotation\Route;
+
+class SecurityController extends AbstractController
+{
+    /**
+    * @Route("/logout", name="app_logout", methods={"GET"})
```
MKodde, line 12: If you keep this construction in favour of my suggestion then please:

1. Move this route annotation to the routes.yml
2. Remove the app_ prefix from the route name
3. +1 for adding the comment, explaining why this was added. But in my opinion this indicates a smelly solution.

MKodde (changes requested):
Thanks for you effort so far. See my points below. I think you can simplify this quite easily from here.

app/Resources/views/base.html.twig
Badlapje: What's the benefit of using it over a simple link?  It adds computation to delivering the html, when we will get the exact same link.  And it's not as if this link will change.  I assume I'm missing someting?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SecurityController.php
Badlapje: Removed this as per your suggestion above.

app/config/security.yml
Badlapje: Good points.  I simplified it and removed the logout logic in the security.yml (as it'd conflict otherwise).  Nice'n simple that way 💯

app/Resources/views/base.html.twig
MKodde: Bigest gain is, when changing the destination of the route, you don't need to update your 'href' components. It also provides a very clean way to pass parameters and other route related modifiers.

Badlapje: Not sure i'd agree that this case warrants it.  But adjusted to comply with best practice 👍

MKodde:
Good to go!

MKodde (approved):

PR #469, opened 2021-05-28
Separate modal oidc for environment

Badlapje:
Prior to this PR there was only one message for both environments to which a OIDC was published.

This PR ensures that a separate message is possible, depending on the environment.  In all other respects the modal remains the same.
It also applies the boyscout rule to fix a broken translation.

Pivotal ticket at https://www.pivotaltracker.com/story/show/175656934

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
@@ -173,7 +173,8 @@ entity.list.modal.oidc_confirmation:
   client_secret:
     label: Secret
   warning: This message will be displayed only once
-  info.html: <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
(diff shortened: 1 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -177,7 +173,8 @@ entity.list.modal.oidc_confirmation:
...
+  info.html.production: <p>PRODUCTION Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+  info.html.test: <p>TEST Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
```
MKodde, line 177: I like your solution, straight forward and concise!

One thing might cause problems. And that is that the translation string no longer ends in 'html'. And that suffix is evaluated in the online translation environment, it enables a WYSIWYG editor. I was pleasantly surprised we documented this oddity `https://github.com/SURFnet/sp-dashboard/blob/90aec1c589c5761e64f24374dd1b0bb569dd123b/docs/translations.md`.

MKodde (changes requested):

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
Badlapje: wel that's weird, but fixed it 👍 Good catch!

MKodde (approved):
Great, thanks for addressing the request for change!

PR #470, opened 2021-05-29
Add a global site notice

Badlapje:
Prior to this change, there was no way to alert the user with temporary content.  For example: what changed in the last release.

This change adds a global site notice.  It uses two parameters to set the allowed html & whether or not to display the notice.  The content of the notice itself can be set in the translation.

Pivotal ticket at: https://www.pivotaltracker.com/story/show/176902060


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SiteNoticeController.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet B.V.
```
MKodde, line 4: Nit: ~~2017~~ 2021

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SiteNoticeController.php
```diff
@@ -0,0 +1,75 @@
...
+     * This desire to not see the notification is stored both in the
+     * session and in cookies.
+     *
+     * @Method("GET")
+     * @Security("has_role('ROLE_USER')")
+     * @Template()
+     */
+    public function showGlobalSiteNoticeAction()
+    {
+        $date = $this->getParameter('global.site_notice.date');
```
MKodde, line 44: I would like to see the parameters DI'ed into the controller. See the EntityDetailController for an example if you need one.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SiteNoticeController.php
```diff
@@ -0,0 +1,75 @@
...
+        $hasBeenClosed = (bool)$this->get('session')->get('site_notice.closed.' . $date);
```
MKodde, line 45: Consider injecting the session into the Controller. I see we used `this->get` on other controllers, but this is discouraged as we prefer dependency injection over grabbing stuff from the container.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Twig/GlobalSiteNotice.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet B.V.
```
MKodde, line 4: Same nit as before :beetle:

MKodde (changes requested):
Nice work again. Some things might need your attention.In addition to the code review actions you might want to update the story according to the new insights, requirements and improvements that have been implemented. 

This request was triggered by my limited understanding of the notice.date component that was only hinted on in the Pivotal story. Is this date added a 'show from this date and forward' date? Or show until, or ... In code I see the date added to the session.

And finally, you mentioned the closing trigger will also be stored in cookie. But I do not see the creation of an explicit cookie other than the session cookie. Is that also intentional?

This might need some additional documentation.

Disclaimer: I have not tested it in my browser yet.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/SiteNoticeController.php
Badlapje: quite a few lines of code removed thanks to this suggestion: nice one!

Badlapje:
> Nice work again. Some things might need your attention.In addition to the code review actions you might want to update the story according to the new insights, requirements and improvements that have been implemented.

I agree there, but i'm awaiting feedback from @teunfransen.  Saves me having to edit it twice ;-)

MKodde (approved):

PR #471, opened 2021-05-31
Fix user redirect after deleting published entity

Badlapje:
Prior to this change, users who deleted a published entity were redirected to /service/{id}, for which one needs admin access.  This resulted in an error.

This change ensures that the redirection takes into account their access level.  Normal users are now redirected to '/', admin users see no change.

Pivotal ticket: https://www.pivotaltracker.com/story/show/178034022


MKodde (approved):
Super! Did you by any chance check the other actions for similar issues?

Badlapje:
> Super! Did you by any chance check the other actions for similar issues?

I did indeed.  Could not find any other errors.

PR #472, opened 2021-06-03
Lowercase client ids and redirect url

Badlapje:
Prior to this change, Client Ids and redirect url could be a mix of lowercase and uppercase.

This change lowercases the redirect urls and the client ids in the saveOidcNgEntity.  It also applies the boyscout rule and fixes an error when viewing the oidc entities.

No tests were added here as i did not find an existing test case.  But they will be added when the cypress ticket is implemented.

Pivotal ticket @: https://www.pivotaltracker.com/story/show/173276689

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
@@ -459,7 +459,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
      */
     public function setEntityId($entityId)
     {
-        $this->entityId = strtolower($entityId);
+        $this->entityId = $entityId;
     }
 
     /**
@@ -491,11 +491,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
      */
     public function getRedirectUrls()
     {
-        if (!is_array($this->redirectUrls)) {
-            return [];
-        }
-
-        return array_values($this->redirectUrls);
+        return $this->redirectUrls;
     }
 
     /**
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
@@ -459,7 +459,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
      */
     public function setEntityId($entityId)
     {
-        $this->entityId = $entityId;
+        $this->entityId = strtolower($entityId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -503,7 +503,11 @@ public function getRedirectUrls()
      */
     public function setRedirectUrls($redirectUrls)
     {
-        $this->redirectUrls = $redirectUrls;
+        $urls = []; // because numeric array keys can be sparse so a for-loop cannot be trusted
+        foreach ($redirectUrls as $url) {
+            $urls[] = strtolower($url);
+        }
```
MKodde, line 509: Maybe an `array_walk` solution is a bit nicer here. But this works well too!

Alternatively to ensure you preserve the keys you could iterate over the urls with a key->value construction.

```
foreach ($redirectUrls as $key => $url) {
    $urls[$key] = strtolower($url);
}

**Review by MKodde — APPROVED** [review]
I think this solution is good! Maybe, to stick more closely to our layered architecture, would it be possible to move this logic to the domain layer? As this can be considered a business rule.

File: src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php (hunk 1 above)
Comment (Badlapje) [reply src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php:509]: I actually used this because i didn't want to preserve the keys.  It's weird behaviour for it to not reorder the keys.  Can lead to subtle bugs & given that they are numeric (and they're not used) there's no reason for us not to reorder.

**Comment by Badlapje** [comment]
> I think this solution is good! Maybe, to stick more closely to our layered architecture, would it be possible to move this logic to the domain layer? As this can be considered a business rule.

It has been done.  Good call!   You probably won't be able to check this before it's merged, but if need be i can amend it on monday in a new PR.

File: src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php [hunk 2]
```diff
@@ -491,19 +491,19 @@ public function setSecret($secret)
      */
     public function getRedirectUrls()
     {
-        if (!is_array($this->redirectUrls)) {
-            return [];
-        }
```
MKodde, line 496: By removing this, you excavated a somewhat intricate buggy behavior. 

1. WebTests are unable to handle form interactions with the Redirect URLs as they are JavaScript based and our curl based test browser is not capable of JS interactions.
2. By removing the !is_array construction, the correct: 'at least one redirect url is required' form error is triggered on the form. This halts execution of the test.
3. Previously we would return an empty array, this somehow passed validation. And the web test was satisfied. 
4. In dev/prod environments this behavior would trigger an error in Manage. As it does not allow empty or unset redirect urls.

Hope this makes some sense.. 

Possible solutions include:
- to add a no-js solution for the redirect urls
- move this test to the cypress test suite
- implement real browser support into the web-tests (should be easy enough using [Panther](https://github.com/symfony/panther))
- insert your genius solutions here..

PR #473, opened 2021-06-08
Finetune organizational unit name attribute

Badlapje:
Prior to this change, the attribute had the wrong name: "organization unit" rather than "organizational unit name".  There was also an error when using the view action.

This change alters the name & fixes the bug.  It also applies the boyscout rule and ensures all types for attributes are correct & use type hinting in the getters.

Pivotal ticket @: https://www.pivotaltracker.com/story/show/177557987


MKodde (approved):
Renaming the attribute properties is a solid effort, renaming the translation is not a necessity, as they can be translated in the translation webinterface.

Anyway, thanks for the change!

PR #475, opened 2021-06-14
Improve global site notice

Badlapje:
Prior to this change, there were some features missing for the global site notice.

This change adds the following:
- wysiwyg editor for translation
- ensure button is on a separate row by wrapping it in a paragraph

Pivotal ticket: https://www.pivotaltracker.com/story/show/176902060


pablothedude (approved):
I'm not allowed to see the Pivotal issue, but the change does what it says in the PR description.

PR #476, opened 2021-07-14
Lowercase protocol & hostname in redirect url

Badlapje:
Prior to this change the entire redirect url was lowercased.  That wasn't the idea but not entirely clear in the orinal feature request.

This change lowercases only the correct parts.

See pivotal comment / ticket @ https://www.pivotaltracker.com/story/show/173276689/comments/225087260

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
@@ -459,7 +459,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
      */
     public function setEntityId($entityId)
     {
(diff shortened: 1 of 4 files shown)


pablothedude (commented):
I'm wondering if this solution would not result in duplicate redirect uri's. Please check what will happen if a user enters the same uri but with different casings. You could check the `UniqueRedirectUrlsValidator` and the `ValidRedirectUrlValidator` for this.

pablothedude:
:+1: the change seems to work as expected.

Badlapje:
> 
> 
> I'm wondering if this solution would not result in duplicate redirect uri's. Please check what will happen if a user enters the same uri but with different casings. You could check the `UniqueRedirectUrlsValidator` and the `ValidRedirectUrlValidator` for this.

still catches duplicate urls
image

pablothedude (commented):
Thanks for checking. All looks good to me!

PR #477, opened 2021-07-14
Only update acl fields on edit idp whitelist

Badlapje:
Prior to this change, the entire metadata for the entity was sent to mange just to update the acl.  This lead to some subtle bugs because too much was updated.  The user only updates the acl, so we should not send more than that to manage.

This change limits the data sent to manage to the data pertinant to the acl change.

Pivotal ticket @ https://www.pivotaltracker.com/story/show/178461498
API info for manage @ https://github.com/OpenConext/OpenConext-manage/wiki/API#merge-write

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
@@ -67,11 +67,14 @@ class JsonGenerator implements GeneratorInterface
         ];
     }
 
-    public function generateForExistingEntity(ManageEntity $entity, string $workflowState): array
-    {
+    public function generateForExistingEntity(
+        ManageEntity $entity,
+        string $workflowState,
+        string $updatedPart = ''
+    ): array {
         // the type for entities is always saml because manage is using saml internally
         $data = [
-            'pathUpdates' => $this->generateDataForExistingEntity($entity, $workflowState),
+            'pathUpdates' => $this->generateDataForExistingEntity($entity, $workflowState, $updatedPart),
             'type' => 'saml20_sp',
             'id' => $entity->getId(),
         ];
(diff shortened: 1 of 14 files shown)


quartje:
Good work! It works as expexted, and will not only fix this bug, but also make the ACL update bit more robust in the future.
I guess that the next candidate for this will be the secret reset for oidc RP's and resource servers.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -67,11 +67,14 @@ public function generateForNewEntity(ManageEntity $entity, string $workflowState
         ];
     }
 
-    public function generateForExistingEntity(ManageEntity $entity, string $workflowState): array
-    {
+    public function generateForExistingEntity(
+        ManageEntity $entity,
+        string $workflowState,
+        string $updatedPart = ''
```
pablothedude, line 73: Wasn't it possible to add a dedicated ACL JsonGenerator which implemented the GeneratorInterface to have better separation of concerns? Adding a parameter like this seems a code smell to me.
Please look into this if this if you see fit to change this.

pablothedude (changes requested):
Changes look good to me, please check if you could implement the code while implementing the GeneratorInterface.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
Badlapje: I would have no idea on how to do that.  Existing generators are for entity types (oidcng/saml/...), not for parts of entities like ACL.

My idea was to go one step at the time towards reforming the updates all-together to only send the necessary data.  Right now updates send the entire entity, rather than what's been changed.  This PR is the first one to alter that.  ACL updates now only send ACL data.  I think the way forward, which is out of scope for now, is to get rid of the `generateForExistingEntity` method on the generatorinterface.  And replace that with dedicated methods depending on what gets changed.  Exactly to separate concerns for the different update methods.

That is not possible within the budget remaining however, as that'd require a rather big overhaul if we did it all in one go.

pablothedude: I agree to change the GeneratorInterface and I personally think this would be the time to change it to prevent introducing technical depth in the first place. But feel free to leave it as is for now. But would you then create a TD ticket in the backlog to secure this discussion in Pivotal. Thanks!

pablothedude (approved):

Badlapje:
Ticket has been added in Pivotal: https://www.pivotaltracker.com/story/show/178946121

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
Badlapje: Ticket added

PR #479, opened 2021-08-25
Show correct translations for oidc attributes

quartje:
The oidcng entry form showed the translations for the saml form


Badlapje (approved):
Nice work 👍

PR #485, opened 2021-10-13
Store the sessions in the database: Nice for HA setups

quartje:
I've tried setting up session storage in the database, since I want to make the SPdashboard suitable for using in a multi node Docker setup, without having to worry about session affinity or saving session files during restarts. 
This setups works, and would even be nicer if we can reuse the Doctrine database settings. Is that possible in any way?

### DIFF
--- a/app/DoctrineMigrations/Version20211013095152.php
+++ b/app/DoctrineMigrations/Version20211013095152.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20211013095152 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+	// this up() migration is auto-generated, please modify it to your needs
+	$this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+        $this->addSql('
+        CREATE TABLE `sessions` (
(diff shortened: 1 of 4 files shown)


thijskh:
Put `database_url` in the .env instead of in config.yml so that is used for doctrine aswell as this?

MKodde (commented):
I would approve this approach as it follows the Symfony best practice. But I have to agree with @thijskh, that adding an additional connection string to the application is not ideal.

You could consider using a Doctrine based session handler like `https://packagist.org/packages/shapecode/doctrine-session-handler-bundle`.

app/DoctrineMigrations/Version20211013095152.php
```diff
@@ -0,0 +1,40 @@
...
+        $this->addSql('
+	   CREATE TABLE `sessions` (
+           `sess_id` VARBINARY(128) NOT NULL PRIMARY KEY,
+           `sess_data` BLOB NOT NULL,
+           `sess_lifetime` INTEGER UNSIGNED NOT NULL,
+           `sess_time` INTEGER UNSIGNED NOT NULL,
+           INDEX `sessions_sess_lifetime_idx` (`sess_lifetime`)
+	   ) COLLATE utf8mb4_bin, ENGINE = InnoDB;
+        ');
+
```
MKodde, line 29: Double nit, feel free to ignor:
1. Maybe indent the SQL 8 spaces to improve readability
2. Remove the linefeed on line 29

MKodde:
While migrating from Symfony 3 to 4, we opted to keep using the 'legacy' parameters.yaml construction that was the default in earlier Symfony version. 

I would applaud moving away from this construction and utilize a .env based solution. But I also agree with Bart when he adds the connection string to the config file based on the existing parameters, provided by the parameters.yaml.

app/DoctrineMigrations/Version20211013095152.php
quartje: :heavy_check_mark:

PR #490, opened 2022-01-12
Fix modal problems

Badlapje:
Prior to this change, there was a regression regarding the modal functionality.  This was broken everywhere save for the membership management.

This change ensures all modals work.

See pivotal @:
- https://www.pivotaltracker.com/n/projects/1400064/stories/180874369
- https://www.pivotaltracker.com/n/projects/1400064/stories/180874379


quartje (approved):
This seems to fix both issues. Let's tag & deploy!

PR #495, opened 2022-03-01
Log authenticated user

thijskh:
While researching a reported issue by a user it was not clear which messages in the logs to look for. Log who authenticates so we can quickly find associated log lines with that session.


quartje (approved):
Works as expected!

PR #500, opened 2022-05-03
Fix some of the testsuite

thijskh:
The testsuite was broken, as it seems since the Teams integration. In this PR I've managed to tackle at least some of the issues. Althoug the build still does not pass yet, this is an improvement over the previous amount of failing tests.


thijskh:
👍

MKodde:
The tests webtests are back in working order.

PR #501, opened 2022-05-03
Add the teamId also to the service switcher

thijskh:
As an administrator, I sometimes want to be able to find a space when I know only the teamId. Add the most significant part of the teamId urn to the service switcher so you can find services also in that way.


MKodde (approved):
A nice addition, the failing web tests have to do with the web tests not using a mock API yet. They try to retrieve data from an actual teams instance, and that probably fails. Merging this, and will fix the failing web-tests in: `tests/fix-build`

PR #502, opened 2022-05-12
Fix visual bug of removal of the wrong team member after deletion

VadimSchmitz:
Work done in this pull request
- Fixed visual bug of removal of the wrong team member after deletion.
- Update teams.md to explain how to add members.

https://www.pivotaltracker.com/story/show/181418526

### DIFF
--- a/app/js/components/delete_member.ts
+++ b/app/js/components/delete_member.ts
@@ -11,7 +11,7 @@ $(() => {
       if (!target.matches('.teams__deleteMemberLink')) return;
       event.preventDefault();
 
-      const td = document.querySelector('.teams__actions') as HTMLTableDataCellElement;
+      const td = target.closest('td') as HTMLTableDataCellElement;
       td.classList.add('loading');
 
       const url = target.getAttribute('href') || '';
--- a/docs/teams.md
+++ b/docs/teams.md
@@ -41,11 +41,14 @@ From the cli of your host: `docker-compose exec openconext tail -50  /var/log/te
 The docker container does not contain a functioning email service, and so sending an invite / creating a team does not actually send a mail.
 
 Therefore: to test with an actual member, you need to do the following:
-1. open the openconext docker cli
-2. go to mysql (`mysql` on the cli)
-3. select a uiid for one of the invited people: 
-   1. `use teams;`
-   2. `select invitation_uiid from invitations where id = ;` => fill in the id
-4. accept the invite by going to the following url: `https://teams.vm.openconext.org/invitation/accept/$uiid_of_the_invite` where you replace the last bit with the uiid found in the previous step.
+1. Find the invitation_uiid of the user you want to accept `docker-compose exec openconext mysql -e "select * from invitations \G" teams`
+When accepting a new invite a new private window needs to be opened, otherwise the following error will show: "You are already a member of this Team."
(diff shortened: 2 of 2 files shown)


app/js/components/delete_member.ts
```diff
@@ -11,8 +11,8 @@ $(() => {
       if (!target.matches('.teams__deleteMemberLink')) return;
       event.preventDefault();
 
-      const td = document.querySelector('.teams__actions') as HTMLTableDataCellElement;
-      td.classList.add('loading');
+      const td = $(target).closest('td');
```
Badlapje, line 14: why are you reverting to jQuery here?  We were phasing out that dependency, are we returning to it?

app/js/components/delete_member.ts
```diff
@@ -11,8 +11,8 @@ $(() => {
...
+      td.addClass('loading');
```
Badlapje, line 15: Why are you reverting to jQuery here?  We were phasing out it's usage when I was working on it, has that decision been reversed?

Badlapje: You can likely use the target. variable, defined on line 10 to navigate to the closest td and delete it without using jQuery?

docs/teams.md
```diff
@@ -41,11 +41,12 @@ From the cli of your host: `docker-compose exec openconext tail -50  /var/log/te
...
-1. open the openconext docker cli
-2. go to mysql (`mysql` on the cli)
-3. select a uiid for one of the invited people: 
-   1. `use teams;`
-   2. `select invitation_uiid from invitations where id = ;` => fill in the id
-4. accept the invite by going to the following url: `https://teams.vm.openconext.org/invitation/accept/$uiid_of_the_invite` where you replace the last bit with the uiid found in the previous step.
-
-If you use the cli to connect to docker: `docker-compose exec openconext mysql -e 'use teams; select invitation_uiid from invitations where id = ;'` & so on.
+1. Find the invitation_uiid of the user you want to accept `docker-compose exec openconext mysql -e "select * from invitations \G" teams`
+When accepting a new invite a new private window needs to opened, otherwise the following error will show: "You are already a member of this Team."
```
Badlapje, line 45: "needs to open" instead of "needs to be opened"

docs/teams.md
```diff
@@ -41,11 +41,12 @@ From the cli of your host: `docker-compose exec openconext tail -50  /var/log/te
...
+2. Accept the invite by going to the following url: `https://teams.vm.openconext.org/invitation/accept/"invitation_uiid"` where you replace the text between the "" with the invitation_uiid found in the previous step.
+
+Once this had been done once you can make the team public so adding new members will be easier.
```
Badlapje, line 48: "has" instead of "had"

second "once" can be removed (word 6)

Badlapje (changes requested):
Looks good to me, so the root issue the wrong element was selected for deletion.  

I'd preserve the jquery-less style unless we decided to go back to jQuery (in general for all things Surf we were moving away from jQuery as a dependency).

app/js/components/delete_member.ts
VadimSchmitz: I was not informed that jquery was being phased out, I will revert this back to regular js.

docs/teams.md
VadimSchmitz: nicely spotted aswell

Badlapje (approved):
Nice work! :shipit:

PR #503, opened 2022-05-18
Fix missing team role icons and make multiple ui improvements

VadimSchmitz:
-Rework the drop downs to no longer use selects, in order to be able to style them better & use icons.
-Fix issues with styling to be more inline with other drop downs.

https://www.pivotaltracker.com/story/show/181352823

### DIFF
--- a/app/js/components/change_member_role.ts
+++ b/app/js/components/change_member_role.ts
@@ -2,39 +2,52 @@ import * as $ from 'jquery';
 import addFlashMessage from './addFlashMessage';
 
 $(() => {
-  const selects = document.querySelectorAll('select[data-url]');
   const flashUl = document.querySelector('.teams__flashMessages') as HTMLUListElement;
+  const changeRoleLinks = document.querySelectorAll('.teams__changeMemberRole');
+  changeRoleLinks.forEach(item => {
+    item.addEventListener('mousedown', (event) => {
+      const target = event.target as HTMLAnchorElement;
+      const roleName = target.getAttribute('data-role');
+      const parentElement = target.closest('div');
+      if (roleName === null) return;
+      if (parentElement === null) return;
 
-  if (!!selects && !!flashUl) {
-    document.addEventListener('change', (event) => {
-      const target = event.target as HTMLSelectElement;
-      if (!target.matches('select[data-url]')) return;
+      if (parentElement.getAttribute('data-url')) {
+        const parent = target.closest('td') as HTMLTableDataCellElement;
+        const name = (parent?.previousElementSibling?.previousElementSibling?.textContent)?.trim() || '';
+        const closeText = flashUl.getAttribute('data-close') || 'Close';
+        const changeRoleText = flashUl.getAttribute('data-roleChange') || `Role changed to ${roleName}`;
+        const url = parentElement!.getAttribute('data-url')?.replace('stand-in', roleName) || '';
 
-      target.classList.add('loading');
-
-      const value = target.value;
-      const name = target?.parentElement?.previousElementSibling?.previousElementSibling?.previousElementSibling?.textContent || '';
-      const closeText = flashUl.getAttribute('data-close') || 'Close';
-      const changeRoleText = flashUl.getAttribute('data-roleChange') || `Role changed to ${value}`;
-      const url = target.getAttribute('data-url')?.replace('stand-in', value) || '';
-
-      fetch(url)
-        .then(response => response.json())
-        .then(data => {
-          if (data !== 'ok') {
-            addFlashMessage(flashUl, data, 'error', closeText);
-            target.classList.remove('loading');
-            return;
-          }
-
-          const message = changeRoleText.replace('<role>', value).replace('<name>', name);
-          addFlashMessage(flashUl, message, 'info', closeText);
-          target.classList.remove('loading');
-        })
-        .catch(error => {
-          addFlashMessage(flashUl, error.message, 'error', closeText);
-          target.classList.remove('loading');
+        fetch(url)
+          .then(response => response.json())
+          .then(data => {
+            if (data !== 'ok') {
+              addFlashMessage(flashUl, data, 'error', closeText);
+              return;
+            }
+            const message = changeRoleText.replace('<role>', roleName).replace('<name>', name);
+            addFlashMessage(flashUl, message, 'info', closeText);
+            parent.querySelectorAll('span').forEach(role => {
+              role.classList.add('hidden');
+            });
+            const newRole = parent.querySelector(`.role__${roleName}`);
+            if (newRole) newRole.classList.remove('hidden');
+          })
+          .catch(error => {
+            addFlashMessage(flashUl, error.message, 'error', closeText);
+          });
+      } else {
+        const parent = target.closest('.roles') as HTMLDivElement;
+        const field = parent.querySelector('input');
+        if (!field) return;
+        field.setAttribute('value', roleName);
+        parent.querySelectorAll('span').forEach(role => {
+          role.classList.add('hidden');
         });
+        const newRole = parent.querySelector(`.role__${roleName}`);
+        if (newRole) newRole.classList.remove('hidden');
+      }
     });
-  }
+  });
 });
--- a/app/js/components/delete_member.ts
+++ b/app/js/components/delete_member.ts
@@ -6,37 +6,39 @@ $(() => {
   const deleteLinks = document.querySelectorAll('.teams__deleteMemberLink');
 
   if (!!flashUl && !!deleteLinks) {
-    document.addEventListener('click', (event) => {
-      const target = event.target as HTMLSelectElement;
-      if (!target.matches('.teams__deleteMemberLink')) return;
-      event.preventDefault();
+    deleteLinks.forEach(item => {
+      item.addEventListener('mousedown', (event) => {
+        event.stopImmediatePropagation();
+        const target = event.target as HTMLSelectElement;
+        if (!target.matches('.teams__deleteMemberLink')) return;
 
-      const td = target.closest('td') as HTMLTableDataCellElement;
-      td.classList.add('loading');
+        const td = target.closest('td') as HTMLTableDataCellElement;
+        td.classList.add('loading');
 
-      const url = target.getAttribute('href') || '';
-      const closeText = flashUl.getAttribute('data-close') || 'Close';
-      const deleteMemberText = flashUl.getAttribute('data-deleteMember') || '';
-      const name = target.getAttribute('data-name') || '';
+        const url = target.getAttribute('href') || '';
+        const closeText = flashUl.getAttribute('data-close') || 'Close';
+        const deleteMemberText = flashUl.getAttribute('data-deleteMember') || '';
+        const name = target.getAttribute('data-name') || '';
 
-      fetch(url)
-        .then(response => response.json())
-        .then(data => {
-          if (data !== 'success') {
-            addFlashMessage(flashUl, data, 'error', closeText);
+        fetch(url)
+          .then(response => response.json())
+          .then(data => {
+            if (data !== 'success') {
+              addFlashMessage(flashUl, data, 'error', closeText);
+              td.classList.remove('loading');
+              return;
+            }
+
+            const message = deleteMemberText.replace('<name>', name);
+            addFlashMessage(flashUl, message, 'info', closeText);
             td.classList.remove('loading');
-            return;
-          }
-
-          const message = deleteMemberText.replace('<name>', name);
-          addFlashMessage(flashUl, message, 'info', closeText);
-          td.classList.remove('loading');
-          td.parentElement?.remove();
-        })
-        .catch(error => {
-          addFlashMessage(flashUl, error.message, 'error', closeText);
-          td.classList.remove('loading');
-        });
+            td.parentElement?.remove();
+          })
+          .catch(error => {
+            addFlashMessage(flashUl, error.message, 'error', closeText);
+            td.classList.remove('loading');
+          });
+      });
     });
   }
 });
--- a/app/scss/components/actions.scss
+++ b/app/scss/components/actions.scss
@@ -93,9 +93,5 @@ div.actions, td.actions > div[tabindex="0"], div.roles, td.roles > div[tabindex=
   &:hover ul, ul:hover, &:focus ul, &:focus-within ul {
     display: block;
   }
-
-  & ul, ul:active {
-    display: none;
-  }
 }
 /* stylelint-enable */
--- a/app/scss/components/actions.scss
+++ b/app/scss/components/actions.scss
@@ -15,11 +15,11 @@ div.add-entity-actions {
   }
 }
 
-td.actions {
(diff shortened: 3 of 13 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/inviteModal.html.twig
```diff
@@ -7,16 +7,13 @@
                     <label for="email">{{ "teams.management.invite.modal.email.label"|trans }}</label>
                     <input type="text" name="email" id="email" />
                 </div>
-                <div class="form__row">
+                <div class="roles">
+                    <input type="hidden" name="role" value="member">
```
MKodde, line 11: Is it okay to set a default value here? This is something to discuss with the PM's. The entire modal is very UX unfriendly. There is no form validation and there is no flash message after the modal window performed his action.

MKodde (commented):
Disclaimer, I'm no frontend expert, so I limited my review to things that stood out to me. Please ask one of our JS developers for an additional review. @Badlapje seems an exellent candidate for this.

I must say I really love the end result of this PR. The select lists have become more homogeneous between the different components/views. And that's a win in my opinion.

Other non code review related remarks
1. I see you tried to bundle the work you did into separate commits, and I commend you for that! 
2. Some of the commit titles are not as good as they could have been. Consider improving this, or squash this PR when we are finished with it. Some examples are the `...`, and the numerous `Fix ` messages.

app/js/components/change_member_role.ts
```diff
@@ -2,39 +2,48 @@ import * as $ from 'jquery';
...
-  const selects = document.querySelectorAll('select[data-url]');
   const flashUl = document.querySelector('.teams__flashMessages') as HTMLUListElement;
+  const changeRoleLinks = document.querySelectorAll('.teams__changeMemberRole');
+  changeRoleLinks.forEach(item => {
+    item.addEventListener('mousedown', (event) => {
+      const target = event.target as any;
+      const roleName = target.getAttribute('role');
+      if (target.closest('div').getAttribute('data-url')) {
+        const parent = target.closest('td') as HTMLTableDataCellElement;
+        const name = (parent?.previousElementSibling?.previousElementSibling?.textContent || '').trim();
```
Badlapje, line 13: I'd rewrite this slightly for better legibility:

`const name = (parent?.previousElementSibling?.previousElementSibling?.textContent).trim()  || ''`  That makes it clearer that you either use the textcontent of that element when trimmed, or an empty string (which doesn't need to be trimmed). 

Aside from that, using the `.closest` method with a CSS selector would be better than traversing via the `previousElementSibling` as that's very closely tied to the HTML implementation.

app/scss/components/actions.scss
```diff
@@ -90,8 +90,12 @@ div.actions, td.actions > div[tabindex="0"] {
     }
   }
 
-  &:hover ul, ul:hover, &:focus ul, &:focus-within ul {
+  &:hover ul, ul:hover {
```
Badlapje, line 93: Where did the focus styles go?

Badlapje: Or in other words: did you try using this with a keyboard?

app/js/components/delete_member.ts
```diff
@@ -6,37 +6,39 @@ $(() => {
   const deleteLinks = document.querySelectorAll('.teams__deleteMemberLink');
 
   if (!!flashUl && !!deleteLinks) {
-    document.addEventListener('click', (event) => {
-      const target = event.target as HTMLSelectElement;
-      if (!target.matches('.teams__deleteMemberLink')) return;
-      event.preventDefault();
+    deleteLinks.forEach(item => {
+      item.addEventListener('mousedown', (event) => {
+        event.stopImmediatePropagation();
```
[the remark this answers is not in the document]
VadimSchmitz: Due to the event getting captured with preventDefault it was still directly tried to go to the endpoint and not return to the actual page. this means that when the endpoint was triggered you would get redirected to an empty page with ok on it.
with the stopImmediatePropagation event bubbling is done instead so it truly stops all actions after the initial click.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/actions.html.twig
```diff
@@ -1,5 +1,6 @@
 <td class="actions teams__actions">
-    <div tabindex="0">
+    <div tabindex="0" class="">
```
Badlapje, line 2: empty class attribute here, that's likely something which can be removed

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/role.html.twig
```diff
@@ -4,39 +4,57 @@
...
+            </span>
+            <span class="pull-left role__member {% if userRole != 'member' %}hidden{% endif %}">
+                <i class="fa fa-user fa-fw" aria-hidden="true"></i>
+                {{ 'teams.management.table.role.member'|trans }}
+            </span>
+    <i class="fa fa-spinner fa-spin loading"></i>
+    <i class="fa fa-caret-down" aria-hidden="true"></i>
+    <ul>
+        <li class="teams__actionItem">
+            <a class="teams__changeMemberRole" role="admin">
```
Badlapje, line 34: This is an a-tag without a href attribute.  That should never happen.  If this is a link, then it needs a href attribute.  If it isn't then it should likely be a button (not sure what the item should do when clicked, but if it's just to change member role without changing page, then a button is best.  This should likely be a button?

Also: the role attribute is an existing attribute, So you shouldn't use that, it should be a data-role attribute.  See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles for more information.

Badlapje (changes requested):
Nice work already!  I found a couple of items that i reckon need to be resolved though.

Aside from those made within the code one more nitpick and one serious comment:
- your description for the PR reads "Rework the drop downs to no longer use
elements to allow styling and icons".  You're still using elements however.  So i assume you meant to say "Rework the drop downs to no longer use selects, in order to be able to style them better & use icons"?
- did you test the dropdown for accessibility?  I mean using keyboard and screenreader?  I see no accessibility-specific code, so i doubt it.  Please see https://w3c.github.io/aria/#select for a basic intro in what's needed to duplicate native select functionality.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/teamTable.html.twig
```diff
@@ -11,7 +11,11 @@
             {% include "@Dashboard/Teams/Table/name.html.twig" %}
             {% include "@Dashboard/Teams/Table/email.html.twig" %}
             {% include "@Dashboard/Teams/Table/status.html.twig" %}
+                {% if not isInvite %}
             {% include "@Dashboard/Teams/Table/roleCell.html.twig" %}
+                {% else %}
+                        <td></td>
```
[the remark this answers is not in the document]
VadimSchmitz: As discussed with @Badlapje i have decided to take the previous implementations approach, which is having a greyed out version of the dropdown.

app/scss/components/drop_down_actions.scss
```diff
@@ -45,3 +45,8 @@ section.drop-down-actions {
     }
   }
 }
+
+.actions div, .roles div {
+  width: 100%;
+  margin-bottom: 0px !important;
```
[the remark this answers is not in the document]
VadimSchmitz: It has been changed to use a class now.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/actions.html.twig
VadimSchmitz: Empty class has been removed

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/role.html.twig
VadimSchmitz: an href="#" has been included
role has been renamed to data-role

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/actions.html.twig
VadimSchmitz: has been removed

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/role.html.twig
Badlapje: I know adding a `href="#"` is a ubiquitous practice, however it's invalid HTML and therefore a bad practice.

If you turn it into a button, then you can use `initial: all` to get rid of all browser styling for the button 😉

Badlapje: see: https://css-tricks.com/almanac/properties/a/all/

app/js/components/change_member_role.ts
```diff
@@ -2,39 +2,48 @@ import * as $ from 'jquery';
 import addFlashMessage from './addFlashMessage';
 
 $(() => {
-  const selects = document.querySelectorAll('select[data-url]');
   const flashUl = document.querySelector('.teams__flashMessages') as HTMLUListElement;
+  const changeRoleLinks = document.querySelectorAll('.teams__changeMemberRole');
+  changeRoleLinks.forEach(item => {
+    item.addEventListener('mousedown', (event) => {
+      const target = event.target as any;
```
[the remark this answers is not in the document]
VadimSchmitz: any has been changed to HTMLAnchorElement

app/js/components/change_member_role.ts
VadimSchmitz: it has been rewritten to your suggestion

app/scss/components/actions.scss
VadimSchmitz: Removing them doesnt seem to make a difference?
They have been readded

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/role.html.twig
VadimSchmitz: I could not get the buttons styled easily, 
The other dropdowns use <a> tags too with an href. Restyling the <button> elements will take too much time for now. 

After asking  @quartje we have agreed to add the remainder  of the work as a technical debt ticket
- Keyboard accesibility
- Use buttons instead of a tags within the dropdown template.
see:
-https://www.pivotaltracker.com/story/show/182261042
- https://www.pivotaltracker.com/story/show/182261075

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/inviteModal.html.twig
MKodde: I feel the same way. But maybe we should run this by the PO? Either way I'm fine with this, we can fix it later if PO wants different behavior.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/teamTable.html.twig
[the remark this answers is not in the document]
MKodde: Alright, that fixes my problem!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/actions.html.twig
```diff
@@ -1,5 +1,6 @@
...
+        <span class="pull-left">Select an action</span>
```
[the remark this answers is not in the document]
MKodde: The final step is to add the translation to the translations file. That's the place where the placeholder is translated with a default text. Which can be overridden using the web-based translator.

MKodde (changes requested):
Getting there, I resolved most of my previous feedback and added some new. Have not re tested this in my browser yet. Will do that after you addressed my new points.

app/js/components/change_member_role.ts
```diff
@@ -37,7 +37,7 @@ $(() => {
         const parent = target.closest('.roles') as HTMLDivElement;
         const field = parent.querySelector('input');
         if (!field) return;
-        field.setAttribute('value', roleName);
+        field.setAttribute('value', roleName!);
```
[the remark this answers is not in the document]
VadimSchmitz: Thanks for the discussion we had about this in person. I agree with you and have done some null checks to be 100% sure that the elements returned are not null.

app/scss/components/actions.scss
Badlapje: It should add the same styles when tabbing around the interface to the ul as when hovering over it.  Only works if the ul is focusable or if the ul has elements within it that are focusable.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Teams/Table/role.html.twig
```diff
@@ -4,39 +4,57 @@
...
-        {{ 'teams.management.table.role.manager'|trans }}
-        {{ iconAfter }}
-    </option>
-    <option class="fontawesome" data-icon="fa-user" value="member" {% if userRole == 'member' %}selected="selected"{% endif %}>
-        &#xf007;{{ whitespaceBefore }}
-        {{ 'teams.management.table.role.member'|trans }}
-        &nbsp;{{ iconAfter }}
-    </option>
-</select>
+<div class="no-margin-bottom" tabindex="0" {% if url is defined %} data-url="{{ url }}"{% endif %}>
```
Badlapje, line 13: From the looks of it, this should likely be a UL with LI's in it, rather than a div with span's in it?

Badlapje: the sub-ul would then need to be in a li

Badlapje (approved):
I've found some smaller items, up to you if they are important enough or not.

Redoing the markup from div/span to ul/li would be benefitial, but as you already have a technical debt ticket for the accessibility features, that one might go in there as well so i leave it up to you.

PR #504, opened 2022-05-23
Docker: Fix xdebug

quartje:
Replacing the config file for xdebug makes xdebug no longer work. After this change, a docker-compose build should build a new fpm image with a working xdebug


VadimSchmitz (approved):
Glad Xdebug is working again, Good work

PR #506, opened 2022-05-23
Remove motivation placeholder from view entity page

VadimSchmitz:
- Remove motivation placeholder from view entity page
- Add webtest for detail attributes

https://www.pivotaltracker.com/story/show/178674074


MKodde (approved):
Nothing but net!

PR #513, opened 2022-08-30
Retrieve only Unresolved issues from Jira

MKodde:
During a peer programming session with Bart, we encountered that the request delete of prod entity status was not reverted back when the issue status in Jira was set to 'resolved'. Resulting in the ticket retaining its faulty status. This feature was encountered when users unintentionally clicked the 'remove prod entity' button. And the support team verified the action with the user, resulting in the resolution of the ticket (without any action). This should have resulted in the entity status reverting back to 'published'.

By simply filtering the Jira results to only show the Unresolved items we resolved this issue. With the additional benefit of only retrieving the active tickets at that point. Decreasing the processing time and resource hogging.

See: https://www.pivotaltracker.com/story/show/176244612


MKodde:
@quartje is it OK to hardcode this condition. My rationale was that it is, on account of it being a non-custom status identifier. Having control over the resolution status would actually result in unexpected query results.

quartje:
Nice optimisation! I've tested it, and it all seems to be working fine. Ready to merge I'd say

quartje (approved):

PR #514, opened 2022-08-30
Refrain from adding epTID for OIDCNG

MKodde:
See: https://www.pivotaltracker.com/story/show/181121296


johanib (approved):
Thanks for the explanation.

Seems like a straightforward solution.

This is the only location this check is attribute is used? Are there any other places where it can be cleaned up?

MKodde:
This was the only place the check was still in place. At least if my grep-fu is not letting me down :)

PR #515, opened 2022-08-31
Install JS and PHP 3rd party dependencies

MKodde:
Especially the JS dependencies have been upgraded more severely as some packages where ridden with security issues. 

Resulting in a stylelint upgrade that entailed some reconfiguration of the linting tool. And inadvertently resulting in having to repair some CSS code that did not adhere to the newer rulesets. I tried to keep the number of changes to a minimum. Having the defautl rules for scss files, and running the stylelint auto-fixer resulted in a host of problems. Where the actual styling of SP Dashboard was compromised.

Finally, the upgrade JS dependencies uppded our Node JS runtime dependency to a minimum of node 14. That was an issue in the GHA build workflow, where a docker container is composed. This problem is resolved by adding a stage that installs the yarn dependencies and performs the yarn encore action.


.stylelintrc
```diff
@@ -1,8 +1,11 @@
 {
   "extends": "stylelint-config-recommended",
```
johanib, line 2:
```suggestion
  "extends": "stylelint-config-standard-scss",
```

Ik ben geen frontend expert, maar er is een standaard voor scss, die zou je misschien kunnen gebruiken, gezien ik in dit PR zie dat hier ook scss gebruikt wordt. De package wordt zelfs al toegevoegd aan de package.json

package.json
```diff
@@ -4,27 +4,28 @@
...
     "jest-canvas-mock": "^1.1.0",
-    "jscpd": "^0.6.23",
+    "jscpd": "^3.4",
     "sass": "^1.43.4",
-    "sass-loader": "^7.1.0",
-    "stylelint": "^9.5.0",
-    "stylelint-config-recommended": "^2.1.0",
+    "sass-loader": "^13.0",
+    "stylelint": "^14.11",
+    "stylelint-config-standard-scss": "^5.0.0",
```
johanib, line 20: Is this stil needed if the standard is used?

package.json
```diff
@@ -4,27 +4,28 @@
...
     "file-loader": "^2.0.0",
     "import-glob-loader": "^1.1.0",
     "jest": "^26.6.3",
     "jest-canvas-mock": "^1.1.0",
-    "jscpd": "^0.6.23",
+    "jscpd": "^3.4",
     "sass": "^1.43.4",
-    "sass-loader": "^7.1.0",
-    "stylelint": "^9.5.0",
-    "stylelint-config-recommended": "^2.1.0",
```
johanib, line 20: Maybe this should not be removed if it is used in the .stylelintrc

johanib:
Tip, use `yarn outdated` to see potential upgrade candidate packages

johanib (changes requested):
Probably need to double check the required packages.

There are no functional changes, so if the build is green, it should be alright.

Not all packages are up to date, but that might not be the scope of this ticket :)

MKodde:
Awesome feedback! I was pleasantly surprised by the usefulness of `yarn outdated`

This is what now remains to be upgraded in the future. These packages upgrades needed more attention than is warranted in this upgrade PR:

```
$ yarn outdated
yarn outdated v1.22.17
info Color legend : 
 "<red>"    : Major Update backward-incompatible updates 
 "<yellow>" : Minor Update backward-compatible features 
 "<green>"  : Patch Update backward-compatible bug fixes
Package                   Current Wanted Latest     Package Type    URL                                                                      
@types/jest               28.1.8  28.1.8 29.0.0     devDependencies https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/jest
chart.js                  2.9.4   2.9.4  3.9.1      dependencies    https://www.chartjs.org                                                  
chartjs-plugin-datalabels 0.7.0   0.7.0  2.1.0      dependencies    https://chartjs-plugin-datalabels.netlify.app                            
jest                      28.1.3  28.1.3 29.0.2     devDependencies https://jestjs.io/                                                       
select2                   4.0.13  4.0.13 4.1.0-rc.0 dependencies    https://select2.org                                                      
tippy.js                  2.6.0   2.6.0  6.3.7      dependencies    https://atomiks.github.io/tippyjs/                                       
Done in 1.69s.
```

Next up: I'll be reviewing the SCSS linting feedback you provided.

.stylelintrc
MKodde: I'm no expert either, but did get my hands dirty in upgrading stylelint. These are my findings. By the way, thanks for the (private) link to your current project that rhymes with 'L'Oréal'. That put some things into perspective for me.

The SCSS stylelint implementation at that project differs slightly from this one. The main reason why we do not use the `"extends": "stylelint-config-standard-scss",` option is that the SCSS is generally NOT compliant with that standard. We adhere to the CSS standards :man_shrugging: 

1. I tried fixing the stylesheets to adhere to the SCSS rules. But that ended in a sea of tears as many of the fixed constructs ended up actually breaking the CSS outputted by SASS.
2. The solution implemented is a sort of concession. It does evaluate the code as if being SCSS due to the `"customSyntax": "postcss-scss",` directive. This allows for the nesting and the mixin stuff we do using SCSS. But it lints according to the regular CSS styling rules.

johanib (approved):
lgtm after rework :+1:

PR #517, opened 2022-09-08
Show flash message when privacy messages are changed

Shaky212:
see: https://www.pivotaltracker.com/story/show/182246029

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Privacy/form.html.twig
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Privacy/form.html.twig
@@ -9,7 +9,13 @@
 
         {% for message in app.flashes('error') %}
             <div class="message error">
-                {{ message }}
+                {{ message|trans }}
+            </div>
+        {% endfor %}
+
+        {% for message in app.flashes('notice') %}
+            <div class="flashMessage info">
+                {{ message|trans }}
             </div>
         {% endfor %}
 
--- a/tests/webtests/EditPrivacyQuestionsTest.php
+++ b/tests/webtests/EditPrivacyQuestionsTest.php
@@ -104,5 +104,10 @@ class EditPrivacyQuestionsTest extends WebTestCase
             '2018-12-31',
             $crawler->filter('#dashboard_bundle_privacy_questions_type_certificationValidTo')->attr('value')
         );
+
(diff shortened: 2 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/Privacy/form.html.twig
```diff
@@ -9,7 +9,13 @@
 
         {% for message in app.flashes('error') %}
             <div class="message error">
-                {{ message }}
+                {{ message|trans }}
+            </div>
+        {% endfor %}
+
+        {% for message in app.flashes('notice') %}
+            <div class="flashMessage info">
```
johanib, line 17: <div class="message error">
vs 
<div class="flashMessage info">

Does the styling look OK?
I would expect it to be: 
<div class="message info">, because the message contains the styling in the error flash?

MKodde: I think Johan's suggestion is very good. And it makes the class declarations in sync with the Symfony standard.

However ;) in our previous sprints Koen created components that are user friendly from an a11y standpoint. Those flash messages are styled slightly different. See the flash message stylesheet for some pointers on how to render them. Also consider implementing the close button. https://github.com/SURFnet/sp-dashboard/blob/066a1b5a6b77927233643a1d219cee5ff088257b/app/scss/components/flash_message.scss

tests/webtests/EditPrivacyQuestionsTest.php
```diff
@@ -104,5 +104,10 @@ public function test_it_can_submit_the_form()
             '2018-12-31',
             $crawler->filter('#dashboard_bundle_privacy_questions_type_certificationValidTo')->attr('value')
         );
+
+        $this->assertContains(
+            'Your changes were saved!',
+            $crawler->filter('div.message.info')->text()
+        );
```
MKodde, line 111: :muscle:

MKodde (approved):

MKodde:
A very nice first contribution! See my 2¢ on the suggestion Johan gave in review. Should you consider following my pointer. You can easily fix this by removing your last commits from the commit history

```
git reset HEAD^^   # Removes the last 2 commits from your local commit history
git push --force      # Force push is required, as upstream is changed in a destructive manner. 
```

PR #520, opened 2022-09-16
Visual feedback after creating, editing and deleting a service or entity

Shaky212:
see: https://www.pivotaltracker.com/story/show/182246029

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
@@ -16,7 +16,7 @@ entity.delete.request.ticket.description: "h2. Details
 *Applicant name*: %applicant_name%
 *Applicant email*: %applicant_email%.
 *Entity name (en)*: %entity_name%."
-entity.delete.flash.message: "Your entity is deleted."
+entity.delete.flash.success: "Your entity is deleted."
 unknownDisplayName: "Unknown"
 
 entity:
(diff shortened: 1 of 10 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -16,7 +16,7 @@ entity.delete.request.ticket.description: "h2. Details
 *Applicant name*: %applicant_name%
 *Applicant email*: %applicant_email%.
 *Entity name (en)*: %entity_name%."
-
+entity.delete.flash.message: "Your entity is deleted."
```
MKodde, line 19: Perhaps, for the sake of consistency: rename the translation key to read: `entity.delete.flash.success`.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/FlashMessage/flashMessages.html.twig
```diff
@@ -0,0 +1,14 @@
+{% set flashes = app.session.flashbag.all %}
```
MKodde, line 1: Thanks for extracting the flash message into a partial of its own! I see you removed the `card messages` class from the container `div` element. Was that intentional? As they seem to have changed the look and feel of the messages. Either use the old look and feel. Or clean up the now unused classes from the SCSS stylesheet.

MKodde (commented):
Great work once again! See some feedback below. 

Additionally, I wonder if you tried to add some web tests coverage on the new flash messages you added. If you did not, maybe it's easy to do so.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
Shaky212: Agree.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/FlashMessage/flashMessages.html.twig
Shaky212: On closer inspection, the card messages aren't that bad after all and line up nicely with the rest of the page. So revert to the old look and feel.

PR #523, opened 2022-09-21
Create a Jira ticket after creating an entity change request

MKodde:
A Jira ticket with a distinctive issue type is created. The actual issue type is yet to be determined, but can be set in the parameters.yml during deployment.

The title and description of the Ticket have been made translatable, a default was added to the default translations file.

See: https://www.pivotaltracker.com/story/show/182411800
See: https://www.pivotaltracker.com/story/show/179749372

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/MailService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/MailService.php
@@ -1,0 +1,45 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * Copyright 2022 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\Service;
+
+use Exception;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\ManageEntity;
+use Surfnet\ServiceProviderDashboard\Domain\Mailer\Mailer;
+use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Factory\MailMessageFactory;
+
+class MailService
+{
+    /** @var MailMessageFactory */
+    private $mailFactory;
+    /** @var Mailer */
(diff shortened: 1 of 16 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/MailService.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function __construct(MailMessageFactory $mailFactory, Mailer $mailer)
+    {
+        $this->mailer = $mailer;
+        $this->mailFactory = $mailFactory;
+    }
+
+    public function sendErrorReport(ManageEntity $entity, Exception $exception)
+    {
+        // Inform the service desk of the unavailability of Jira
```
Shaky212, line 41: Is the unavailability the only case or could there be several reasons? The comment seems unnecessary

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
```diff
@@ -56,6 +68,16 @@ public function findByManageIds(array $manageIds)
...
     }
 
+    public function createJiraTicket(
+        ManageEntity $entity,
+        PublishProductionCommandInterface $command,
+        string $issueType,
+        string $summaryTranslationKey,
+        string $descriptionTranslationKey
+    ): Issue {
+        return new Issue('KEY-27', 'fake-type', Issue::STATUS_OPEN);
```
Shaky212, line 78: Just for my understanding. What does KEY-27 mean? Perhaps, use a more explaining constant?

Shaky212 (commented):
Just a few remarks from my side. For me it is still difficult to see substantive what is really happening.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
MKodde: Jira makes a human readable ID for each ticket. SURF-9 for example. I just created a simple look alike here. Given this is a very explicit dev-only test class. I do not see the need to extract this into a fancy constant.

src/Surfnet/ServiceProviderDashboard/Application/Service/MailService.php
MKodde: At this point, we only send an error report when the Jira API was down or otherwise unavailable. I extracted this code from the CommandHandler, so you might be right in your observation that the comment has become obsolete. 

When I'm back on my branch, I'll be sure to check that class with this in mind.

src/Surfnet/ServiceProviderDashboard/Application/Service/MailService.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
johanib, line 2: strict?

johanib (approved):
As discussed, the changes look good :+1:

PR #524, opened 2022-09-22
Docker: Copy yarn files from the build stage

quartje:
This copies the files created in the node build stage to the final Docker images


MKodde (approved):
Excellent!

PR #525, opened 2022-09-22
Provide a change request overview

Shaky212:
Withing sp dashboard we want to make the  outstanding changes be transparent for a supplier. Therefor a new screen "Requested changes" can be selected from the context menu at an published entity. Each outstanding change contains the list of changes together with the date of creation and a possible note.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
@@ -19,6 +19,9 @@
 namespace Surfnet\ServiceProviderDashboard\Application\Dto;
 
 use DateTime;
+use Exception;
+use Surfnet\ServiceProviderDashboard\Application\Exception\InvalidDateTimeException;
+use Webmozart\Assert\Assert;
 
 class ChangeRequestDto
 {
@@ -39,7 +42,7 @@ class ChangeRequestDto
 
     private $pathUpdates = [];
 
-    private function __construct(string $id, ?string $note, DateTime $created, array $pathUpdates)
+    private function __construct(string $id, string $note, DateTime $created, array $pathUpdates)
     {
         $this->id = $id;
         $this->note = $note;
@@ -47,12 +50,26 @@ class ChangeRequestDto
         $this->pathUpdates = $pathUpdates;
     }
 
+    /**
+     * @throws InvalidDateTimeException
+     */
     public static function fromChangeRequest(array $changeRequest): ?ChangeRequestDto
     {
-        $created = new DateTime($changeRequest['created']);
+        Assert::isArray($changeRequest);
+        Assert::keyExists($changeRequest, 'id', 'No id specified');
+        Assert::keyExists($changeRequest, 'created', 'No create datetime specified');
+        Assert::keyExists($changeRequest, 'pathUpdates', 'No pathUpdates specified');
+        Assert::isNonEmptyMap($changeRequest['pathUpdates'], 'No pathUpdates available');
+
+        try {
+            $created = new DateTime($changeRequest['created']);
+        } catch (Exception $e) {
+            throw new InvalidDateTimeException();
+        }
+        $note = $changeRequest['note'] ?? '';
 
         return new self($changeRequest['id'],
-            $changeRequest['note'],
+            $note,
             $created,
             $changeRequest['pathUpdates']);
     }
--- a/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
@@ -22,6 +22,7 @@ use DateTime;
 use Exception;
 use Surfnet\ServiceProviderDashboard\Application\Exception\InvalidDateTimeException;
 use Webmozart\Assert\Assert;
+use function array_key_exists;
 
 class ChangeRequestDto
 {
@@ -68,10 +69,26 @@ class ChangeRequestDto
         }
         $note = $changeRequest['note'] ?? '';
 
+        self::flattenArp($changeRequest);
+
         return new self($changeRequest['id'],
             $note,
             $created,
             $changeRequest['pathUpdates']);
+    }
+
+    private static function flattenArp(&$changeRequest)
+    {
+        if (array_key_exists('arp', $changeRequest['pathUpdates'])) {
+            $arp = $changeRequest['pathUpdates']['arp'];
+            if (array_key_exists('attributes', $arp) && !empty($arp['attributes'])) {
+                foreach ($arp['attributes'] as $urn => $attribute) {
+                    // Include ARP entries in the pathupdates, not in pathUpdates/arp
+                    $changeRequest['pathUpdates'][$urn] = $attribute[0]['motivation'];
+                }
+            }
+            unset($changeRequest['pathUpdates']['arp']);
+        }
     }
 
     public function getId(): string
--- a/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
@@ -29,11 +29,6 @@ class ChangeRequestDto
     /**
      * @var string
      */
-    private $id;
-
-    /**
-     * @var string
-     */
     private $note;
 
     /**
@@ -43,9 +38,8 @@ class ChangeRequestDto
 
     private $pathUpdates = [];
 
-    private function __construct(string $id, string $note, DateTime $created, array $pathUpdates)
+    private function __construct(string $note, DateTime $created, array $pathUpdates)
     {
-        $this->id = $id;
         $this->note = $note;
         $this->created = $created;
         $this->pathUpdates = $pathUpdates;
@@ -57,7 +51,6 @@ class ChangeRequestDto
     public static function fromChangeRequest(array $changeRequest): ?ChangeRequestDto
     {
         Assert::isArray($changeRequest);
-        Assert::keyExists($changeRequest, 'id', 'No id specified');
         Assert::keyExists($changeRequest, 'created', 'No create datetime specified');
         Assert::keyExists($changeRequest, 'pathUpdates', 'No pathUpdates specified');
         Assert::isNonEmptyMap($changeRequest['pathUpdates'], 'No pathUpdates available');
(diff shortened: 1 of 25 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -176,15 +176,6 @@ private function generateDataForExistingEntity(
...
     }
 
-    private function generateForChangeRequest(ManageEntity $entity, EntityDiff $differences)
-    {
-        $metadata = $differences->getDiff();
-        // Arp is to be sent in its entirety as it does not support the MERGE WRITE feature
-//        $metadata['arp'] = $this->arpMetadataGenerator->build($entity);
-
-        return $metadata;
-    }
```
MKodde: Thanks for cleaning up my merge conflict :bow:

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
```diff
@@ -167,6 +167,15 @@ public function allowSecretResetAction()
         return $meetsProtocolRequirement && $meetsPublicationStatusRequirement;
     }
 
+    public function allowChangeRequestAction(): bool
+    {
+        if ($this->readOnly) {
+            return false;
+        }
+        return $this->status == Constants::STATE_PUBLISHED;
+    }
```
MKodde, line 176: I believe we have a thorough test coverage of the neighboring methods. Please cover this one in test too.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityChangeRequestController.php
```diff
@@ -0,0 +1,51 @@
...
+     * @Route("/entity/change-request/{environment}/{manageId}/{serviceId}", name="entity_published_change_request")
+     */
+    public function changeRequestAction(
+        Request $request,
+        ChangeRequestService $service,
+        int $serviceId,
+        string $manageId,
+        string $environment
+    ): Response {
+
```
MKodde, line 43: I'm surely going to test this, but I have a strong feeling we need additional authorization checking here. For example, verify that nobody outside of the currently active service is allowed to view this data.

tests/unit/Domain/ValueObject/ChangeRequestDtoTest.php
```diff
@@ -0,0 +1,79 @@
...
+        $changeRequest = ChangeRequestDto::fromChangeRequest($changes);
+
+        $this->assertEquals(1, $changeRequest->getId());
+        $this->assertEquals('a cracked note', $changeRequest->getNote());
+        $this->assertEquals('2022-09-21 15:00:00', $changeRequest->getCreated()->format('Y-m-d H:i:s'));
+        $this->assertIsArray($changeRequest->getPathUpdates());
+        $this->assertEquals(2, count($changeRequest->getPathUpdates()));
+    }
+
+    public function test_it_throw_exception_on_invalid_date_time()
```
MKodde, line 48: Now I understand why the method had the `@throws /Exception` annotation.

In the domain layer we do not like to be dependent on constructs from outside our domain. I would expect a domain specific exception at this point. So in the input validation of your DTO you should validate if the data time makes sense. In other words place a try catch around the creation of the DateTime. And throw a custom exception when shit hits the fan.

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityActions.php
Shaky212: That is correct. This one is covered in test_change_request_action_is_allowed

src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
```diff
@@ -0,0 +1,79 @@
...
+    private $note;
+
+    /**
+     * @var datetime
+     */
+    private $created;
+
+    private $pathUpdates = [];
+
+    private function __construct(string $id, ?string $note, DateTime $created, array $pathUpdates)
```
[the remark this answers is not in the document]
Shaky212: Agree, i will change that.

src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDtoCollection.php
```diff
@@ -0,0 +1,45 @@
...
+     * @var ChangeRequestDto[]
+     */
+    private $changeRequests = [];
+
+    public function __construct(array $changeRequests)
+    {
+        foreach ($changeRequests as $id => $changeRequest) {
+            $this->changeRequests[$id] = ChangeRequestDto::fromChangeRequest($changeRequest);
+        }
+        array_multisort($this->changeRequests, SORT_ASC, SORT_REGULAR);
```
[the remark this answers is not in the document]
Shaky212: Ah, I don't know what I was thinking there. But refactored it with a usort and a ChangeRequestDtoComparer class which can be used to add compare methods.

src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDtoCollection.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function getChangeRequests(): array
+    {
+        return $this->changeRequests;
+    }
+
+    public function count(): int
+    {
+        return count($this->changeRequests);
+    }
```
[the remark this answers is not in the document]
Shaky212: I was struggling with the collection in twig and tried something with the count. I will remove the count together with the test.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/changeRequest.html.twig
```diff
@@ -0,0 +1,34 @@
...
+    {% else %}
+        {{ 'entity.change_request.none.text.html'|trans|wysiwyg }}
+    {% endif %}
+
+    {%  for changeRequest in changeRequests.getChangeRequests() %}
+
+        <h2>{{ changeRequest.created|date }}</h2>
+
+        {% if changeRequest.note is not null %}
+            <h2>{{ changeRequest.note }}</h2>
```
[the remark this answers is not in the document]
Shaky212: Agree

MKodde:
## Functional test observations/findings

### 1. ARP
When the ARP is changed, the path updates are not created correctly. The arp
is an array, and the ChangeRequestDto assumes every entry of the pathUpdates
are of a non array type. This causes a 500 error in the Twig template where 
you loop over the path updates.

Steps to reproduce: change some ARP values and reload the overview page that you
changed the ARP for.

### 2. Visual observation
Figure 1 shows a Change request overview page with a host of changes. As you can
see, the overview page is not very easy on the eyes. The table data width should
be set uniformly IMHO. Improving the user experience. Also consider adding a
background color to the table head row. 

Another good thing to do: compare the styling on the entity detail page with that of the 
new change requests overview page. And sync some of the styling. All to make SPD more homogenous.

Screenshot from 2022-10-05 09-59-07

### 3. Opening change requests for an entity without any
When an entity has no change requests. And you open them for that entity, I would
expect any of the following:

1. The feature shoud not be in the entity actions to begin with
2. If that proves to difficult, then opening the page should not result in a 500 error. But maybe just show a 'no results found' message.

Be sure to ask which of these options Bart & Teun prefer

### 4. Stuck
When on the overview page, there is no means to return to the previous page. 
To be consistent in this regard. I suggest you implement the same toolbar we also
have on the entity details page. Maybe onmitting the edit and delete features here.

MKodde:
Regarding the timezone issue in your test:

1. We can now reproduce this 'issue' in both travis and our dev vm's as the correct timezone is now set in the php fpm container.
2. The test still fails as the date-time that is asserted is off by two hours. Manage stores times in UTC. We show them in our local TZ. So you should add +2h to your asserted time.

src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDto.php
```diff
@@ -0,0 +1,96 @@
...
+
+        return new self($changeRequest['id'],
+            $note,
+            $created,
+            $changeRequest['pathUpdates']);
+    }
+
+    public function getId(): string
+    {
+        return $this->id;
```
MKodde, line 79: The ID seems unused throughout the application. If we don't need it. Remove it from the DTO please!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityChangeRequestController.php
```diff
@@ -0,0 +1,63 @@
...
+class EntityChangeRequestController extends Controller
+{
+    use EntityControllerTrait;
+
+    /**
+     * @Method({"GET", "POST"})
+     * @Route("/entity/change-request/{environment}/{manageId}/{serviceId}", name="entity_published_change_request")
+     */
+    public function changeRequestAction(
+        Request $request,
```
MKodde, line 39: Request $request seems unused. If not used, please remove it from the parameters.

src/Surfnet/ServiceProviderDashboard/Application/Dto/ChangeRequestDtoCollection.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * @var ChangeRequestDto[]
+     */
+    private $changeRequests = [];
+
+    public function __construct(array $changeRequests)
+    {
+        Assert::isArray($changeRequests);
+        Assert::isNonEmptyList($changeRequests, 'No change requests available');
```
MKodde, line 34: An empty collection should be allowed I guess. I ran into this Exception during my functional review. See feedback below for more info.

MKodde (changes requested):
Getting there! See some additional feedback below

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/changeRequest.html.twig
```diff
@@ -0,0 +1,42 @@
...
+        <div class="fieldset card">
+            <h2>{{ changeRequest.created|date }}</h2>
+
+            {% if changeRequest.note is not null %}
+                <p>{{ changeRequest.note }}</p>
+            {% endif %}
+
+            <table>
+                <tr>
+                    <th>{{ 'entity.change_request.value'|trans }}</th>
```
MKodde, line 29: Setting the width on the first column makes more sense to me. At this point the attribute name is much wider than the possibly longer value it can have.

Also your attempt at naming the `th` class is quite good. But while the rest of the block is not named accordingly, maybe it is better to just name it in a more conventional manner. For example set the classname on the table. And target the first th from the stylesheet.

```
table.change-request-overview th:first-child {
    width: 24em;
    overflow: hidden;
}
```

MKodde (approved):

PR #526, opened 2022-09-26
Ignore colons and thus a port in a clientId

Shaky212:
A validation is applied that it is no longer possible to include a ':' in a clientId (type oidc, resource server or client credential client). This happens e.g. if a port number is provided 

see: https://www.pivotaltracker.com/story/show/183361067


tests/integration/Infrastructure/DashboardBundle/Validator/Constraints/ValidClientIdValidatorTest.php
```diff
@@ -76,4 +76,18 @@ public function test_invalid_client_id_url()
             'Expected certain violation but dit not receive it.'
         );
     }
+
+    public function test_invalid_client_id_containing_a_port()
```
MKodde, line 80: Nice test coverage! Consider adding a dataProvider here to test multiple client Id's in one go. That way you can test a multitude of illegal clientIds.

Also, what happens when a non numeric port 'number' is specified? Is that also covered by the UrlParser? Say: `https://foobar.example.com:illegal`

MKodde (commented):
Thanks for fixing this bug! See one suggestion below. And what's up with the other two commits in this branch? Did you branch of an outdated develop branch? Or from one of your feature branches? Maybe try to rebase your work onto HEAD of develop. Or drop the two other commits from your branch?

MKodde (approved):
Thanks for hardening the test!

PR #527, opened 2022-09-27
Clear selected service on service switcher when serving sp dashboard

Shaky212:
As soon as you go to spdashboard you will come to the Service overview page. When you select a service and click on the logo the  selected service is cleared and allows you to select any service from the service switcher.

see: https://www.pivotaltracker.com/story/show/182233103


MKodde (approved):
Works like a charm!

Thanks for adding the web test!

PR #528, opened 2022-09-29
Provide a service to fetch and store entity attribute properties

Shaky212:
First attempt to setup a service to fetch and store entity attribute properties. The responsibility of loading the attributes still resides with the legacy AttributeMetaDataRepository.

see https://www.pivotaltracker.com/story/show/182220241


src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeService.php
```diff
@@ -0,0 +1,108 @@
...
+namespace Surfnet\ServiceProviderDashboard\Application\Service;
+
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Attribute;
+use Surfnet\ServiceProviderDashboard\Legacy\Repository\AttributesMetadataRepository;
+
+class AttributeService implements AttributeServiceInterface
+{
+    /**
+     * @var AttributesMetadataRepository
+     */
```
MKodde, line 28: We need to decouple from this legacy Attributes Metadata Repo. Feel free to create a new one and leave the present one where it is for now.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeService.php
```diff
@@ -0,0 +1,108 @@
...
+    private $repository;
+
+    private $attributes = [];
+
+    private $attributeProperties = [];
+
+    public function __construct(AttributesMetadataRepository $repository)
+    {
+        $this->repository = $repository;
+        $this->load();
```
MKodde, line 38: A nice approach to load the attributes at construction time! Tho this will probably look much different when we created our own taylor made repository. But maybe this construct is one to hold on to

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/AttributeProperty.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Surfnet\ServiceProviderDashboard\Application\Service\AttributeServiceInterface;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Attribute;
+
+class AttributeProperty implements AttributeInterface
+{
+    /**
+     * @var AttributeServiceInterface
+     */
+    private $attributeService;
```
MKodde, line 28: An entity should not know about the data provider it originated from. And for another rule of thumb. No application classes can ever be used in the domain layer. I suggest you create the entities in the AttributeService. Possibly useing the static constructor trick we used on one of your previous PR's.

MKodde (commented):
Interesting to see your approach on tackling this problem! See my feedback below. I will whip up a quick class diagram to show you what I personally had in mind. Well not my idea per-se. But how this problem would be solved using the design philosophy implemented in this project. I will slack that to you in the near future.

MKodde (changes requested):
Nice work! Almost good to go.

PR #530, opened 2022-10-05
Set the Timezone on the PHP FPM container

MKodde:
See: https://www.pivotaltracker.com/story/show/183402970


Shaky212 (approved):
Looks good!

PR #531, opened 2022-10-07
Feature/manage entity attributes forms

Shaky212:
SInce the attributes.json has been evolved to a structure with a language sensitive form label and  text an attribute view object has been introduced to hide the complexity the internal structure of attributes. The attribute view object can be used to setup the saml and oidcng entity types attributes dynamically.


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
```diff
@@ -123,7 +134,12 @@ private function buildOptions($environment)
     {
         $options = [];
         if ($environment === Constants::ENVIRONMENT_PRODUCTION) {
-            $options = ['validation_groups' => ['Default', 'production']];
+            $options = ['validation_groups' => ['Default', 'production'],
+                'attribute_service' => $this->attributeService
+            ];
+        }
+        if ($environment === Constants::ENVIRONMENT_TEST) {
+            $options['attribute_service'] = $this->attributeService;
```
MKodde: I don't think you need to inject the attribute_service this way into the type. But I might be wrong..

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -358,6 +231,28 @@ public function __construct()
     {
     }
 
+    public function __set(string $property, ?Attribute $value)
```
MKodde, line 234: Should we allow setting null for an attribute value? You probably walked into this situation so you had to allow for this. But can you explain why?

It might be better to not actually set the value when a null-attribute is encountered. That shoudl also allow you to get rid of the ?Attribute return type on the `__get` and `getAttribute` methods.


An finally as a rule of thumb: methods starting with `get` should either return a result or throw an error when the result can not be found. Where find might result in a null result. I'm not saying you should change this just for the sake of convention. But I just wanted to share that knowledge.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeFormLanguage.php
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Dto;
+
+class Language
```
MKodde: 1. Should this language value object not also be knowledgable about the language it represents?
2. Naming of this VO (not a DTO) is not following the filename.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Dto;
+
+class Form
```
MKodde: The class name should reflect the filename. So either call the file `Form.php` or rename the classname to `AttributeForm`

Regardless of that, the name does not sit right with me. This is a value object that aids the attribute type in displaying the label & info popup text. So a name in that direction would make more sense to me. Suggestions: `AttributeTypeInformation`, `AttributeTypeAttributes`, naming things is hard :smile:

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    public $languages = [];
+
+    private function __construct(
+        array $languages
+    ) {
+        foreach ($languages as $language => $values) {
+            $this->languages[$language] = language::fromLanguage($values);
```
MKodde: 1. As per my previous suggestion. The Language value object should itself also be aware about the language it represents. Its perfectly okay to key your lang-collection on the language as well. 
2. Typo `language::fromLanguage` Classname starts with lower case `l`

tests/unit/Infrastructure/DashboardBundle/Dto/AttributeFormLanguageTest.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $en = [
+                'label' => 'emailAddressAttribute',
+                'info' => 'Description is placed here'
+        ];
+        $language = Language::fromLanguage($en);
+
+        $this->assertEquals('emailAddressAttribute', $language->label);
+        $this->assertEquals('Description is placed here', $language->info);
+    }
```
MKodde, line 35: What happens when malformed data is presented to the VO? Maybe add some tests to verify this?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/AttributeTypeFactory.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
```
MKodde, line 2: Enable strict types?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
```diff
@@ -134,12 +134,7 @@ private function buildOptions($environment)
     {
         $options = [];
         if ($environment === Constants::ENVIRONMENT_PRODUCTION) {
-            $options = ['validation_groups' => ['Default', 'production'],
-                'attribute_service' => $this->attributeService
-            ];
-        }
-        if ($environment === Constants::ENVIRONMENT_TEST) {
-            $options['attribute_service'] = $this->attributeService;
+            $options = ['validation_groups' => ['Default', 'production']];
```
MKodde, line 137: Carefull here. You are removing the distinction between the production and test entity validation groups here.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeService.php
```diff
@@ -30,19 +30,56 @@ class AttributeService implements AttributeServiceInterface
...
+
+    public function __construct(AttributeRepository $attributeRepository, $language)
     {
         $this->attributeRepository = $attributeRepository;
-        $this->load();
+        $this->language = $language;
     }
 
-    public function load(): void
+    public function getAttributes(): array
```
MKodde, line 44: I feel you should reflect in the name of the method, this entails returning a value object specifically used for AttributeType assistance. That way, we are not tempted to use this for other purposes later on.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeServiceInterface.php
```diff
@@ -18,11 +18,13 @@
...
 {
-    public function load(): void;
-
     public function getAttributes(): array;
+
+    public function getNameById(string $id): string;
+
+    public function getLabelById(string $id): string;
+
+    public function getInfoById(string $id): string;
```
MKodde, line 29: Are these actually used? If not. Please remove them for the time being!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Repository/AttributeRepository.php
```diff
@@ -54,4 +51,13 @@ public function findAll(): array
...
     }
+
+    public function findAllNameSpaceIdentifiers(): array
+    {
+        $nameSpaceIdentifiers = [];
+        foreach ($this->getAttributes() as $attribute) {
+            $nameSpaceIdentifiers[] = $attribute->urns[0];
+        }
+        return $nameSpaceIdentifiers;
+    }
```
MKodde, line 62: Is this method ever used? If not please remove it.

MKodde (changes requested):
## General feedback
Aside from some feedback in the code review below, things are looking good. In this review round I also tried to give some feedback on your commit messages. Feel fee to try and improve the existing commits, but you can also use the feedback as a take away for the next PR.

### PR dressing

Please aim to give the PR a descriptive title. Now we have a derivative of the branch name. Which does give a clou.. But having an engaging title helps recognizing features in the future.

Also please describe the purpose of this PR. Judging from the vast ammount of changes, there seem to be a couple of things going on at one time. That is not a bad thing, but an introduction text on what is changed, and why might help a lot.

This can be further cemented by adding good descriptions to your commit messages. You are not yet in the habit of writing these. But they will help you in making better commits! As you are now forced to write about why, when you changed certain parts of the code.

### Git log review feedback

`Add end of file new line` [#b4babb](https://github.com/SURFnet/sp-dashboard/pull/531/commits/b4babb727cb87ff9dfff771e76dc331928f9f4b2)

This commit message seems unrelated to the change made in code. 

`wip` [#5e15cf](https://github.com/SURFnet/sp-dashboard/pull/531/commits/5e15cf3871d5a5f2e4fa184278e9e88fad3916ab)

I think you forgot to reset this commit? This seems to be a very diverse (non atomic commit). 

`Introduce attribute view object to hide internal complexity of attributes`
`Refactor Attribute to meet new attribute structure with form and language fields`

Titles are too long (restrict to 50 characters) The quality of the commit title is much better tho!

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
Shaky212: Thanks! On second thought I think a better way to use the __set and __get is to call a method the setAttribute and getAttribute and move the existence of the key in those methods, leaving the __get quite clean. I will refactor that in the next stage.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
Shaky212: Agree, AttributeTypeInformation makes more sense.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
Shaky212: I will add a language property to make aware about the language.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/AttributeTypeFactory.php
Shaky212: Yes, I will add this to the next stage (commands)

PR #532, opened 2022-10-10
Allow closing of flash messages

MKodde:
In testing the recent flash message changes. PM found out that a means of closing the messages would be nice. This PR adds the possibility to close the flash messages by clicking on a `x` button at the topmost right position of the message.

A feature that was already supported in the teams flash messages.

Another avenue of (auto) closing the flashes was suggested. But that proved difficult without rewriting the CSP policies. Allowing for (unsafe) eval functionality as an auto-close feature would ential use of the `setTimeout` function.

https://www.pivotaltracker.com/story/show/182246029/comments/233434060


Shaky212 (approved):
Looks good!

PR #533, opened 2022-10-10
Manage the entity attributes commands

Shaky212:
The existing saml and oidc entity forms let loose their data into corresponding Command objects. These objects now also enumerate every possible Attribute.  Now these attributes are removed from the commands and replaces by an Attribute array which is loaded from the attributes.json config. 

The saml and oidc form use the magic setter and getter of the command objects. On their turn these magic getter and setter use a more readable and useful local setter and getter. Perhaps a more elegant solution is possible after migration of the project to version 5 or higher by using PropertyAccessors in the form, or the 'getter' and 'setter' features of the form builder. Ideally the magic getter and setter could possibly be removed.

Some simple validation is done on the attributes by using a assertion call back on the properties of the saml and oidc commands.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
@@ -25,6 +25,7 @@ use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Attribute;
 use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Contact;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Validator\Constraints as SpDashboardAssert;
 use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
 
 /**
  * @SuppressWarnings(PHPMD.TooManyFields)
@@ -222,12 +223,41 @@ class SaveSamlEntityCommand implements SaveEntityCommandInterface
     {
     }
 
+    /**
+     * @Assert\Callback()
+     */
+    public function validate(ExecutionContextInterface $context, $payload)
+    {
+        foreach ($this->attributes as $name => $attribute) {
+            if (isset($attribute) && !($attribute instanceof Attribute)) {
+                $context->buildViolation(sprintf('entity.edit.attribute.invalid', $name))
+                    ->atPath('attribute')
+                    ->addViolation();
+            }
+        }
+    }
+
+    /**
+     * The magic getters and setters are consulted by the saml form builder.
+     * Another option would be to implement a dataMapper on the
+     * form or attribute container, but this might lead to needless complexity.
+     */
     public function __set(string $property, ?Attribute $value)
+    {
+        $this->setAttribute($property, $value);
+    }
+
+    public function __get(string $property): ?Attribute
+    {
+        return $this->getAttribute($property);
+    }
+
+    public function setAttribute(string $property, ?Attribute $value)
     {
         $this->attributes[$property] = $value;
     }
 
-    public function __get(string $property): ?Attribute
+    public function getAttribute(string $property): ?Attribute
     {
         if (array_key_exists($property, $this->attributes)) {
             return $this->attributes[$property];
@@ -235,15 +265,6 @@ class SaveSamlEntityCommand implements SaveEntityCommandInterface
         return null;
     }
 
-    public function setAttribute(string $property, ?Attribute $value)
-    {
-        $this->__set($property, $value);
-    }
-
-    public function getAttribute(string $property): ?Attribute
-    {
-        return $this->__get($property);
-    }
     /**
      * @param Service $service
      * @return SaveSamlEntityCommand
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
@@ -44,7 +44,7 @@ class ArpGenerator implements MetadataGenerator
         $attributes = [];
         $entityAttributes = $entity->getAttributes();
         foreach ($this->attributeRepository->findAll() as $definition) {
-            $urns = $definition->getUrns();
+            $urns = $definition->urns;
             $urn = reset($urns);
             $attributeList = $entityAttributes->findAllByUrn($urn);
             // Only add the attributes with a motivation
@@ -80,8 +80,8 @@ class ArpGenerator implements MetadataGenerator
 
     private function addManageOnlyAttributes(array &$attributes, ManageEntity $entity)
     {
-        $spDashboardTracked = $this->attributeRepository->findAllNameSpaceIdentifiers();
         $originalAttributes = $entity->getAttributes()->getOriginalAttributes();
+        $spDashboardTracked = $this->attributeRepository->findAllNameSpaceIdentifiers();
         $nonTrackedUrns = array_diff(array_keys($originalAttributes), $spDashboardTracked);
 
         foreach ($nonTrackedUrns as $urn) {
--- a/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
@@ -19,6 +19,7 @@
 namespace Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator;
 
 use stdClass;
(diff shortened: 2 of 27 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
```diff
@@ -44,7 +44,7 @@ public function build(ManageEntity $entity): array
         $attributes = [];
         $entityAttributes = $entity->getAttributes();
         foreach ($this->attributeRepository->findAll() as $definition) {
-            $urns = $definition->getUrns();
+            $urns = $definition->urns;
```
MKodde, line 47: Should this change be in this PR? I'm not going to reject your PR over this. But please be aware of what you include in your commits.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeNameService.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->attributeRepository = $attributeRepository;
+    }
+
+    public function getAttributeTypeNames(): array
+    {
+        $attributes = $this->attributeRepository->findAll();
+
+        $attributeNames = [];
+        foreach ($attributes ?? [] as $attribute) {
+            $attributeNames[] = $attribute->id . Constants::ATTRIBUTE_NAME_SUFFIX;
```
MKodde, line 44: I'd prefer you add the constant to the `AttributeNameServiceInterface`. Given this is a very application/infrastructural thing. The Constants are (hopefully) reserved for more domain related settings.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeNameService.php
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Application\Service;
+
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
+use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Repository\AttributeRepository;
+
+class AttributeNameService implements AttributeNameServiceInterface
```
MKodde, line 26: Love the addition of this service!

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
```diff
@@ -177,7 +176,7 @@ private function buildAttributesFromCommand(SaveEntityCommandInterface $command)
             return $attributeList;
         }
 
-        foreach ($this->attributeRepository->findAll() as $definition) {
+        foreach ($this->attributeService->getAttributeTypeAttributes() as $definition) {
```
MKodde, line 179: Please create a service method that reads the attributes tailor-made for this application. The entity merge service has nothing to do with the AttributeType form component. So using those value objects here is convenient, but misplaced.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Constants.php
```diff
@@ -46,6 +46,8 @@ class Constants
 
     const OIDC_SECRET_LENGTH = 20;
 
+    const ATTRIBUTE_NAME_SUFFIX = 'Attribute';
```
MKodde, line 49: As already proposed, please move this to the `AttributeNameServiceInterface`.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/MetaData.php
```diff
@@ -257,6 +257,18 @@ private static function getAcsLocationsFromMetaDataFields(array $metaDataFields)
...
+    {
+        $data = [];
+        foreach ($this->getAcsLocations() ?? [] as $index => $location) {
+            $location = sprintf('metaDataFields.AssertionConsumerService:%d:Location', $index);
+            $binding = sprintf('metaDataFields.AssertionConsumerService:%d:Binding', $index);
+            $data[$location] = $location;
+            $data[$binding] = Constants::BINDING_HTTP_POST;
+        }
+        return $data;
+    }
```
MKodde, line 270: Boyscout rule applied :medal_military: 

`$this->getAcsLocations() ?? []`: convenient solution to not having to wrap the foreach in an if block :+1: Might have to add this one to my own toolbelt

tests/integration/Legacy/Metadata/ParserTest.php
```diff
@@ -43,9 +44,10 @@ public function setup()
         $this->logger = m::mock(LoggerInterface::class);
 
         $rootDir = __DIR__.'/../../../../app/Resources/';
+        $attributeRepository = new AttributeRepository(__DIR__ . '/../../../../app/config/attributes.json');
```
MKodde, line 47: Is using the real config file the way to go here? I'd propose you copy paste it to a `fixtures` folder in the folder where this test resides.

This pattern also emerges in: 
- EntityMergeServiceTest.php
- ArpGeneratorTest.php

tests/unit/Application/Metadata/JsonGenerator/ArpGeneratorTest.php
```diff
@@ -36,7 +36,7 @@ class ArpGeneratorTest extends MockeryTestCase
 
     public function setUp()
     {
-        $this->attributeRepository = new AttributeRepository(__DIR__ . '/../../../../../app/Resources/metadata');
+        $this->attributeRepository = new AttributeRepository(__DIR__ . '/../../../../../app/config/attributes.json');
```
MKodde, line 39: The arp generator should not have been changed in this PR right? Maybe this is an attempt to fix the unit tests? The real fix here should be to use the old AttributeRepository on the ArpGenerator. 

Lets discuss if this should be fixed here, or if this is just some collateral damage we want to live with at this point.

MKodde (changes requested):
I'm super exited about the git-log quality improvement! No more meaningless 'Fixed this/that' commits! All of them are clearly portraying the purpose they intended.

I did find some breaking changes that need some attention. See the code review below.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -222,28 +223,48 @@ public function __construct()
...
+    public function validate(ExecutionContextInterface $context, $payload)
+    {
+        foreach ($this->attributes as $name => $attribute) {
+            if (isset($attribute) && !($attribute instanceof Attribute)) {
+                $context->buildViolation(sprintf('entity.edit.attribute.invalid', $name))
+                    ->atPath('attribute')
+                    ->addViolation();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Shaky212: Discussed this and fix this in a new PR

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -222,28 +223,48 @@ public function __construct()
...
+        $this->attributes[$property] = $value;
     }
 
     public function getAttribute(string $property): ?Attribute
     {
-        return $this->__get($property);
+        if (array_key_exists($property, $this->attributes)) {
+            return $this->attributes[$property];
+        }
+        return null;
```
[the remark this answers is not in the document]
Shaky212: The PropertyAccessor of the symfony form requires an Attribute or a null value. Returning an exception results in a Fatal Error. Such an exception might be caught when using a dedicated PropertyAccessor at the form builder. But that is for a future version of Symfony to find out.

src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
Shaky212: Refined that and used an  EntityMergeAttribute valueObject to service the entityMergeService

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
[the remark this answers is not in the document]
MKodde: Ahh, okay. Let's leave it at this then. Or consider renaming the method to findAttribute.. But lets leave it at this.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator/ArpGenerator.php
MKodde: Almost there! You forgot to add the new service methods (getUrns && getEntityMergeAttributes) to the `AttributeServiceInterface`

MKodde (approved):
:shipit:

PR #535, opened 2022-10-11
Exclude the EPTI attribute on the non SAML forms

MKodde:
This was a left-over from the #531 PR. Where the Oidc form would display the EPTI attribute. Something that should not be the case.


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
```diff
@@ -20,22 +20,28 @@
...
     }
 
-    public static function fromForm(array $languages): ?AttributeForm
+    public static function fromForm(array $form): ?AttributeForm
     {
-        return new self(
-            $languages
-        );
+        $excludeFrom = [];
+        if (array_key_exists('excludeOnEntityType', $form)) {
```
Shaky212, line 42: Is it justified to use a constant for 'excludeOnEntityType'?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/OidcngEntityType.php
```diff
@@ -366,7 +366,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
 
     private function buildAttributeTypes(FormBuilderInterface $container): FormBuilderInterface
     {
-        return $this->attributeTypeFactory->build($container);
+        return $this->attributeTypeFactory->build($container, Constants::TYPE_OPENID_CONNECT_TNG);
```
Shaky212, line 369: Not for now, but maybe an idea for the future to divide a large builder into smaller methods to apply structure,  methods like buildMetadata, buildOrganization, buildAttributesTypes, etc

app/config/attributes.json
```diff
@@ -2,9 +2,11 @@
   {
     "id": "givenName",
     "form" : {
-      "en": {
-        "label": "Given name attribute",
-        "info": "Text should be set in web translations"
+      "translations" : {
```
Shaky212, line 5: Adding translations makes sense and gives more structure!

Shaky212 (approved):
Nice solution to exclude an attribute from the premises. Added some suggestions/remarks but no show stoppers.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Dto/AttributeForm.php
MKodde: Good suggestion. At this point, this is the only place where we consume that key in the application. And it should remain the only place. I do not see the need to use a class constant in this case.

MKodde:
Merging :shipit:

PR #539, opened 2022-10-17
Attribute validator for an entity attribute array 

Shaky212:
When the attributes of an entity were hard coded each attribute had its own validation. In the new situation were the attribute are dynamic and built from the attributes.json a new validator (service) has been introduced to check if the attributes are still valid. In the new situation all attributes in the form have to validated against the attribute.json.


tests/integration/Application/Validator/Constraints/ValidAttributeValidatorTest.php
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+namespace integration\Application\Validator\Constraints;
+
+use Surfnet\ServiceProviderDashboard\Application\Service\AttributeService;
+use Surfnet\ServiceProviderDashboard\Application\Validator\Constraints\ValidAttributeValidator;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\Attribute;
+use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Repository\AttributeRepository;
+use Surfnet\ServiceProviderDashboard\Application\Validator\Constraints\ValidAttribute;
+use Surfnet\ServiceProviderDashboard\Tests\Integration\Application\Validator\Constraints\mock;
```
MKodde, line 26: Forgot to remove a dev leftover?

tests/integration/Application/Validator/Constraints/ValidAttributeValidatorTest.php
```diff
@@ -0,0 +1,153 @@
...
+        return new ValidAttributeValidator($attributeService);
+    }
+
+    public function test_success_no_attributes()
+    {
+        $constraint = new ValidAttribute();
+        $this->validator->validate([], $constraint);
+
+        $this->assertNoViolation();
+    }
```
MKodde, line 44: May I suggest you add another happy flow test. Where you verify it can test multiple attributes? Something along these lines:

```
    public function test_success_multiple_requested_attributes()
    {
        $constraint = new ValidAttribute();
        $this->validator->validate($this->buildMultipleAttributes(
            ['emailAddressAttribute', 'eduPersonTargetedIDAttribute', 'surNameAttribute',]
        ), $constraint);
        $this->assertNoViolation();
    }


    private function buildMultipleAttributes(array $names): array {
        $attributes = [];
        foreach ($names as $name) {
            $requested = true;
            $motivation = 'Motivation';
            $attributes[$name] = new Attribute();
            $attributes[$name]
                ->setRequested($requested)
                ->setMotivation($motivation);
        }
        return $attributes;
    }
```

MKodde (changes requested):
These are my code review findings. I'm very content with your solution. There are some details that need some attention.

In my functional review I found another issue that needs attention:

Screenshot from 2022-10-17 12-14-46
1. The attributes that did not pass validation are missing the attribute friendly name in the validation message. 
2. What might be even better: display the messages in the form, right above the form field. But having that preciese of a mapping to the field might be hard?

MKodde:
I ended up fixing my review feedback myself. And got stuck on trying to set the path of the validation messages. With the aim to display them on the attribute form instead of having them on top of the page. This proved difficult as the attributes are created dynamically.

MKodde (approved):

PR #540, opened 2022-10-17
Clean up of attribute related messages

Shaky212:
All attribute related messages that are no longer used or moved to the attributes.json, due to the dynamic making of the attributes of the entities, have now been cleaned up.


MKodde (changes requested):
Aside from the removed `crlf`, all is looking good!

MKodde (approved):

PR #541, opened 2022-10-17
Set only email on custom reporter Jira field

MKodde:
Previously we would also show the reporter name on that field. But this field can only be a mail address


quartje:
This fixes the issue! :rocket:

quartje (approved):

PR #544, opened 2022-10-18
Attribute schema validator

Shaky212:
When attributes are maintained within the attributes.json some validation has to be done to avoid unwanted errors. An attributes.scheme.json has been added to check to attributes.json against this scheme. The initial version of the scheme is generated by https://extendsclass.com/json-schema-validator.html. Further refinement is done by adding so called 'patternProperties' to support multiple languages easily by extending the pattern "^en$" by e.g. "^en|nl$" to support Dutch. Another adjustment to mention is the enumeration of supported entityt ypes within the excludeOnEntityType property.

see: https://www.pivotaltracker.com/story/show/182219948

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -22,6 +22,7 @@ test code as well as production code;
  - PHP Mess Detector checks a number of metrics, and if they exceed a certain treshold the build will fail;
  - PHP CodeSniffer ensures that the code adheres to the chosen coding standard (PSR-2);
  - PHP Copy-Paste Detector ensures that there is no substantial duplication within the source code.
+ - A JSON schema validator ensures that the JSON-schema is valid and validates the JSON-data against the JSON-schema.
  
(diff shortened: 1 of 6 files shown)


docs/testing.md
```diff
@@ -22,6 +22,7 @@ test code as well as production code;
  - PHP Mess Detector checks a number of metrics, and if they exceed a certain treshold the build will fail;
  - PHP CodeSniffer ensures that the code adheres to the chosen coding standard (PSR-2);
  - PHP Copy-Paste Detector ensures that there is no substantial duplication within the source code.
+ - A JSON schema validator ensures that the JSON-schema is valid and validates the JSON-data against the JSON-schema.
```
MKodde, line 25: Super nice to also have the `compile` step, to validate the validator :100:

MKodde (changes requested):

MKodde (approved):

PR #546, opened 2022-10-18
Test failure on removed attribute(s)

Shaky212:
Some tests failed when attributes were removed from the attributes.json. This is now fixed to fetch the attributes from the attributes service and build up the form data for the attributes dynamically


MKodde (approved):
Feel free to merge once the build is green! Nice job!

PR #547, opened 2022-10-19
Set ACS location on MetaData array representation

MKodde:
Not the actual acs uri was set on the array representation, but the field identifier. That resulted in the change request not registering any changes between entity revisions.

See: https://www.pivotaltracker.com/story/show/182411730/comments/233831779

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/MetaData.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/MetaData.php
@@ -272,10 +272,10 @@ class MetaData implements Comparable
 
         if ($this->getAcsLocations()) {
             foreach ($this->getAcsLocations() as $index => $location) {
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/MetaData.php
```diff
@@ -272,10 +272,10 @@ public function asArray(): array
 
         if ($this->getAcsLocations()) {
             foreach ($this->getAcsLocations() as $index => $location) {
-                $location = sprintf('metaDataFields.AssertionConsumerService:%d:Location', $index);
-                $binding = sprintf('metaDataFields.AssertionConsumerService:%d:Binding', $index);
-                $data[$location] = $location;
```
Shaky212, line 277: Grrrr, index changed...

Shaky212 (approved):

PR #550, opened 2022-10-19
Repair hardcoded saml change request to protocol within the entity

Shaky212:
The protocol was hardcoded to saml_sp. Now the protocol used within the entity is used when requesting the changes from manage.

see: https://www.pivotaltracker.com/story/show/182411816

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
@@ -19,6 +19,7 @@
 namespace Surfnet\ServiceProviderDashboard\Application\Service;
 
 use Surfnet\ServiceProviderDashboard\Application\Dto\ChangeRequestDtoCollection;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Protocol;
 use Surfnet\ServiceProviderDashboard\Domain\Repository\EntityChangeRequestRepository;
 
 class ChangeRequestService implements ChangeRequestServiceInterface
@@ -37,9 +38,9 @@ class ChangeRequestService implements ChangeRequestServiceInterface
     /**
      * @throws \Exception
      */
-    public function findById(string $id): ChangeRequestDtoCollection
+    public function findByIdAndProtocol(string $id, Protocol $protocol): ChangeRequestDtoCollection
     {
-        $values = $this->repository->getChangeRequest($id);
+        $values = $this->repository->getChangeRequest($id, $protocol);
         return new ChangeRequestDtoCollection($values);
     }
 }
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * Copyright 2022 SURFnet B.V.
  *
@@ -18,7 +20,6 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\Service;
 
-use Exception;
 use Surfnet\ServiceProviderDashboard\Application\Dto\ChangeRequestDtoCollection;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Protocol;
 use Surfnet\ServiceProviderDashboard\Domain\Repository\EntityChangeRequestRepository;
@@ -36,19 +37,9 @@ class ChangeRequestService implements ChangeRequestServiceInterface
         $this->repository = $repository;
     }
 
-    /**
-     * @param string $id
-     * @param Protocol $protocol
-     * @return ChangeRequestDtoCollection|Exception
-     */
     public function findByIdAndProtocol(string $id, Protocol $protocol): ChangeRequestDtoCollection
     {
-        $values = [];
-        try {
-            $values = $this->repository->getChangeRequest($id, $protocol);
-        } catch (Exception $exception) {
-            throwException($exception);
-        }
+        $values = $this->repository->getChangeRequest($id, $protocol);
         return new ChangeRequestDtoCollection($values);
     }
 }
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
@@ -18,6 +18,7 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\Service;
 
+use Exception;
 use Surfnet\ServiceProviderDashboard\Application\Dto\ChangeRequestDtoCollection;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Protocol;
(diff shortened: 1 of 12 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestServiceInterface.php
```diff
@@ -19,8 +19,9 @@
 namespace Surfnet\ServiceProviderDashboard\Application\Service;
 
 use Surfnet\ServiceProviderDashboard\Application\Dto\ChangeRequestDtoCollection;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Protocol;
 
 interface ChangeRequestServiceInterface
 {
-    public function findById(string $id): ChangeRequestDtoCollection;
+    public function findByIdAndProtocol(string $id, Protocol $protocol): ChangeRequestDtoCollection;
```
MKodde, line 26: Great name change!

I sometimes also like to keep the base name of the method short, and let the parameters describe by what criteria we are searching. Example:

`findBy(string $id, Protocol $protocol)`

But please leave it as-is!

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/changeRequest.html.twig
```diff
@@ -32,7 +32,17 @@
                     {% for index, pathUpdate in changeRequest.pathUpdates %}
                     <tr>
                             <td>{{ index }}</td>
-                            <td>{{ pathUpdate }}</td>
+                            <td>
+                            {% if pathUpdate is iterable %}
+                                <ul>
+                                    {%  for name, value in pathUpdate %}
+                                        <li>{{ name }}: {{ value }}</li>
```
MKodde, line 39: Name would usually be a numeric index. And that makes little or no sense to the end user I suppose. I think we can get away with just echoing the value here.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Client/EntityChangeRequestClient.php
```diff
@@ -79,12 +80,12 @@ public function openChangeRequest(ManageEntity $entity, ?ManageEntity $pristineE
...
     }
 
-    public function getChangeRequest(string $id): array
+    public function getChangeRequest(string $id, Protocol $protocol): array
     {
         $this->logger->info(sprintf('Get outstanding change requests from manage for entity "%s"', $id));
 
         return $this->client->read(
-            '/manage/api/internal/change-requests/saml20_sp/' . $id
+            '/manage/api/internal/change-requests/' . $protocol->getManagedProtocol() . '/' . $id
```
MKodde, line 88: Now that we are starting to create a concoction of several strings, I feel it's time to use good 'ol `sprintf` to make things readable and maintainable.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/Protocol.php
```diff
@@ -77,4 +77,9 @@ public function asArray(): array
             'type' => $this->getProtocol(),
         ];
     }
+
+    public function getManagedProtocol(): string
```
MKodde, line 81: Please cover this method in test. Simple as it is.. It might prove useful at a later stage.

MKodde (changes requested):
Great work so far!

*Edit* after functional review:
Editing a published oauth_ccc entity did not result in the correct situation yet. The change request is created in Manage. But the wrong protocol is read to get the data from. You can see which protocol mapping to use for that entity type in the Protocol value object.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/Protocol.php
```diff
@@ -77,4 +77,9 @@ public function asArray(): array
...
+    {
+        return array_search($this->protocol, self::$protocolMapping);
+    }
```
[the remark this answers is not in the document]
Shaky212: Nope, did not see any yellow swirly lines. I guess this is because the array_search returns a mixed value of integer|string|false. So the function covers the string, but not the other types. I will make the function more verbose and implement your suggestion, together with the exception.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/views/EntityPublished/changeRequest.html.twig
Shaky212: Indeed meaningless index

src/Surfnet/ServiceProviderDashboard/Infrastructure/Manage/Client/EntityChangeRequestClient.php
Shaky212: Agree

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/Protocol.php
Shaky212: Added a few protocol tests

src/Surfnet/ServiceProviderDashboard/Application/Service/ChangeRequestService.php
```diff
@@ -35,11 +37,18 @@ public function __construct(
...
      */
-    public function findById(string $id): ChangeRequestDtoCollection
+    public function findByIdAndProtocol(string $id, Protocol $protocol): ChangeRequestDtoCollection
     {
-        $values = $this->repository->getChangeRequest($id);
+        $values = [];
+        try {
+            $values = $this->repository->getChangeRequest($id, $protocol);
+        } catch (Exception $exception) {
+            throwException($exception);
```
MKodde, line 50: 1. This construction seems irrelevant. You catch an exception and again throw it. Which is basically a no-operation.
2. Also, using `throwException` (a function I'm unaware of) is not in line with the way we are used to throw exceptions (just use `throw $exception` or `throw new FoobarException('Rudy, a message to you, Rudy')`
3. The find method should not throw an exception, but return an empty result if the data is not found. (getter throws exceptions, find methods allow for returning a null result.)

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/Protocol.php
```diff
@@ -18,6 +18,7 @@
 
 namespace Surfnet\ServiceProviderDashboard\Domain\Entity\Entity;
 
+use Exception;
```
MKodde, line 21: Using a self-made exception is more in line with how we roll in the domain layer. Suggestion: `ProtocolNotFoundException`, `ProtocolMappingFailedException`,...

MKodde (changes requested):
Thanks for addressing the changes from the previous review. Two new things popped up.

MKodde:
Thanks for addressing the final feedback. I did a small additional cleanup task to remove some forgotten import statements. And removed an unused dependency from the ChangeRequestClient.

MKodde (approved):

PR #551, opened 2022-10-19
Handle grants and redirect urls correctly when updating an OIDC entity

MKodde:
When diffing entity changes, the grants and redirect urls must be included on the 'add all values' and not in a diff manner. I previously implemented this. But it was added to the diff also when no actual changes where applied to the data.

By only adding them when the diff result yielded a result. I worked around that problem.

The test shows the different combinations. And the expected outcomes. That should help in getting a better grasp of the correct behavior.

### DIFF
--- a/tests/unit/Infrastructure/Manage/Dto/EntityDiffTest.php
+++ b/tests/unit/Infrastructure/Manage/Dto/EntityDiffTest.php
@@ -71,6 +71,52 @@ class EntityDiffTest extends MockeryTestCase
         $this->assertSame([1 => 'is for duck'], $diff['d']);
     }
 
+
+    public function test_grants_and_redirect_urls_are_handled_their_own_special_way()
+    {
+        // No change should result in the redirectUrls not showing up
+        $data = ['metaDataFields.redirectUrls' => ['https://uri1.example.com']];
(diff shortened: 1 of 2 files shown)


tests/unit/Infrastructure/Manage/Dto/EntityDiffTest.php
```diff
@@ -71,6 +71,52 @@ public function test_recursion()
         $this->assertSame([1 => 'is for duck'], $diff['d']);
     }
 
+
+    public function test_grants_and_redirect_urls_are_handled_their_own_special_way()
+    {
+        // No change should result in the redirectUrls not showing up
+        $data = ['metaDataFields.redirectUrls' => ['https://uri1.example.com']];
```
Shaky212, line 78: I understand that the arrays can be nested and iterated recursively. I don't see any reflection of this in the test.

Shaky212 (approved):

tests/unit/Infrastructure/Manage/Dto/EntityDiffTest.php
MKodde: The new tests and the `test_recursion` test go into this feature. In retrospect. We probably never actually needed the recursive nature in the Diff feature. As all array typed metadata options are not to be PUT in a diffed manner. But rather in the 'provide all array values' way..

PR #552, opened 2022-10-20
Show an overview of open Jira tickets concerning connection requests

Shaky212:
We want to make the open Jira tickets with regard to the connection requests transparent for a user. For the saml and oidcng entities an extra option is served to show the open connection requests.

In this first implementation of the connection request no checks are done to see if there are any connections request. This is a rudimentary setup and will be expanded from there. The results from jira are simulated by returning some hard coded data from the EntityConnectionRequestClient.

see: https://www.pivotaltracker.com/story/show/183572833


MKodde:
This can be closed now. As we no longer need to create an overview of the connection requests.

PR #555, opened 2022-10-25
Create connection requests and send them to Jira

Shaky212:
Via JIRA a supplier has the option to request an IdP to link the service. So they can make an indirect link request.

After publishing a production link, the SP Dashboard displays a form where the SP has to fill in the institution,name and email. In addition, the form can also be called up from the context menu on the services overview.

see: https://www.pivotaltracker.com/story/show/183579834


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateConnectionRequestController.php
```diff
@@ -0,0 +1,141 @@
...
+        if ($form->isSubmitted()) {
+            if ($this->isCancelAction($form) || !count($command->getConnectionRequests())) {
+                return $this->returnToOverview($serviceId);
+            }
+
+            if ($form->isValid()) {
+                $this->commandBus->handle($command);
+                return $this->render(
+                    '@Dashboard/EntityPublished/publishedConnectionRequest.html.twig'
+                );
```
MKodde, line 123: This should be a redirect response to a stand alone 'thank you' endpoint (as per point 2 of the functional feedback :point_up_2: ). 

Also semantically the name of the twig template is not very good. The request is not published. It stores a Jira ticket for the Service Desk to action at some point.

tests/webtests/CreateConnectionRequestTest.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Webtests;
+
+use Surfnet\ServiceProviderDashboard\Application\Command\Entity\CreateConnectionRequestCommand;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\ConnectionRequest;
+use Symfony\Component\DomCrawler\Field\ChoiceFormField;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+
+class CreateConnectionRequestTest extends WebTestCase
```
MKodde, line 26: I'd say. Drop this test in favor of the Cypress test.

MKodde (changes requested):
1. Visually, the create connection request form collection renders the filled in connection requests beneath the form. They are visually separated with a blue horizontal rule. Nice for the first entry, but in my opinion, the rest of the connection requests do not benefit from that hard visual separation.
2. On `https://spdashboard.vm.openconext.org/entity/create-connection-request-from-overview/production/entity-uuid-prod-entity/2`, pressing F5 causes the form to be submitted again. Resulting in the `CreateConnectionRequestCommand` to be handled over and over again. Resulting in multiple (duplicate) Jira tickets being created?
3. Nitpick: The text displayed on the `create-connection-request-from-overview` thank you page is somewhat vague. `Successfully published the connection request(s) to production`: The connection requests are not published to a production environment. Rather, a connection request is submitted to the service desk. This text can be translated by Bart and Teun, so this is not that big of a deal. But having a sane default is not a bad idea either. 
4. I'd like the Cypress test to function in this PR. Added bonus. This test can be used as a blueprint to test the ACS locations test we spoke of yesterday.

MKodde (approved):

PR #559, opened 2022-11-04
At the change request overview rename table header fields

Shaky212:
At the change request overview the requested header table field names are: 
- left column: Field
- right column: New value

see: https://www.pivotaltracker.com/story/show/182411816


MKodde:
I see no harm in this change. FYI this translation could also be translated in the web-translation system. That way, no code change is required to translate `messages.en.yml` entries.

MKodde (approved):

PR #563, opened 2022-11-21
Log the entityID of the entity that has double ACS locations

quartje:


tests/unit/Domain/Entity/Entity/MetaDataTest.php
```diff
@@ -46,7 +46,7 @@ public function test_it_throws_exception_on_invalid_acs_location()
     public function test_it_throws_exception_on_double_acs_locations()
     {
         $data['data'] = json_decode(file_get_contents(__DIR__.'/fixture/read_response_double_acs_location.json'), true);
-        $this->expectExceptionMessage('Double acs locations. Expected unique locations');
+        $this->expectExceptionMessage('Double acs locations. Expected unique locations for entity: ' . $entityId);
```
thijskh, line 49: `$entityID` is not defined in this test. The test aims to verify that the correct exception message is generated. So you need to put the actual entityID from the testdata in the line above into the message.

quartje: I already thought so. I seemed counterintuitive to use a hardcoded value.

thijskh: For a test it's not so problematic since both the testdata and the test itself are closely coupled.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Entity/MetaData.php
```diff
@@ -93,7 +93,7 @@ public static function fromApiResponse(array $data)
         Assert::string($nameNl);
 
         if (count(array_unique($acsLocations)) !== count($acsLocations)) {
-            throw new Exception('Double acs locations. Expected unique locations');
+            throw new Exception('Double acs locations. Expected unique locations for entity: ' . $entityId);
```
MKodde, line 96: Nitpick hour :warning:  Our go-to standard is to use `sprintf` for string concatenations. Given this is a simple enough onstruction, feel free to leave as-is!

PR #564, opened 2022-11-21
Docker: Install assets during the build

quartje:


docker/Dockerfile
```diff
@@ -3,7 +3,8 @@ COPY . /var/www/html/
 WORKDIR /var/www/html/
 ENV SYMFONY_ENV=prod
 RUN composer install -n --prefer-dist -o --ignore-platform-reqs --no-dev && \
-    composer run-script symfony-scripts
+    composer run-script symfony-scripts && \
```
MKodde, line 6: Symfony 4 with Flex no longer recognizes the `symofny-scripts` scripts. One of the SF scripts will probably be the assets:install portion. But that is already taken care of in the composer.json (where SF now stores its post install scripts). 

@Shaky212 will probably be able to fill you in with the fine details.

MKodde (changes requested):

PR #566, opened 2022-11-24
Change button label to Change Request when editing an entity

Shaky212:
During the creation of an entity, the submit button in the form is labeled Publish. When an entity is edited the label name is Change request.

see: https://www.pivotaltracker.com/story/show/183736253

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
@@ -70,18 +70,18 @@ class EntityTypeFactory
             case ($type === Constants::TYPE_SAML):
                 $command = SaveSamlEntityCommand::forCreateAction($service);
                 $command->setEnvironment($environment);
-                return $this->formFactory->create(SamlEntityType::class, $command, $this->buildOptions($environment));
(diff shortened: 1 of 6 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
```diff
@@ -119,23 +119,34 @@ public function createEditForm(ManageEntity $entity, Service $service, string $e
...
         }
         return $options;
     }
+
+    private function editBuildOptions($environment)
+    {
+        $options = [];
+        if ($environment === Constants::ENVIRONMENT_PRODUCTION) {
+            $options = ['validation_groups' => ['Default', 'production'],
+                'publish_button_label' => 'Change request',
```
MKodde, line 147: Shouldn't this text be translatable? I see this used to be a hard-coded text before your change. But still..

MKodde (approved):
LGTM! See my one question.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Factory/EntityTypeFactory.php
Shaky212: Agree, but perhaps better to check all buttons and make a general change, for example in the upgrade SF 5

PR #567, opened 2022-12-05
Label and translation adjustments in attributes.json

Shaky212:
The following major changes have been applied to attributes.json and it's scheme:

#1 Labels and info attributes must be different for oidcng entities and saml20 entities
#2 No distinction between the translations for the edit screen and info screen.

see: https://www.pivotaltracker.com/story/show/183736168


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidAttribute.php
```diff
@@ -31,11 +32,18 @@ class ValidAttribute extends Constraint
...
      * @return string
      */
     public function validatedBy()
     {
         return 'valid_attribute';
     }
+
+    public function getDefaultOption()
+    {
+        return 'type';
```
MKodde, line 47: I don't get this, does this indicate it should look to the `type` field for the default option value?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidAttribute.php
```diff
@@ -31,11 +32,18 @@ class ValidAttribute extends Constraint
 
     public $messageAttributeMotivationNotSet = 'validator.attribute.motivation_not_set';
 
+    public $type = [Constants::TYPE_SAML, Constants::TYPE_OPENID_CONNECT_TNG];
```
MKodde, line 35: These seem to be the allowed values for the type field, is that correct? 

If that's the case. Would these values not be overwritten when setting the `type` on the command? I'd expect these values to be used to verify we are setting a valid value for `type`.

src/Surfnet/ServiceProviderDashboard/Application/Service/AttributeService.php
```diff
@@ -90,31 +91,75 @@ public function getUrns(): array
...
+        Attribute $attribute,
+        string $entityType
+    ) {
+        switch ($entityType) {
+            case Constants::TYPE_SAML:
+                return $attribute->getSaml20Label();
+            case Constants::TYPE_OPENID_CONNECT_TNG:
+                return $attribute->getOidcngLabel();
+            default:
+                return '';
```
MKodde, line 134: 1. Returning an empty string when the entity type is not supported feels like a situation to create strange situations. This look like a great situation to be throw an exception instead. (also goes for the other getters below)
2. Please add a return type definition on these private methods.

MKodde (changes requested):

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidAttribute.php
Shaky212: It is a dev left-over

MKodde (approved):
Looks good! See my one question. Maybe this is irrelevant. You decide.

app/Resources/schemas/attributes.scheme.json
```diff
@@ -1,149 +1,101 @@
...
 					"en"
 				],
-				"patternProperties": {
-					"^en$": {
-						"$id": "#root/items/detail/en", 
-						"title": "En", 
+				"properties": {
+					"en": {
+						"$id": "#root/items/translations/en",
+						"title": "En",
```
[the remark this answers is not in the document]
Shaky212: New languages can be entered, but are not checked. Introduce patternProperties here will help to match other 2 character patterns to indicate a language and check the properties at a lower level.

MKodde (approved):
Nice work, lets go ahead with this.

Some CI observations: 

1. the security checkers should not be an allowed failure
2. both yarn audit and the composer audit tasks fail in this build. If this is fixed before a new 4.3 release, thats fine with me. But be aware of this!

PR #568, opened 2022-12-12
Show change request action only when available

Shaky212:
Only show the option 'Change request' in the action list of an  entity when an entity has pending change request.

see: https://www.pivotaltracker.com/story/show/183710218


MKodde (commented):
This feature seems to work as intended! Nice work. also kudos for adding the additional test coverage :+1: 

One thing that might be a nice addition is to also reject access to the `change-request` page when the entity does not have a change request.

You can test this by finding/creating a published production entity, and manually open the route to its change-request page.

I suggest you ask the PM and let him decide if he wants this fixed.

MKodde (approved):
Discussed this with @quartje 

The consensus: it would be best if the user is rejected access physically. But given this is a read only view. No harm can be done. So for now, we will not harden the access restrictions.

Time to :shipit:

PR #569, opened 2022-12-13
Update changes of organization names to manage

Shaky212:
Organization names are registered on the Service
and copied to the metadata of the entity. The copy action 
should not be applied to the entity data
fetched from Manage when used for comparison.

see: https://www.pivotaltracker.com/story/show/184014629


MKodde (changes requested):
Two things came up while reviewing this that not yet seem to work correctly

1. When editing a published production entity. The change request does not have the updated organization names
2. When editing a published test entity. The Organization name is not updated correctly
3. I think you should comment in code what the `getManageEntityById` vs `getPristineManageEntityById` is intended for. That will soon raise eyebrows in the future when we forgot about this bugfix.

MKodde:
Thanks for the updates! This fixed the issue for the change requests and the test entities :+1:

MKodde (approved):

PR #573, opened 2023-01-12
Validate that at least one ACS location is present

Shaky212:
For a SAML entity at least one ACS location is required.

The validation is placed only on the entity. Validation on the form by setting 
the required option of the CollectionType has the side effect that the collection input 
always needs a value even though locations have been entered.

see https://www.pivotaltracker.com/story/show/184195917


MKodde (commented):
This fix seems to do the trick.

Given this feature yielded a bug, I would prefer adding a test that verifies this rule is enforced correctly. Can you try to add this test? A unit test would be preferable, but maybe adding a web test is the easiest solution here

MKodde (approved):
As discussed during stand up:

1. Upgrading to a real browser web test driver (Panther) is not possible in SF3.4. But it will be in the new 5.0 release that is about to happen.
2. For now we skip these tests and create a Pivotal story to reinstate them in the near future. Either by using Panther. Or by rewriting them in Cypress.

https://www.pivotaltracker.com/story/show/184235479

PR #575, opened 2023-01-24
Repair button change request

Shaky212:
For new production entities or template based entities the submit
button label is 'Publish' and 'Change' otherwise.

see: https://www.pivotaltracker.com/story/show/184302283

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
@@ -34,6 +34,8 @@ use function in_array;
  * entity is used in a specific context. This particularly applies for the factory
  * methods found in the 'Entity/Entity' namespace.
  * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ * @SuppressWarnings(PHPMD.ExcessiveClassComplexity) This object contains a lot of properties which cannot not be easily
+ * simplified.
  */
 class ManageEntity
 {
@@ -349,5 +351,16 @@ class ManageEntity
         }
         $data += $this->protocol->asArray();
         return $data;
+    }
+
+    private function isStatusPublicationRequested()
+    {
+        return $this->getStatus() === Constants::STATE_PUBLICATION_REQUESTED;
+    }
+
+    public function isRequestedProductionEntity(
+        bool $isCopy
+    ) {
+        return $isCopy || ($this->isStatusPublicationRequested() && $this->isProduction());
     }
 }
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
@@ -368,6 +368,8 @@ entity.edit.label.implicit: Implicit
 entity.edit.label.authorization_code_token: response_type=id_token
 entity.edit.label.implicit_id_token_token: response_type=id_token token
 entity.edit.label.implicit_id_token: response_type=id_token
+entity.edit.label.change: Change
+entity.add.label.publish: Publish
 entity.acl.title: Entity Idp access
 entity.acl.info.title: Info
 entity.acl.info.html: "
--- a/tests/unit/Infrastructure/DashboardBundle/Factory/EntityTypeFactoryTest.php
+++ b/tests/unit/Infrastructure/DashboardBundle/Factory/EntityTypeFactoryTest.php
@@ -25,6 +25,7 @@ use Surfnet\ServiceProviderDashboard\Application\Command\Entity\SaveOidcngResour
 use Surfnet\ServiceProviderDashboard\Application\Command\Entity\SaveSamlEntityCommand;
 use Surfnet\ServiceProviderDashboard\Application\Service\AttributeServiceInterface;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\ManageEntity;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Factory\EntityTypeFactory;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Form\Entity\OidcngEntityType;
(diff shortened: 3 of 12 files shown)


MKodde (changes requested):
1. The test logic is not correct yet. When opening a non-published prod entity, I see the wrong button text. I slacked a suggestion to fix this.
2. I'd like to see this new logic tested in the `EntityTypeFactoryTest`. It should be possible to record and test this new logic rule there.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -87,6 +89,7 @@ class ManageEntity
      * @return ManageEntity
      *
      * @SuppressWarnings(PHPMD.ElseExpression)
+     * @SuppressWarnings(PHPMD.ElseExpression)
```
MKodde, line 92: Is this an unintentional copy paste?

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -350,4 +353,18 @@ public function asArray(): array
         $data += $this->protocol->asArray();
         return $data;
     }
+
+    public function isStatusPublicationRequested()
+    {
+        return $this->getStatus() === Constants::STATE_PUBLICATION_REQUESTED;
+    }
```
MKodde, line 360: This method is only used within this class and in the unit test. That means that it (c|sh)ould be private. In the test you could simply get the status and verify if it equals `Constants::STATE_PUBLICATION_REQUESTED`

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/translations/messages.en.yml
```diff
@@ -368,6 +368,8 @@ entity.edit.label.implicit: Implicit
 entity.edit.label.authorization_code_token: response_type=id_token
 entity.edit.label.implicit_id_token_token: response_type=id_token token
 entity.edit.label.implicit_id_token: response_type=id_token
+entity.edit.label.change: Change
+entity.add.label.publish: Publish
```
MKodde, line 372: :1st_place_medal:

tests/unit/Infrastructure/DashboardBundle/Factory/EntityTypeFactoryTest.php
```diff
@@ -225,4 +226,140 @@ public function test_build_create_new_oauth20_rs_form()
 
         $this->assertInstanceOf(FormType::class, $form);
     }
+
+    /**
+     *  @dataProvider testOptions
```
MKodde, line 231: Nitpick

Please leave as is, but next time try to use a somewhat more descriptive name, like: `providePublishOrChangeOptions` your provider also describes what it does, but imagine this class having multiple tests with similarly named providers. Then it would soon be hard to distinguish between them.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -350,4 +353,18 @@ public function asArray(): array
...
+
+    public function isRequestedProductionEntity(
+        bool $isCopy
+    ) {
+        if ($isCopy || ($this->isStatusPublicationRequested() && $this->isProduction())) {
+            return true;
+        }
+        return false;
```
MKodde, line 368: You can turn this into a one-line return statement

tests/unit/Infrastructure/DashboardBundle/Factory/EntityTypeFactoryTest.php
```diff
@@ -225,4 +226,140 @@ public function test_build_create_new_oauth20_rs_form()
...
+        $isRequestedProductionEntity,
+        $manageEntityStatus,
+        $environment,
+        $expectation
+    ) {
+        $manageEntity = m::mock(ManageEntity::class);
+        $manageEntity
+            ->shouldReceive('isRequestedProductionEntity')
+            ->with($isCopy)
+            ->andReturn($isRequestedProductionEntity);
```
MKodde, line 244: Might be out of scope for this test, but you could consider making the `ManageEntity` a partial mock. You could make the `isRequestedProductionEntity` method the only 'real' method  that is executed (and is not mocked). That way you can make your data provider simpler. As you no longer have to provide the `isRequestedProducitonEntity` and `manageEntityStatus`.

If you'd like me to demonstrate this feature, feel free to hook up IRL.

tests/unit/Infrastructure/DashboardBundle/Factory/EntityTypeFactoryTest.php
```diff
@@ -225,4 +226,140 @@ public function test_build_create_new_oauth20_rs_form()
...
+                ]
+            ],
+            [
+                false,
+                false,
+                Constants::STATE_PUBLISHED,
+                Constants::ENVIRONMENT_TEST,
+                [
+                ]
+            ],
```
MKodde, line 362: Wouldn't this result in some kind of exception? 
Is this a relevant test avenue?

tests/unit/Infrastructure/Manage/Dto/ManageEntityTest.php
```diff
@@ -102,4 +103,29 @@ public function test_diff_oidc()
...
+    {
+        $entity = ManageEntity::fromApiResponse(json_decode(file_get_contents(__DIR__ . '/fixture/saml20_sp_requested.json'), true));
+        $this->assertTrue($entity->isRequestedProductionEntity(true));
+    }
+
+    public function test_is_requested_production_entity_no_copy()
+    {
+        $entity = ManageEntity::fromApiResponse(json_decode(file_get_contents(__DIR__ . '/fixture/saml20_sp_requested.json'), true));
+        $this->assertFalse($entity->isRequestedProductionEntity(false));
+    }
```
MKodde, line 130: Consider merging these tests into the test_is_not_status_publication_requested that way you can save up on a couple of clock cycles in building the manage entities from json. You might have to rename the test name though.

MKodde (changes requested):
Nice work! See some minor, and a couple of more pressing points of feedback.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
Shaky212: Yes, will remove it

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
Shaky212: Only used here so private is a better option

MKodde (approved):
Nice work! When merging, dont forget to also merge back these changes to `develop`, `release/5.0`. If you'd like we can do this together as an excercise.

PR #577, opened 2023-01-27
Move status logic to manage entity

Shaky212:
To logic to set the status (requested/published) of a manage entity based on the 'exclude_from_push' is moved from the entity to the aggregated manage entity. This logic applies to all entities and has therefore been moved.

see: https://www.pivotaltracker.com/story/show/184313228


src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -132,6 +132,15 @@ public function getMetaData(): ?MetaData
         return $this->metaData;
     }
 
+    public function updateStatusByExcludeFromPush(): void
+    {
+        if ($this->isExcludedFromPush()) {
+            $this->updateStatus(Constants::STATE_PUBLICATION_REQUESTED);
+            return;
+        }
+        $this->updateStatus(Constants::STATE_PUBLISHED);
```
MKodde, line 141: Behavior-wise your change is slightly different from the original. And in this case, the original (wonky looking) behavior is actually intentional.

The coin value for exclude from push can have three different values. Either 1, 0 or null. Where null means it is not set. And 1, that is should be excluded, and 0 that it should not.

When the coin value is `null`, the status should not be updated.

tests/unit/Domain/Entity/Entity/ManageEntityTest.php
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+
+namespace Surfnet\ServiceProviderDashboard\Tests\Unit\Domain\Entity\Entity;
+
+use PHPUnit\Framework\TestCase;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\ManageEntity;
+
+class ManageEntityTest extends TestCase
```
MKodde, line 25: The ManageEntityTest already exists: `sp-dashboard/tests/unit/Infrastructure/Manage/Dto/ManageEntityTest.php`

Please add your test coverage there. If you want a separate test file for your test coverage, that is OK too. But in that case please use the correct namespace & test folder.

tests/unit/Domain/Entity/Entity/ManageEntityTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $manageEntity->updateStatusByExcludeFromPush();
+
+        static::assertEquals('requested', $manageEntity->getStatus());
+        static::assertTrue($manageEntity->isExcludedFromPushSet());
+
+        $manageEntity->updateStatus(Constants::STATE_PUBLISHED);
+
+        static::assertEquals('published', $manageEntity->getStatus());
+        static::assertTrue($manageEntity->isExcludedFromPushSet());
```
MKodde, line 57: Is this correct? When the entity is published, the exclude from push flag should be '0' (when a production entity). A test entity would probably never have the exclude from push flag.

MKodde (commented):
Good idea to move this logic into the domain :medal_military: 

See some questions and remarks below.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -134,7 +134,25 @@ public function getMetaData(): ?MetaData
...
+    }
+
+    public function updateStatusToRemovalRequested()
+    {
+        $this->status = Constants::STATE_REMOVAL_REQUESTED;
+    }
```
MKodde, line 153: Super, thanks for moving business logic to the place it belongs.

tests/unit/Infrastructure/Manage/Dto/ManageEntityTest.php
```diff
@@ -104,17 +104,84 @@ public function test_diff_oidc()
...
-    }
-
     public function test_is_requested_production_entity_copy()
     {
         $entity = ManageEntity::fromApiResponse(json_decode(file_get_contents(__DIR__ . '/fixture/saml20_sp_requested.json'), true));
         $this->assertTrue($entity->isRequestedProductionEntity(true));
         $this->assertFalse($entity->isRequestedProductionEntity(false));
     }
+
+    public function test_it_can_has_status_published_when_excluded_from_push()
```
MKodde, line 114: Nitpick: `it can has`

Reminds me of the good old lolcat days [icanhazcheeseburger](https://icanhas.cheezburger.com/)

tests/unit/Infrastructure/Manage/Dto/ManageEntityTest.php
```diff
@@ -104,17 +104,84 @@ public function test_diff_oidc()
...
+    {
+        $data = json_decode(
+            file_get_contents(__DIR__.'/fixture/manage_entity_saml20_excluded_from_push.json'),
+            true
+        );
+        $manageEntity = ManageEntity::fromApiResponse($data);
+        static::assertTrue($manageEntity->isExcludedFromPushSet());
+
+        $manageEntity->updateStatusToRemovalRequested();
+        static::assertEquals('removal requested', $manageEntity->getStatus());
```
MKodde, line 169: Nitpick: Also goes for the other status codes used throughout these tests, you can use the class constants de-notating these values here. The one benefit of doing this, if/when the status value changes, you only have to update the `const` value.

MKodde (approved):

PR #582, opened 2023-02-02
Throw exception on oauth_ccc when ccc is disabled

Shaky212:
An authorization check has been added to prevent creation of an oAuth client credential clients (ccc) when the option to add client credential clients is disabled.

see: https://www.pivotaltracker.com/story/show/183828026

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
@@ -159,6 +159,12 @@ class EntityCreateController extends AbstractController
             );
         }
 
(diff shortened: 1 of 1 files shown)


MKodde:
Word of warning, if you intend to target this for release 4.4, you might want to rebase from the 4.4 release branch and 'front'port this to the SF5 develop branch. 

If we do not need to backport this, then all is fine!

MKodde (commented):
While performing the functional review, the following things caught my eye.

I started of with a Service that has the CCC feature enabled. I created a CCC entity, and stored all the action links I could think of (view, edit, delete, secret reset). Next I disabled the CCC capability for the entity. But aside from creating a new entity I was able to perform all other tasks. And furthermore, the CCC entities were still visible in the list. 

This feedback might be out of scope here. But I feel we should at least ask @quartje and @teunfransen on what the expected behavior should be.

MKodde (approved):
Thanks for adding the access restriction. I would have loved to see a web test where this scenario is verified. But given the time constraint we might be ok with skipping that feature.

PR #583, opened 2023-02-03
Add default text for revision notes

Shaky212:
Manage requires that revision notes are set and not null. When no comments are set a default text is added to the revision note to meet this requirement.

see: https://www.pivotaltracker.com/story/show/184380521


.github/workflows/test-integration.yml
```diff
@@ -1,9 +1,6 @@
 name: Run integration tests
 
-on:
-  pull_request:
-  push:
-    branches: [ develop ]
+on: [ pull_request, push ]
```
MKodde, line 3: :ok_hand:

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -35,6 +35,10 @@
  */
 class JsonGenerator implements GeneratorInterface
 {
+    public const DEFAULT_CREATED_REVISION_NOTE = 'Entity created';
+    public const DEFAULT_CHANGED_REVISION_NOTE = 'Entity changed';
+    public const DEFAULT_CHANGE_REQUEST_NOTE = 'Change request';
```
MKodde, line 40: I'm fine with a hard coded text, but maybe the PM's expect them to be translatable? Personally I always try to make them translatable, and revert them to hard coded when explicitly required.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -101,6 +105,8 @@ public function generateEntityChangeRequest(
 
         if ($entity->hasComments()) {
             $payload['note'] = $entity->getComments();
+        } else {
+            $payload['note'] = self::DEFAULT_CHANGE_REQUEST_NOTE;
```
MKodde, line 109: Try to prevent `else` expressions at all costs. I think you can work around them here quite easily.

One thing to consider, this might be good candidate to extract the logic into a method?

MKodde (changes requested):
I hate to reject this PR. Especially given the outcome is spot on. The one thing missing from this PR are the tests. I assume you can quite easily cover the new business rules in the JsonGeneratorTest suites.

Also, I feel we should continue making work of moving the business logic away from the application layer. In this instance for example. The notes are set using the data from the form, but when the comment is empty we set a pre-configured text. That can be considered a business rule. And those should be implemented in the domain layer. 

Maybe you can see if it's not too big of a pita to fix it. Otherwise I'm fine with leaving it like this.

Great work so far :+1:

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
Shaky212: Business logic will be moved to ManageEntity.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -18,6 +18,8 @@
 
 namespace Surfnet\ServiceProviderDashboard\Domain\Entity;
 
+use SebastianBergmann\Type\UnknownType;
```
MKodde, line 21: This dependency should not be used here in the domain! It seems to be a dev-leftover. So no man overboard

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -18,6 +18,8 @@
...
+use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator;
```
MKodde, line 22: Rule of thumb, no application or infra dependencies should be present in the domain layer. This also looks like a dev-leftover as I see no further references to JsonGenerator.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/ManageEntity.php
```diff
@@ -26,6 +28,7 @@
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\OidcngClient;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\OidcngResourceServerClient;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Entity\Protocol;
+use Surfnet\ServiceProviderDashboard\Domain\Exception\UnknownTypeException;
```
MKodde, line 31: Could this also be a dev-leftover? My IDE marks it as an undefined class.

tests/unit/Infrastructure/Manage/Dto/ManageEntityTest.php
```diff
@@ -117,4 +123,32 @@ public function test_is_requested_production_entity_copy()
...
+    public function test_it_has_entity_changed_revision_notes_for_a_requested_state()
+    {
+        $entity = ManageEntity::fromApiResponse(json_decode(file_get_contents(__DIR__ . '/fixture/saml20_changed_entity.json'), true));
+        $entity->setStatus(Constants::STATE_PUBLICATION_REQUESTED);
+        $this->assertSame('Entity changed', $entity->getRevisionNote());
+    }
+
+    public function test_it_has_changed_request_revision_notes_for_a_publisched_state()
+    {
+        $entity = ManageEntity::fromApiResponse(json_decode(file_get_contents(__DIR__ . '/fixture/saml20_changed_request.json'), true));
```
MKodde, line 151: The fixture name is very confusing. As the fixture results in a published production entity.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/OauthClientCredentialsClientJsonGenerator.php
```diff
@@ -21,6 +21,7 @@
 use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\PrivacyQuestionsMetadataGenerator;
 use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\SpDashboardMetadataGenerator;
 use Surfnet\ServiceProviderDashboard\Application\Parser\OidcngClientIdParser;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
```
MKodde, line 24: No longer used?

src/Surfnet/ServiceProviderDashboard/Application/Metadata/OidcngJsonGenerator.php
```diff
@@ -18,10 +18,12 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\Metadata;
 
+use PHPUnit\TextUI\XmlConfiguration\Constant;
 use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\ArpGenerator;
 use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\PrivacyQuestionsMetadataGenerator;
 use Surfnet\ServiceProviderDashboard\Application\Metadata\JsonGenerator\SpDashboardMetadataGenerator;
 use Surfnet\ServiceProviderDashboard\Application\Parser\OidcngClientIdParser;
+use Surfnet\ServiceProviderDashboard\Domain\Entity\Constants;
```
MKodde, line 26: Unused in the generator

src/Surfnet/ServiceProviderDashboard/Application/Metadata/OidcngResourceServerJsonGenerator.php
```diff
@@ -18,6 +18,7 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\Metadata;
 
+use PHPUnit\Util\Json;
```
MKodde, line 21: Not used import

MKodde (changes requested):
Thanks for this superb change! I love it when the prodigal son (application code) returns home to the welcoming father (the domain layer). Joking aside, here are some questions you might need to address:

1. Mustn't you call the getReveisonNotes method here? `\Surfnet\ServiceProviderDashboard\Domain\Entity\data`
2. Is the `getComments` method still relevant? And must we not make it a private method?
3. What about test entities? They are omitted from test, but I recon they should also display a correct revision note, no?

MKodde (approved):
Thanks for the final changes. Love what this PR ended up being!

You might want to experiment with doing some branch commit history rewriting. Or squash merge this PR to get a clean history.

PR #584, opened 2023-02-15
Show secrets modal for new oidc entities after connection request

Shaky212:
When I  an OIDC entity is published to production, the modal with the secret and ClientID has to appear after submittance or skipping the connection request form. 

see: https://www.pivotaltracker.com/story/show/184477504

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
@@ -236,12 +236,12 @@ trait EntityControllerTrait
     private function findDestinationForRedirectToCreateConnectionRequest(
         PublishProductionCommandInterface $publishEntityCommand
     ): string {
-    
         switch (true) {
             case $publishEntityCommand instanceof EntityChangeRequestCommand:
                 return 'entity_change_request';
             case $publishEntityCommand instanceof PublishEntityProductionCommand:
-                if ($this->allowToRedirectToCreateConnectionRequest($publishEntityCommand)) {
+                if ($publishEntityCommand->getManageEntity()->getStatus() !== Constants::STATE_PUBLICATION_REQUESTED &&
+                    $this->allowToRedirectToCreateConnectionRequest($publishEntityCommand)) {
                     return 'entity_published_create_connection_request';
                 }
                 return 'entity_published_production';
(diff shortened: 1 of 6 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
```diff
@@ -234,14 +234,14 @@ private function findParametersForRedirectToCreateConnectionRequest(
...
-        PublishProductionCommandInterface $publishEntityCommand
-    ): string {
+        PublishProductionCommandInterface $publishEntityCommand): string {
     
         switch (true) {
             case $publishEntityCommand instanceof EntityChangeRequestCommand:
                 return 'entity_change_request';
             case $publishEntityCommand instanceof PublishEntityProductionCommand:
-                if ($this->allowToRedirectToCreateConnectionRequest($publishEntityCommand)) {
+                if ($publishEntityCommand->getManageEntity()->getStatus() &&
```
MKodde, line 243: I forgot to actually verify the 'requested' status is set.

MKodde (changes requested):
Hi, as we already discussed IRL. Your proposal triggered a tension with me. And it turned out the entity status was updated incorrectly. I added a commit (including a bug) to roll back your change and add a proof of concept on how to correctly go about this status check.

Secondly: you already told me this logic was hard to test. But I feel it is important to do have test coverage of on this feature.

My suggestion is to add a pivotal story to address this issue. I would really like to have a test matrix to aid us in writing those tests correct. This matrix should tell us which screens/application state is expected on the (many) different create/edit/change request flows on the different environments and publication statuses.

| Protocol | Environment | Is published | Is excluded from push | Expected entity status | Should show connection request? | Should show client secret | Should display thank you page | Should display flash message |
|-----|-----|-----|-----|-----|-----|-----|-----|-----|
|saml20|prod|yes|yes|requested|yes|no|yes|no|
|oidc10|prod|yes|yes|requested|yes|yes|yes|no|

This matrix could be used to create a web test that (using a data provider) test all expected behavior. This woyuld help enormously in preventing future regressions.

Would you be so kind to create a technical debt pivotal story for creating this test?

MKodde (approved):
This should do the trick. 

1. Thanks for the TD ticket in pivotal. I updated the story with some details, pointers, thoughts. feel free to add to that.
2. I still would have liked to see some test coverage in this PR, but given the very impractical nature of realizing this. I can live with this for now. Especially given we posted the TD ticket.

PR #585, opened 2023-02-16
Display secrets after adding a connection request for an OIDC entity

Shaky212:
The secrets of an OIDC were only displayed when the dialog of the connection request was skipped. This PR fixes this issue that the secrets are also shown when an connection request is posted.

see: https://www.pivotaltracker.com/story/show/184460472


MKodde (approved):
I approve of this. Disclaimer, I have not functionally tested this. So proceed with caution. :warning:

PR #588, opened 2023-06-20
Remove unused privacy questions

MKodde:
PART 1 of the Privacy Questions overhaul

In this part all previously required, but no longer relevant privacy questions are removed from SPD

They are:
1. Dropped from database
2. Removed from code
3. Removed from tests

As a bonus, the default config was updated to ensure a correct database connection is created when using the .env.dist file. And the doctrine_migratoins config was updated to more closely represent the pre SF5 situation.

See: https://www.pivotaltracker.com/story/show/185441606


config/packages/doctrine_migrations.yaml
```diff
@@ -5,7 +5,9 @@ doctrine_migrations:
     storage:
             table_storage:
                 table_name: 'migration_versions'
-                version_column_name: 'SP_Dashboard_migrations'
+                version_column_name: 'version'
```
pablothedude, line 8: This change won't conflict in production?

pablothedude (approved):

PR #589, opened 2023-06-20
Update security scanner policy

MKodde:
QA builds should not break and halt the development process.

1. Security scanning is performed periodically via weekly sceduled IRL checks
2. By utilizing the security scanner from the OpenConext .github actions templates
3. By relying on the dependabot

### DIFF
--- a/.github/workflows/daily-security-check.yml
+++ b/.github/workflows/daily-security-check.yml
@@ -1,0 +1,84 @@
+---
+name: Daily security check
+on:
+  schedule:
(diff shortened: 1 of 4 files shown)


quartje:
Excellent idea to keep it out of the other pipeline. You mention weekly in the PR, but the task is performed daily. How often do we do this in other projects?

quartje (approved):

MKodde:
The weekly check is the one we perform IRL during refinement/sprint planning meetings.

PR #591, opened 2023-06-22
Introduce new privacy questions

MKodde:
PART 2 of the Privacy Questions overhaul

In the second part, I added the new fields. Which are the `dpaType` and the multilingual `PrivacyStatementUrl` (nl+en) fields. They are added throughout the entire stack.

While at it, I found out the Privacy Questions where not included upon entity update actions. That was also addressed in this PR.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/PrivacyQuestions/PrivacyQuestionsCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/PrivacyQuestions/PrivacyQuestionsCommand.php
@@ -18,10 +18,11 @@
 
 namespace Surfnet\ServiceProviderDashboard\Application\Command\PrivacyQuestions;
 
-use DateTime;
 use Surfnet\ServiceProviderDashboard\Application\Command\Command;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\PrivacyQuestions;
 use Surfnet\ServiceProviderDashboard\Domain\Entity\Service;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\DpaType;
+use Symfony\Component\Validator\Constraints as Assert;
 
 /**
  * @SuppressWarnings(PHPMD.TooManyFields)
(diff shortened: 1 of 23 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/DpaType.php
```diff
@@ -0,0 +1,101 @@
...
+ *      "dpa_supplied_by_service",
+ *      "other"
+ *    ],
+ *    "default" : "service_has_own_dpa",
+ *    "info": "Determines what DPA this service has to offer"
+ *  }
+ */
+class DpaType
+{
+    public const DPA_TYPE_NOT_APPLICABLE = 'dpa_not_applicable';
```
pablothedude, line 49: Do all those constant have to be public?

src/Surfnet/ServiceProviderDashboard/Application/Command/PrivacyQuestions/PrivacyQuestionsCommand.php
```diff
@@ -68,6 +69,18 @@ class PrivacyQuestionsCommand implements Command
      */
     private $otherInfo;
 
+    private DpaType $dpaType;
```
pablothedude, line 72: No assertion is needed here?

pablothedude (approved):

src/Surfnet/ServiceProviderDashboard/Domain/ValueObject/DpaType.php
MKodde: Good point, only the `default` was used outside of the VO scope. 

Updated

src/Surfnet/ServiceProviderDashboard/Application/Command/PrivacyQuestions/PrivacyQuestionsCommand.php
MKodde: A required Assertion should be set here. Thanks for spotting that!

PR #592, opened 2023-06-26
Repair jira_enable_test_mode feature flag

MKodde:
The jira_enable_test_mode parameter is read. By default, the setting is set to be enabled. This aids dev + test environments.


quartje:
I've tested this on our test JIRA. Everything works as expected. Merge & Tag!

PR #594, opened 2023-06-27
Push test metadata upon entity deletion

MKodde:
When a test entity is removed from manage, also push the updated metadata to EngineBlock and the OidcNg gateway.

`https://www.pivotaltracker.com/story/show/184915015`


rubencatshoek (approved):
LGTM

PR #595, opened 2023-07-06
When no privacy Q&A set, do not try to read them

MKodde:
The JSON generator has been made more robust, only adding the questions (and associated answers) when they are present on the Service entity.

PR #602, opened 2023-07-20
Allow edits of the Collection widget entries

MKodde:
By removing the readonly state when pressing the edit button, the value can now be modified. The collection widget was modified to allow for this behavior. Three fields currenlty utilize this widget (acs locations, redirect urls and the connection requests) The ACS and Redirect urls are affected by this change, as I thought it would be a good idea to start small. Editing the connection requests may come at a later stage.

See: https://www.pivotaltracker.com/story/show/185431458

Screenshots:

Screenshot from 2023-07-20 10-26-22
Screenshot from 2023-07-20 10-23-06

Disclaimer on the broken build:
- The integration test was fixed in #595 
- The failing webtest is a false positive

### DIFF
--- a/assets/js/components/collection_widget.test.ts
+++ b/assets/js/components/collection_widget.test.ts
@@ -40,7 +40,7 @@ describe('validate collection widget', function() {
     $('.add_collection_entry').click();
 
     let expected = `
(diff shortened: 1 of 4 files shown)


MKodde:
Functionally this PR was reviewed by @quartje code review was not formally done, but I checked the code for a second time before merging.

PR #603, opened 2023-07-20
Service switcher improvements

MKodde:
1. The empty option was removed from the service switcher (was caused by a missing placeholder text)
2. Given you are admin and only one service is activated, the service overview of that service is now auto-loaded.

See: https://www.pivotaltracker.com/story/show/182154117

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/ServiceService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/ServiceService.php
@@ -58,7 +58,7 @@ class ServiceService
      *
      * @param array $allowedServices The input should be service names keyed by service id.
      *                               As provided by: AuthorizationService::getAllowedServiceNamesById
-     * @return array
+     * @return Service[]
      */
     public function getServicesByAllowedServices(array $allowedServices)
     {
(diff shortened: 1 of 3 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/ServiceService.php
```diff
@@ -58,7 +58,7 @@ public function getServiceNamesById()
      *
      * @param array $allowedServices The input should be service names keyed by service id.
      *                               As provided by: AuthorizationService::getAllowedServiceNamesById
-     * @return array
+     * @return Service[]
```
MKodde, line 61: :fallen_leaf: Boyscout rule

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceController.php
```diff
@@ -83,8 +84,17 @@ public function overviewAction()
...
             return $this->redirectToRoute('service_add');
         }
+        // If only one service is available and you are admin, select that one
+        if ($this->isGranted('ROLE_ADMINISTRATOR') && count($services) === 1) {
+            $service = reset($services);
+            $this->authorizationService->changeActiveService($service->getId());
+            return $this->redirect(
+                $this->generateUrl('service_admin_overview', ['serviceId' => $service->getId()])
+            );
+        }
```
MKodde, line 94: Scenario when only one service is present in the SP dashboard
And you log in as an `admin` user
Then you should be auto-forwarded to the service overview page of that one service

Previously you would land on the 'select a service' admin welcome page. However, the service switcher would not be displayed as only one service is present.

johanib (approved):

PR #605, opened 2023-08-15
Prevent double actions

MKodde:
When clicking certain buttons in the teams screens, events would be triggered multiple times. This was already fixed in #600. But other construct also used the mousedown event. That was fixed in this commit for the change member the resent invitation listeners. The issue was described in the storie mentioned below.

See: https://www.pivotaltracker.com/story/show/185785974
See: https://github.com/SURFnet/sp-dashboard/pull/600/files

### DIFF
--- a/assets/js/components/resend_invite.ts
+++ b/assets/js/components/resend_invite.ts
@@ -7,7 +7,8 @@ $(() => {
 
   if (!!flashUl && !!resendInviteLinks) {
     resendInviteLinks.forEach(item => {
-      item.addEventListener('mousedown', (event) => {
+      item.addEventListener('click', (event) => {
+        event.preventDefault();
         event.stopImmediatePropagation();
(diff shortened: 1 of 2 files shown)


assets/js/components/resend_invite.ts
```diff
@@ -7,7 +7,8 @@ $(() => {
 
   if (!!flashUl && !!resendInviteLinks) {
     resendInviteLinks.forEach(item => {
-      item.addEventListener('mousedown', (event) => {
+      item.addEventListener('click', (event) => {
+        event.preventDefault();
         event.stopImmediatePropagation();
```
johanib, line 12: Isnt it strange that this solves the issue?

Mousedown should only be invoked once on each click?

johanib: discussed, approved

PR #608, opened 2023-08-17
Add DPA choice specific tooltips

MKodde:
Each DPA type now has a dedicated tooltip as described in:

https://www.pivotaltracker.com/story/show/185441616

The hints are rendered inside the choices. 
The base form template needed an adjustment to allow the hint to show up. 
The hints rely on the existing tippy tooltip implementation. So no additional JS was required to add event listeners.

### DIFF
--- a/assets/scss/pages/privacy_questions.scss
+++ b/assets/scss/pages/privacy_questions.scss
@@ -1,0 +1,7 @@
+.dpa-types {
+  .help-button {
+    position: initial;
+    margin-left: auto;
+    padding: 8px 4px 0;
+  }
+}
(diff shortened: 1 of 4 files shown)


assets/scss/pages/privacy_questions.scss
```diff
@@ -0,0 +1,7 @@
+.dpa-types {
+  .help-button {
+    position: revert;
```
johanib, line 3: oof, postition revert sounds complicated. Did you come up with this, or is it a copy paste?
Why not use initial or the position you want? Now you are making it dependent on the parent.

Just wondering

johanib (approved):

PR #76, opened 2017-11-03
Publish production entity

MKodde:
**Review notes**
A database migration was added.

**Todo**
 - [x] Integration test the CommandHandler
 - [x] Implement feature to edit published production form
 - [x] Add extra information to service mail message 

**Checklist**
- [x] Tests are written
- [x] Travis build is passing
- [x] Git history is clean
- [x] Issue in Pivotal is updated

PR #612, opened 2023-10-04
Upgrade JS CPD upgrading blamer in the process

MKodde:
Hardening against vulnerability described here: 

blamer vulnerable to Arbitrary Argument Injection via the blameByFile() API #76


quartje (approved):
All tests seem to pass, so let's merge & roll

PR #614, opened 2023-10-10
Fix the mixup of DPA_TYPE_MODEL_SURF and DPA_TYPE_IN_SURF_AGREEMENT

quartje:
We have fixed this in production by changing the translations. Better to fix this in code


MKodde (commented):
When looking at the key-value pairs they seem to make sense. By changing this up like you propose I'm losing this association. 

This might need to be changed at another point in the code? Or maybe in translations?

Of maybe, most probably, I'm missing the point?

parijke (approved):

quartje:
> When looking at the key-value pairs they seem to make sense. By changing this up like you propose I'm losing this association.
> 
> This might need to be changed at another point in the code? Or maybe in translations?
> 
> Of maybe, most probably, I'm missing the point?

You are completely right of course. That'll teach me trying to change code :)

It was the default translation that was mixed up. I've reworked the PR to change the translations in stead!

PR #621, opened 2024-02-07
Add token based authentication

parijke:


.env.dist
```diff
@@ -101,9 +101,7 @@ mail_no_reply=no-reply@surfconext.nl
 jira_test_mode_storage_path='../var/issues.json'
 
 # Jira settings
-JIRA_HOST='https://jira.example.com'
-JIRA_USER=sp-dashboard
-JIRA_PASS=secret
+# The basic authetication settings are supplied via env.j2
```
MKodde, line 104: Thanks for adding this nugget of wisdom. For consistency sake, I'd like to see a default being set for the `jira_personal_access_token`

It's okay for it to be a bogus `set your own token` text as a token value. But having a default value that is automatically set when provisioning the .env file is a philosophy we always cling to.

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Factory/JiraServiceFactory.php
```diff
@@ -24,26 +24,22 @@
...
             'jiraHost' => $host,
-            'jiraUser' => $username,
-            'jiraPassword' => $password
+            'useTokenBasedAuth' => true,
+            'personalAccessToken' => $personalAccessToken,
         ]);
     }
 
-    public function buildIssueService()
+    public function buildIssueService(): IssueService
```
MKodde, line 42: :heart:

MKodde (commented):
Havn't given this a functional test spin. But it seems to fit the bill. See my one suggestion which I think should be addressed.

parijke (commented):
See discussion

.env.dist
MKodde: Our policy is: if the app relies on configuration, either from parameters.yaml or .env vars. We should:
- ensure the presence of said params in a `dist` config example
- describe the how and why of those config params. 
- and if possible validate the existence and correct type of those config files.

Im totally fine with deploying a key via a deploy script! But we should still have a dummy placeholder in the .dist file. Or, maybe better. Add a validation in the app to verify that env-var is present. If not. The app should warn the user that an important config option is missing. Usually I do this by throwing a config exception resulting in a 500 error. 

Given we are dealing with env vars here. This exception wil probably be thrown during runtime?!

MKodde (changes requested):

.env.dist
parijke: > Our policy is: if the app relies on configuration, either from parameters.yaml or .env vars.

In this case, none of the above. The tests are green btw. For me it was consfusing that I added a value to the .dist file, but it wasn't picked up. This is why I explicitely added the comment that the ENV value is set by an external process.

I have no problem to add defaults back, let the client decide.

parijke (commented):
Changed according aggreements

quartje:
Functional testing has been done, it works as proposed

quartje (approved):

MKodde (approved):

PR #628, opened 2024-05-14
Remove 'production entities enabled' feature toggle from the Service entity

MKodde:
The 'production entities enabled' feature toggle is not used by the PO's of SPD. They always allow the services to create production entites. Having this featue is illogical, and it can be cleaned up.

Note that a Doctrine migration was created to remove the column from the `service` table.

https://www.pivotaltracker.com/story/show/187053888


src/Surfnet/ServiceProviderDashboard/Application/Command/Service/EditServiceCommand.php
```diff
@@ -51,7 +50,7 @@ public function __construct(
         private ?string $intakeStatus,
         private ?string $contractSigned,
         private ?string $surfconextRepresentativeApproved,
-        private ?string $privacyQuestionsAnswered,
+        private bool $privacyQuestionsAnswered,
```
parijke, line 53: not nullable anymore?

BTW, I heard a statement, that bool variables "sh|c"ould be named with `is` in front, like $isPrivacyQuestionAnswered.

Butt hat is probably not for now

parijke: If is is not nullebale, maybe initilize it with false to be sure?

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
```diff
@@ -21,12 +21,9 @@
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
-use Surfnet\ServiceProviderDashboard\Domain\Entity\PrivacyQuestions;
```
parijke, line 24: Not sure why this is gone, was it obsolete already?

parijke (approved):
Small suggestions / questions, no blockers to merge IMHO

src/Surfnet/ServiceProviderDashboard/Application/Command/Service/EditServiceCommand.php
MKodde: I looked it up, and it should always be set. My own testing and the CI functional tests did confirm my assumption. I feel confident to leave it like this.

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
MKodde: That was a useless import statement, the PrivacyQuestions are in the same namespace. So importing them is not usefull.

MKodde (commented):
Thanks for your review @parijke

src/Surfnet/ServiceProviderDashboard/Application/Command/Service/EditServiceCommand.php
parijke: https://3v4l.org/gWhSF

PR #629, opened 2024-05-15
Use devconf as basis for development.

quartje:
We used a custom OpenConext container before. Since we use the [devconf](https://github.com/OpenConext/OpenConext-devconf) project for all docker related development, SPD will use that as well. Some work needs yet to be done to rewrite the integration tests to use the same environment.

you will need to checkout devconf in the same dir as  the sp-dashboard code 

codedir:
  - sp-dashboard/
  - OpenConext-devconf/


MKodde (approved):
Tested this. Works like a charm!

PR #636, opened 2024-06-12
Add `createdAt` to Service entity

MKodde:
This additional field will show us when a service is created. It is not displayed anywhere in the UI (by design). It is just there for the TPMs to inform them for how long the SP has been granted access to the dashboard.

See: https://www.pivotaltracker.com/story/show/187715470

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
@@ -18,9 +18,11 @@
 
 namespace Surfnet\ServiceProviderDashboard\Domain\Entity;
 
+use DateTime;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
```diff
@@ -63,6 +65,13 @@ class Service
     #[ORM\GeneratedValue(strategy: 'AUTO')]
     private $id;
 
+    /**
+     * @var DateTime $created
+     */
+    #[Gedmo\Timestampable(on: 'create')]
+    #[ORM\Column(type: 'datetime')]
+    private DateTime $created;
```
parijke, line 73: Isn't it better to rename this to `createdAt`

parijke (changes requested):

src/Surfnet/ServiceProviderDashboard/Domain/Entity/Service.php
MKodde: It's a better name, thanks for the suggestion.

parijke (approved):

PR #642, opened 2024-06-18
Remove filter column and add the extra explanation text

MKodde:
1. The filter column is removed from the template as it is out of scope for now.
2. An additional translatable explanation text was added 
3. Webtests have been reinstated

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
@@ -49,6 +49,12 @@ class DevelopmentIssueRepository implements TicketServiceInterface
         $this->failIssueCreation = true;
         $this->storeData();
     }
+    public function shouldNotFailCreateIssue(): void
+    {
+        $this->loadData();
+        $this->failIssueCreation = false;
+        $this->storeData();
+    }
 
     public function findByManageIds(array $manageIds): IssueCollection
     {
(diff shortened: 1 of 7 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
```diff
@@ -49,6 +49,12 @@ public function shouldFailCreateIssue(): void
         $this->failIssueCreation = true;
         $this->storeData();
     }
+    public function shouldNotFailCreateIssue(): void
```
parijke, line 52: This is a weird method name. No method should ever fail, but they do... not clear to me

parijke (changes requested):
Maybe consider renaming the weird method name

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
MKodde: Haha, this a method that modifies the behaviour of a fixture. You can enable it to simulate Jira being down.

parijke: Why not call it enableJiraDownSimulation() then?

parijke (approved):
Why not call it enableJiraDownSimulation() then?

src/Surfnet/ServiceProviderDashboard/Infrastructure/Jira/Repository/DevelopmentIssueRepository.php
MKodde: I find this more fitting in functional tests, but my pardigm is mainly gherkin oriented.

PR #643, opened 2024-06-18
Rename responsible -> representative

MKodde:
That is a more fitting name for the concept it represents


src/Surfnet/ServiceProviderDashboard/Application/Service/ServiceConnectionService.php
```diff
@@ -177,4 +179,42 @@ private function assertValidMetadata(ManageEntity $entity): MetaData
...
+            if (array_key_exists($identityProvider, $testIdpsIndexed)) {
+                $connectedIdps[$identityProvider] = $testIdpsIndexed[$identityProvider];
+            }
+        }
+        return $connectedIdps;
+    }
+
+    private function getTeam(MetaData $metadata): string
+    {
+        $team = $metadata->getCoin()->getServiceTeamId();
```
parijke, line 213: Not that it is a biggy, but the method you call returns an Id (I think) than it would be clearer to use $teamId and even an opportunity to use a VO here. No blocker

src/Surfnet/ServiceProviderDashboard/Application/ViewObject/EntityConnection.php
```diff
@@ -47,13 +50,22 @@ public function listConnected(): array
         return $list;
     }
 
+    public function hasConnectedOtherIdp(): bool
```
parijke, line 53: I love these simple methods, immediatelly clear to me. As Martin Fowler said, "any fool can write code that machines understand, but only good programmers write code that humans understand"

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceConnectionsController.php
```diff
@@ -27,25 +27,50 @@
...
 
-    #[IsGranted(new Expression('is_granted("ROLE_ADMINISTRATOR") or is_granted("ROLE_SURFCONEXT_RESPONSIBLE")'))]
+    #[IsGranted(new Expression('is_granted("ROLE_ADMINISTRATOR") or is_granted("ROLE_SURFCONEXT_REPRESENTATIVE")'))]
     #[Route(
         path: '/connections',
         name: 'service_connections',
         methods: ['GET', 'POST']
     )]
-    public function myServices()
+    public function __invoke(): RedirectResponse|Response
```
parijke, line 51: I love __invoke <3

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceConnectionsController.php
```diff
@@ -27,25 +27,50 @@
...
+            '@Dashboard/Service/my-services.html.twig',
+            [
+                'testIdps' => $entities->getTestIdps(),
+                'entities' => $entities,
+            ]
+        );
+    }
+
+    #[IsGranted("ROLE_ADMINISTRATOR")]
+    #[Route(
```
parijke, line 68: Owww... when I see __invoke, I always assume there's only one route in the controller. I even wrote a rector rule that if a controller has __invoke, it only can have one public method (__invoke)

Consider splitting into two seperate controllers

tests/webtests/CreateConnectionRequestTest.php
```diff
@@ -38,7 +39,7 @@ public function setUp(): void
             'SURF SP2',
             'https://sp2-surf.com',
             'https://sp2-surf.com/metadata',
-            'urn:collab:group:vm.openconext.org:demo:openconext:org:surf.nl'
+            WebTestFixtures::TEAMNAME_SURF
```
parijke, line 42: Yes!

parijke (approved):
Small recommendations, no blockers

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/ServiceConnectionsController.php
MKodde: I knew you would come up with this feedback. And you are right of course! I also think that this use case is quite valid (two views on the same data, with a different response format). It seems fitting to have 1 controller with these exact dependencies to serve both purposes.

PR #646, opened 2024-06-25
Introduce the 'Type of service' options to SP Dashboard

MKodde:
This PR allows us to configure the types in a json blob (format is open for suggestions)
This data is then displayed on the SAML, OIDC and CCC forms
The data is not yet shipped to Manage

Possible improvements: 

[] - json schema validation?
[] - Store the data in Manage
[] - Read the data from Manage and introduce it on the ManageEntity in the (Coin) value object


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -106,6 +107,19 @@ class SaveSamlEntityCommand implements SaveEntityCommandInterface
...
+    ])]
+    #[Assert\Count(
+        min: 1,
+        max: 3,
+        minMessage: 'validator.type-of-service.min',
+        maxMessage: 'validator.type-of-service.max',
+    )]
+    private array $typeOfService;
+
     #[Assert\Url]
```
parijke, line 123: Do I see some repetition here? DRY?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
```diff
@@ -235,6 +235,13 @@ services:
   Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Repository\AttributeRepository:
     arguments: [ '%kernel.project_dir%/assets/attributes.json' ]
 
+  Surfnet\ServiceProviderDashboard\Domain\Repository\TypeOfServiceRepository:
+    alias: Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Repository\TypeOfServiceRepositoryFromConfig
+
```
parijke, line 240: Why the alias? Can't we use autowire?

parijke (changes requested):

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
MKodde: You are looking at a (long) form with several elements that have similar characteristics. They however represent entities of their own. I'm considering making a custom type of service validator. Suitable for the 3 entity types.

PR #647, opened 2024-06-27
Update logout Url and allow slashes in Service name (in url path)

parijke:
- set the correct logout url 
- fix that a slash can be used in a service name on route level


parijke:
@quartje can you please test the fix? I think this bug will also appear on delete and updates, so if this solves creation we can use it on the other routes as well

.env.dev
```diff
@@ -1,4 +1,4 @@
-APP_ENV=dev
+APP_ENV=prod
```
MKodde, line 1: This might need to be reverted back to `dev`.

.env.dev
```diff
@@ -11,9 +11,9 @@ mailer_password=null
 secret=ThisTokenIsNotSoSecretChangeIt
 session_max_absolute_lifetime=3600
 session_max_relative_lifetime=600
-logout_redirect_url='https=//www.surf.nl/over-surf/werkmaatschappijen/surfnet'
+logout_redirect_url='https://www.surf.nl/over-surf/werkmaatschappijen/surfnet'
 # All users in these teams get the administrator role
-administrator_teams="'urn:collab:group:dev.openconext.local:dev:openconext:local:spd_admin'"
+administrator_teams="'urn:collab:group:dev.openconext.local:dev:openconext:local:spd_admin','urn:collab:org:dev.openconext.local'"
```
MKodde, line 16: I removed the 'default' team membership to prevent you are always admin when logging in. If this not breaks the tests, its OK to leave it like this! I can always change it if I want to. (just feedback to explain to you why the team is not there by default)

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
```diff
@@ -63,6 +63,7 @@ public function __construct(
     #[Route(
         path: '/entity/create/type/{serviceId}/{targetEnvironment}/{inputId}',
         name: 'entity_type',
+        requirements: ['inputId' => '.+'],
```
MKodde, line 66: Perhaps link to the why (in comment)? Either link to the story here, or maybe to the SF docs?

`https://symfony.com/doc/6.4/routing.html#slash-characters-in-route-parameters`

MKodde (changes requested):
Nice job on fixing the issues! I see that it is a work in progress. See some feedback below to hit it home next time you work on it.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityCreateController.php
parijke: I think it is in the commit message

.env.dev
parijke: @quartje this is about your commit

MKodde (approved):

MKodde (approved):

PR #649, opened 2024-07-03
Type of service UI improvements

MKodde:
1. The graphical visualisation has been upgraded to match that of the other form components
2. Show the Type of service on the entity detail page


MKodde (commented):
Disclaimer: Contents of the PR already landed in main via the parent branch. 

The remaining change on this PR is now only the removal of an imported function. And the setting of the typeOfService on the command, and I later found out that this is a faulty call, as the CCC entity does not have a typeOfService.

Closing this PR!

PR #651, opened 2024-07-04
Add properties to the ARP values

parijke:
A number of fields will be added in the arp. Besides source, value and motivation, use_as_name_id and "release_as" are added. If these already exist, they should be returned with the same values (analogous to "source" which we also leave untouched)

Before: 

```json
 "arp" : {
    "attributes" : {
      "urn:mace:dir:attribute-def:givenName" : [ {
        "source" : "idp",
        "value" : "*",
        "motivation" : "1243"
      } ],
      "urn:mace:dir:attribute-def:uid" : [ {
        "source" : "idp",
        "value" : "*",
        "motivation" : "test"
      } ]
}
```

After:
```json
 "arp" : {
    "attributes" : {
      "urn:mace:dir:attribute-def:givenName" : [ {
        "source" : "idp",
        "value" : "*",
        "motivation" : "1243",
        "release_as" : "123",
        "use_as_name_id" : true
      } ],
      "urn:mace:dir:attribute-def:uid" : [ {
        "source" : "idp",
        "value" : "*",
        "motivation" : "test"
        "release_as" : "123",
        "use_as_name_id" : true
      } ]
}
```

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
@@ -158,11 +158,7 @@ class EntityMergeService
     {
         $attributeList = new AttributeList();
 
-        // Oidc TNG resource servers do not track attributes in manage
-        // Neither do the Oauth Client Credentials clients
-        if ($command instanceof SaveOidcngResourceServerEntityCommand
-            || $command instanceof SaveOauthClientCredentialClientCommand
-        ) {
+        if ($this->isNotTrackedInManage($command)) {
             return $attributeList;
         }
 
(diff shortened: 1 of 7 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Service/EntityMergeService.php
```diff
@@ -158,11 +158,7 @@ private function buildAttributesFromCommand(SaveEntityCommandInterface $command)
     {
         $attributeList = new AttributeList();
 
-        // Oidc TNG resource servers do not track attributes in manage
-        // Neither do the Oauth Client Credentials clients
-        if ($command instanceof SaveOidcngResourceServerEntityCommand
-            || $command instanceof SaveOauthClientCredentialClientCommand
-        ) {
+        if ($this->isNotTrackedInManage($command)) {
```
MKodde, line 161: :evergreen_tree:

MKodde (approved):
Code changes look great, I rebased the code onto main. And added some addition test coverage to the `AttributeListTest`

PR #655, opened 2024-07-10
Use Symfony UX component to improve TOS selection

MKodde:
The Type of Service checkbox group did not fully meet the requirements of the PM's. They initially asked for a select box to choose the types of service.

This SF UX component adds that functionality to the app. This is the first  Symfony UX component for SPD. Its CSS is updated quite severely. The component styling does not yet sit quite well with me. But it works, it just does not adhere to our CSS philosophy... I can :see_no_evil: that for now..


src/Surfnet/ServiceProviderDashboard/Application/Service/ServiceConnectionService.php
```diff
@@ -119,12 +119,12 @@ private function addEntitiesToCollection(
...
             $adminContact = $this->composeContactString($entity, 'admin');
-            if ($service === null || !in_array($entity->getProtocol()->getProtocol(), $allowedProtocols)) {
-                // Skipping entities of which we do not know the service (team is not set on any of our Services)
+            if (!in_array($entity->getProtocol()->getProtocol(), $allowedProtocols)) {
+                // Skipping entities that are not saml or oidc rp
                 continue;
             }
             if ($entity->getAllowedIdentityProviders()->isAllowAll()) {
-                $serviceName = $service->getOrganizationNameEn();
+                $serviceName = $service?->getOrganizationNameEn() ?: 'Unknown';
```
MKodde, line 127: This change is unrelated to this PR. In short this loosens the requirement for the service to be present on the ManageEntity. If the service is not known (the entity is not tracked by SP Dashboard) the entity may still be displayed on the SURFconext representative page. @quartje requested me to categorize these entities under the service 'Unknown' section.

I call upon the boyscout rule here.

assets/scss/components/auto_complete.scss
parijke, line 1: That is a lot of css. But I read that you already aggree and willing to leave it for now

parijke (approved):
No blockers, few questions

assets/scss/components/auto_complete.scss
MKodde: Yup, normally Syfmony UX would load the CSS automatically. But that kite did not fly in our project. Due to the way we use SASS/SCSS in conjunction with Encore.

Second, much of the basic styling needed to be adjusted. And TOM is not super modern in the CSS department. So this is kind of a monster. But nicely isolated in a single CSS file.

PR #657, opened 2024-07-17
Allow both ROLE_USER&SURFCONEXT_REPRESENTATIVE

MKodde:
At the same time. This entails rearanging the SamlProvider slightly.

Tests where made to reflect the combinations between the roles

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
@@ -126,22 +126,31 @@ class SamlProvider implements SamlProviderInterface, UserProviderInterface
             $authorizations = null;
         }
         // Default to the ROLE_USER role for services.
-        $role = 'ROLE_USER';
         if ($authorizations && $authorizations->isSurfConextRepresentative()) {
-            $role = 'ROLE_SURFCONEXT_REPRESENTATIVE';
             $contact->setInstitutionId($authorizations->getOrganizationCode());
-            $contact->assignRole($role);
+            $contact->assignRole('ROLE_SURFCONEXT_REPRESENTATIVE');
         }
 
         if (array_intersect($this->administratorTeams, $teamNames) !== []) {
-            $role = 'ROLE_ADMINISTRATOR';
-            $contact->assignRole($role);
+            $contact->assignRole('ROLE_ADMINISTRATOR');
+            return;
         }
 
-        if ($role === 'ROLE_USER') {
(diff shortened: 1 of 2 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProvider.php
```diff
@@ -126,22 +126,31 @@ private function assignRoles(
             $authorizations = null;
         }
         // Default to the ROLE_USER role for services.
-        $role = 'ROLE_USER';
         if ($authorizations && $authorizations->isSurfConextRepresentative()) {
-            $role = 'ROLE_SURFCONEXT_REPRESENTATIVE';
             $contact->setInstitutionId($authorizations->getOrganizationCode());
-            $contact->assignRole($role);
+            $contact->assignRole('ROLE_SURFCONEXT_REPRESENTATIVE');
```
parijke, line 131: Can we use information msnor Const for roles?

tests/unit/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProviderTest.php
```diff
@@ -85,14 +89,85 @@ public function test_administrator_teams_validation_rejects_invalid_teams()
...
+            ]
+        );
+        $service1 = m::mock(Service::class);
+        $service2 = m::mock(Service::class);
+        $this->serviceRepo->shouldReceive('findByTeamNames')->with(['team-1', 'team-2'])->andReturn([$service1, $service2]);
+        $this->contactRepo->shouldReceive('findByNameId')->with('does-not-exist-yet')->andReturnNull();
+        $this->serviceRepo->shouldReceive('findByTeamNames')->andReturnNull();
+        $this->contactRepo->shouldReceive('save');
+        $user = $provider->getUser($assertion);
+        self::assertEquals($user->getRoles(), ['ROLE_SURFCONEXT_REPRESENTATIVE', 'ROLE_USER']);
```
parijke, line 160: What happens if the user has an extra rikenin the future? Then this test would fail I guess.

Isn't it better to check if the roles intersect?

parijke (approved):
No blockers, some suggestions

tests/unit/Infrastructure/DashboardSamlBundle/Security/Authentication/Provider/SamlProviderTest.php
MKodde: Im totally fine with having to update (and specify) this assertion if anything in authz changes in the future. That only shows that the behavior changes. And allows us to verify if the existing workings are not broken by the new change.

PR #664, opened 2024-08-06
New field to set 'coin:ss:idp_visible_only' 

MKodde:
The SAML and OIDC entities now allow for setting the coin:ss:idp_visible_only metadata setting. See the story for the fine details.

See: https://www.pivotaltracker.com/story/show/187515050

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOauthClientCredentialClientCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOauthClientCredentialClientCommand.php
@@ -543,4 +543,10 @@ class SaveOauthClientCredentialClientCommand implements SaveEntityCommandInterfa
         // not implemented for oauth_ccc entities
         return [];
     }
+
+    public function isPublicInDashboard(): ?bool
+    {
+        // This coin is not used for CCC entities
+        return null;
+    }
 }
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
@@ -153,6 +153,11 @@ class SaveSamlEntityCommand implements SaveEntityCommandInterface
     #[Assert\Choice(callback: [Constants::class, 'getValidNameIdFormats'], strict: true)]
     private ?string $nameIdFormat = Constants::NAME_ID_FORMAT_TRANSIENT;
 
+    /**
+     * When checked on the form, the idpVisibleOnly value becomes false
+     */
+    public bool $isPublicOnDashboard;
+
     private ?string $manageId = null;
     private ?Attribute $organizationUnitAttribute = null;
 
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
@@ -156,7 +156,7 @@ class SaveSamlEntityCommand implements SaveEntityCommandInterface
     /**
      * When checked on the form, the idpVisibleOnly value becomes false
      */
-    public bool $isPublicOnDashboard;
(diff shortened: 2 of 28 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOauthClientCredentialClientCommand.php
```diff
@@ -542,4 +542,10 @@ public function getTypeOfService(): array
         // not implemented for oauth_ccc entities
         return [];
     }
+
+    public function isPublicInDashboard(): ?bool
+    {
+        // This coin is not used for CCC entities
+        return null;
```
parijke, line 549: Why not return false?

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -487,4 +492,14 @@ public function setTypeOfService(array $typeOfServices): void
...
+
+    public function isPublicInDashboard(): ?bool
+    {
+        return $this->isPublicInDashboard;
+    }
+
+    public function setIsPublicInDashboard(bool $isPublic): void
+    {
+        $this->isPublicInDashboard = $isPublic;
+    }
```
parijke, line 504: Same, why not reutn false. Also, I see some repetition. Don;t know if that is okay, because code duplication is way cheaper than using the wrong abstractie - Sandy Metz

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
```diff
@@ -227,7 +227,9 @@ private function generateMetadataFields(ManageEntity $entity): array
         if ($entity->getMetaData()->getLogo() instanceof Logo && $entity->getMetaData()->getLogo()->getUrl() !== '') {
             $metadata = array_merge($metadata, $this->generateLogoMetadata($entity));
         }
-
+        if ($entity->getMetaData()?->getCoin()->isIdpVisibleOnly() !== null) {
```
parijke, line 230: What does this mean? Tis is a fairly complex equation, and it is not clear what is meant. Using a descriptive method could maybe clarify a bit

templates/EntityDetail/detailBooleanField.html.twig
```diff
@@ -1,4 +1,4 @@
-{% if value is not empty %}
+{% if value == true or value == false %}
```
parijke, line 1: This is weird

parijke (commented):
Some suggestions

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOauthClientCredentialClientCommand.php
MKodde: When a coin is not set in Manage, we consider it to be `null` in SP Dashbaord.
When a boolean coin is present (like `coin:ss:idp_visible_only`) the other values can be true or false.

When a certain entity type does not support a certain coin. I feel returning `null` by default works out the best.

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
MKodde: What part of the equation is complex in your opinion? I find it rather expressive, but I'm kind of biassed ;P

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
MKodde: See previous comment

src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php
parijke: The whole part... I have no idea what true and false mean in this case

parijke (approved):

PR #667, opened 2024-08-06
Remove certificate data from SAML entities

MKodde:
Certificate data is no longer visible on SPD, if the value is set in Manage, SPD will not overwrite it.

In addition to Form and Command logic, some validators and other helper utils have been removed

See: https://www.pivotaltracker.com/story/show/186341612


parijke (approved):
Delete is the best refactoring tool :-)

PR #668, opened 2024-08-08
Hide nameIdFormat when `unspecified`

MKodde:
When the 'subjectType' or 'nameIdFormat' is unspecified. The matching form field is not displayed on the form. That way we do not allow the user to overwrite this value in Manage (which is explicitly set for that SP by a product owner)

To test this, the mock SP storage needed to track the nameIdFormat value as a manage entity. That requires quite some test code to be edited..

See: https://www.pivotaltracker.com/story/show/185952423

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/SamlEntityType.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/SamlEntityType.php
@@ -31,6 +31,8 @@ use Symfony\Component\Form\Extension\Core\Type\SubmitType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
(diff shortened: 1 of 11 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/SamlEntityType.php
```diff
@@ -323,6 +325,18 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+            /** @var SaveSamlEntityCommand $data */
+            $data = $event->getData();
+            if ($data->getNameIdFormat() === Constants::NAME_ID_FORMAT_UNSPECIFIED) {
+                $form = $event->getForm();
+                if ($form->has('metadata') && $form->get('metadata')->has('nameIdFormat')) {
+                    $form->get('metadata')->remove('nameIdFormat');
+                }
+            }
+        });
     }
```
parijke, line 340: I see some repetition

tests/webtests/EntityEditTest.php
```diff
@@ -81,6 +82,33 @@ public function test_it_renders_the_form()
...
+            'saml20_sp',
+            '88888888-1111-1111-1111-888888888888',
+            'SP1',
+            'https://spx-entityid.example.com',
+            'https://spx-entityid.example.com/metadata',
+            WebTestFixtures::TEAMNAME_SURF,
+            '12',
+            Constants::NAME_ID_FORMAT_UNSPECIFIED
+        );
+        $crawler = self::$pantherClient->request('GET', "/entity/edit/test/88888888-1111-1111-1111-888888888888/1");
```
parijke, line 105: This does not have to be with a full blown panther instance I guess, could also be done with lightweigth kernelbrowser

parijke (approved):
Excellent job again!, just a few questions/ suggestions

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Form/Entity/SamlEntityType.php
MKodde: It's not realy a dry thing in my opinion; the one block deals with the nameIdFormat and the other with the clientId. They kind-of are the same thing. Leaving it as is for now

PR #669, opened 2024-08-12
Add detailed revision note when editing an entity

MKodde:
See: https://www.pivotaltracker.com/story/show/187992345


parijke (approved):
Code wise ok. The only  thing is a lot of repetition, but that is a know issue in these generator classes

PR #670, opened 2024-08-21
Unblue the `/connections` name column values

MKodde:
The names of the services preivously were blue. Implying you can click on them. But they are no anchor or otherwise interactable. So having them grey makes more sense.

See: https://www.pivotaltracker.com/story/show/186881796/comments/242257354


MKodde:
Bart functionally reviewed this. Given the simple nature of this PR.. Merging..

PR #671, opened 2024-08-21
Show correct menu items according to the granted authorizations

MKodde:
As of 17th of July it was possible to both be SurfConextResponsible AND have a regular ROLE_USER at the same time. This was already correctly handled by the Authz logic, but the menu did not yet show the correct menu items for the roles the user is authorized for..

That's what fixed here. At first I thought the SamlProvider was broken, hence the (cosmetic) change in the Test.

See: https://www.pivotaltracker.com/story/show/186881796/comments/242257534


MKodde:
Bart functionally reviewed this. Given the simple nature of this PR.. Merging..

PR #672, opened 2024-08-22
Add all IdPs to the export when ACL allowAll=true

MKodde:
It makes sense to show all connected IdP's in the CSV export for the surfconext representative. When the ACL is set to allow all IdPs connected to the SP, we should list all IdPs of the test env there.

This can be a long list. But now we would only show the Organizations IdPs, and the explicitly marked test-idps.

See: https://www.pivotaltracker.com/story/show/187805216/comments/242257758

### DIFF
--- a/tests/unit/Infrastructure/Manage/Client/fixture/identity_provider_response.json
+++ b/tests/unit/Infrastructure/Manage/Client/fixture/identity_provider_response.json
@@ -42,7 +42,7 @@
     "_id": "0c3febd2-3f67-4b8a-b90d-ce56a3b0abb6",
     "version": 0,
     "data": {
-      "entityid": "https://engine.dev.support.surfconext.nl/authentication/idp/metadata2",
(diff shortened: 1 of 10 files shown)


tests/unit/Infrastructure/Manage/Client/fixture/identity_provider_response.json
```diff
@@ -42,7 +42,7 @@
     "_id": "0c3febd2-3f67-4b8a-b90d-ce56a3b0abb6",
     "version": 0,
     "data": {
-      "entityid": "https://engine.dev.support.surfconext.nl/authentication/idp/metadata2",
+      "entityid": "https://engine.dev.support.surfconext.nl/authentication/idp/metadata3",
```
MKodde, line 45: This would not be possible in Manage, have two entities with the same entityId, fixed that in test

:evergreen_tree:

johanib (approved):
I don't know the context, but it looks OK to me.

MKodde:
@quartje and @teunfransen  will ensure this feature is tested for functional correctness

PR #674, opened 2024-08-26
Optimize WebTests

MKodde:
The WebTests fail regularly, this is an effort to stabilize them.


MKodde:
My hope that the increase in PHP Mem limit would positively effect the Webtets. But this does not seem to be the issue.

PR #675, opened 2024-08-26
Test workflows: Bootstrap first, then start spdashboard

quartje:


MKodde (changes requested):
Might be ahead of the game.. But reviewing your changes (which look great by the way) I noticed that the new strategy to run the tests as 33:33 (www-data) is causing the app to not be able to write to the readonly the database

MKodde (approved):

MKodde:
@quartje I removed the change to the .env.test and added some fixes for the webtests.

PR #681, opened 2024-10-24
Fix webtests in dev env.

johanib:
Prior to this change, the webtests would fail when ran in the dev env, unless the `team_prefix_default_stem_name` env was adjusted manually.

This change updates the .dist env and several webtest fixtures to use the updated stem.

See: https://www.pivotaltracker.com/n/projects/1400064/stories/179572218/comments/227653860

Also scouted: Fix wrong manage credentials in `.env.dev`
Scouted: Add missing `yarn encore dev` to Readme

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DataFixtures/ORM/WebTestFixtures.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DataFixtures/ORM/WebTestFixtures.php
@@ -27,8 +27,8 @@ use Symfony\Component\Uid\Uuid;
 
 class WebTestFixtures extends Fixture
 {
-    public const TEAMNAME_SURF = 'urn:collab:group:vm.openconext.org:demo:openconext:org:surf.nl';
-    public const TEAMNAME_IBUILDINGS = 'urn:collab:group:vm.openconext.org:demo:openconext:org:ibuildings.nl';
+    public const TEAMNAME_SURF = 'urn:collab:group:dev.openconext.local:demo:openconext:org:surf.nl';
+    public const TEAMNAME_IBUILDINGS = 'urn:collab:group:dev.openconext.local:demo:openconext:org:ibuildings.nl';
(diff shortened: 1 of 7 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/DataFixtures/ORM/WebTestFixtures.php
```diff
@@ -27,8 +27,8 @@
 
 class WebTestFixtures extends Fixture
 {
-    public const TEAMNAME_SURF = 'urn:collab:group:vm.openconext.org:demo:openconext:org:surf.nl';
-    public const TEAMNAME_IBUILDINGS = 'urn:collab:group:vm.openconext.org:demo:openconext:org:ibuildings.nl';
+    public const TEAMNAME_SURF = 'urn:collab:group:dev.openconext.local:demo:openconext:org:surf.nl';
+    public const TEAMNAME_IBUILDINGS = 'urn:collab:group:dev.openconext.local:demo:openconext:org:ibuildings.nl';
     public const TEAMNAME_ACME = 'demo:openconext:org:acme.com';
```
johanib, line 32: How come sometimes the stem is sometimes added, and other times not?
Even if it's the default stem. If the default changes, it feels like things could break?

Should they not be saved consistently with or without the stem?

MKodde: The stem can change between environments. Thats why we configure them. For test, they should match the one set in the `.env.test` IIRC.

I think we now save the teamname without the stem right?

MKodde: The stem is also configured to later recognize the team/invide role name (urn). It should start with the stem. And then have the teamname proposed by us (based on the service name)

MKodde (approved):

PR #684, opened 2024-10-30
Branch of main with no changes.

johanib:


johanib:
@MKodde 
Based on the symfony csrf docs, I cannot find any more leads what might cause csrf issues in this webtest. 
https://symfony.com/doc/current/security/csrf.html

So I disabled the csrf protection in tests. 
The other solution I can come up with is disabling the offending test.

Let me know what you think..

It might be related to the csrf feature starting the session which might interfere with the pather/webdriver session or something.

tests/webtests/EditServiceTest.php
```diff
@@ -74,8 +75,7 @@ public function test_privacy_questions_admin_toggle()
             ->form();
         $form->setValues($formData);
         self::$pantherClient->executeScript("document.getElementsByClassName('service-form').item(0).submit();");
-        self::$pantherClient->wait(3);
-        self::assertOnPage('Your changes were saved!');
+        self::$pantherClient->wait(10)->until( WebDriverExpectedCondition::elementTextContains(WebDriverBy::cssSelector('body'), 'Your changes were saved!'));
```
MKodde, line 78: Do you need the `wait` in conjunction with `until`? Wouldn't `until` wait as long as need be?

johanib: https://github.com/php-webdriver/php-webdriver/wiki/HowTo-Wait

```
// Wait for at most 10s and retry every 500ms if it the title is not correct.
$driver->wait(10, 500)->until(
  WebDriverExpectedCondition::titleIs('My Page')
);
```

MKodde: TIL!

MKodde (approved):

PR #868, opened 2024-10-30
Update patch versions in composer and yarn packages.

johanib:
Minor bumps:
* OpenConext/Monitor-bundle

Issues:
* https://github.com/symfony/symfony/issues/54739 caused a regression that needed to be addressed.


composer.json
MKodde, line 1: Out of curiosity; is this the result of `composer bump`?

ci/qa/phpstan-baseline.php
```diff
@@ -1016,11 +1016,6 @@
 	'count' => 1,
 	'path' => __DIR__ . '/../../src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php',
 ];
-$ignoreErrors[] = [
-	'message' => '#^Right side of && is always true\\.$#',
-	'count' => 1,
-	'path' => __DIR__ . '/../../src/Surfnet/ServiceProviderDashboard/Application/Metadata/JsonGenerator.php',
-];
```
MKodde, line 1023: Did PHPStan drop this sniff? I see no change to the `JsonGenerator` in this changeset to warrant the removal of these lines. :man_shrugging:

MKodde (approved):
Nice work! I assume the changes to the `src` files where required because of the upgrades to the package constraints? Might be good to summarize in the PR description why you fixed those things? But I'm fine with the way this is as well.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -175,11 +175,26 @@ public function __set(string $property, ?Attribute $value)
...
+     * If __isset would return false, the __get would not be called, and null would not be returned, resulting in an error.
+     * Therefore, __isset will always return true.
+     *
+     * @see getAttribute
+     * @see https://github.com/symfony/symfony/issues/54739
+     */
+    public function __isset(string $name): bool
+    {
+        return true;
+    }
```
MKodde, line 190: Thanks for pinching this!

composer.json
johanib: apart from Monitor-bundle, yes

ci/qa/phpstan-baseline.php
johanib: maybe phpstan got smarter, or an upstream library changed.

The other change, `if (!isset($organizationCodeMatches[1])) {` suddenly was an issue that phpstan wanted fixed.

PR #1292, opened 2024-10-31
When webtest fails, upload a screenshot of the error to github actions.

johanib:
Example output:
https://github.com/SURFnet/sp-dashboard/actions/runs/11608163793


MKodde (approved):
Super useful! :shipit:

PR #1293, opened 2024-10-31
Restore the production_entities_enabled feature

johanib:
Please note the individual commits for context and scope to make reviewing easier.

Restores feature deleted in https://github.com/SURFnet/sp-dashboard/pull/628
See https://www.pivotaltracker.com/n/projects/1400064/stories/187053888


MKodde (approved):
From code review perspective this looks great! I did not have the opportunity to functionally test this feature. Will try that now

johanib:
Discussed with Michiel. He could not test because his env would not start.
I tested the migration on my machine, and both scenario's work. 
The button also works.

Merging.

Issue #1305, opened 2024-11-07
Further decouple team urn from teams


johanib:
In issue https://github.com/SURFnet/sp-dashboard/issues/1167 the team name creation is delegated to Invite. So team names are no longer created by sp-dashboard.

This has some consequences that will be adressed in this ticket:
* sp-dashboard should no longer show the team name input on the service created screen
* When a service is updated in sp-dashboard, it should no longer call Teams to check if the team name exists
* When a service is updated in sp-dashboard, it should make sure the team name is unique in the sp-dashboard service table.
* Because the team name will no longer be validated in teams, any team name urn manipulation (stem, group/context) should be removed, as both Invite and sp-dashboard will use the full team name.
* Update 11-11-24: Discussed and agreed to remove teamname urn validation on service edit, as it does not make sense anymore. urns are created by invite, and can potentially not match the validator.

Afaics the sp-dashboard service table always contains the full team urn.

PR #1306, opened 2024-11-07
Further decouple team urn from teams

johanib:
Creating a service no longer generates a teamname urn

Prior to this change, a teamname urn would be created by sp-dashboard.
This feature has been delegated to Invite.
This change removes the code needed to create a new urn.

See https://github.com/SURFnet/sp-dashboard/issues/1167
See https://github.com/SURFnet/sp-dashboard/issues/1305


MKodde (approved):
Awesome, this demystifies the service logic severely!

PR #1308, opened 2024-11-08
Feature/add release gha

quartje:
Add a GitHub action to do automatic releases
Updated the Changelog to reflect all changes. 
Tag 6.1 never lead to a real release, so we decided to create on big 6.2 release.


MKodde (approved):
LGTM!

PR #1311, opened 2024-11-12
Address some deprecations

johanib:
Prior to this change, the ServiceOverviewTest would load the fixtures twice.
This would make doctrine complain about primary keys being forcefully reused on new enities.
This change does not load the fixtures twice, resolving the issue.

Prior to this change, doctrine would note that we need to make a decision whether to use savepoints or not.
This change explicitly enables savepoints, as Innodb engine and sqlite support savepoints.
This was always the default behavior, and by explicitly setting it to 
true, we keep it that way. Thil will tell doctrine that it can use 
nested database transactions.

Also rewrite test assertion to not rely on deprecated Hamcrest bundle.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -1,6 +1,7 @@
 doctrine:
     dbal:
         url: '%env(resolve:DATABASE_URL)%'
+        use_savepoints: true
 
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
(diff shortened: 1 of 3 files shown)


config/packages/doctrine.yaml
```diff
@@ -1,6 +1,7 @@
 doctrine:
     dbal:
         url: '%env(resolve:DATABASE_URL)%'
+        use_savepoints: true
```
MKodde, line 4: You mention we enable this as our db engine supports it. But I do not see what change of behavior this brings? 

Might be worthwhile to describe this is the default behavior that is made explicit here? And that it ensures nested transactions are handled in a certain manner?

MKodde (approved):

config/packages/doctrine.yaml
johanib: Thanks, I clarified the commit message!

PR #1312, opened 2024-11-12
Attempt to remove StaleElementReferenceException errors from pipeline

johanib:
Prior to this change, the webtests would randomly fail with
StaleElementReferenceException errors.
This change attempts to work around those errors by adding a retry.
Most of the time, the error is thrown on the consent check.


MKodde:
I remember trying several things to get this test to pass on every build. If this fixes the issue, I'm all for it!

MKodde (approved):

PR #1317, opened 2024-11-25
Improve UX of Create connection request form

johanib:
### Improve UX of Create connection request form
Prior to this change, when you enterd the form data, it was unclear you needed to press the + button in order to 'add' it. This change shows the list headers + a clarifying text to clarify this. The clarifying text should only be shown if there are no items in the list.

Resolves https://github.com/SURFnet/sp-dashboard/issues/1201

---

### Prevent 'Create connection request' broken form state
Prior to this change, when you would click '+' on an invalid row, then click 'Send', the row would become disabled while being invalid, effectively freezing the form, requiring a manual page refresh to continue.
This change will not disable the row unless the row is valid.

Old
https://github.com/user-attachments/assets/d98522a9-609f-45c7-90d1-cb72424de4d0

New
https://github.com/user-attachments/assets/3874871f-7c99-410c-a164-612d03a17617

### DIFF
--- a/templates/form/fields.html.twig
+++ b/templates/form/fields.html.twig
@@ -162,6 +162,9 @@
                     <th class="text">Email</th>
                     <th>&nbsp</th>
                 </tr>
+                <tr class="empty-connection-list-message">
+                    <td colspan="4">{{ 'entity.create_connection_request.emptyConnectionListPlaceholder'|trans }}</td>
+                </tr>
             </table>
         {% else %}
         <ul class="collection-list">
--- a/templates/form/fields.html.twig
+++ b/templates/form/fields.html.twig
@@ -160,7 +160,7 @@
                     <th class="text">Institution</th>
                     <th class="text">Name</th>
                     <th class="text">Email</th>
-                    <th>&nbsp</th>
+                    <th>&nbsp;</th>
                 </tr>
                 <tr class="empty-connection-list-message">
                     <td colspan="4">{{ 'entity.create_connection_request.emptyConnectionListPlaceholder'|trans }}</td>
--- a/tests/webtests/CreateConnectionRequestTest.php
+++ b/tests/webtests/CreateConnectionRequestTest.php
@@ -20,6 +20,7 @@ namespace Surfnet\ServiceProviderDashboard\Webtests;
 
 use Facebook\WebDriver\WebDriverBy;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\DataFixtures\ORM\WebTestFixtures;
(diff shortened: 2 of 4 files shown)


templates/form/fields.html.twig
```diff
@@ -162,6 +162,9 @@
                     <th class="text">Email</th>
                     <th>&nbsp</th>
```
MKodde, line 163: #Nitpick
#Outside context of your PR

Needs a closing `;` ?

tests/webtests/CreateConnectionRequestTest.php
```diff
@@ -61,15 +61,17 @@ public function test_a_connection_request_can_be_created()
         self::assertSelectorIsDisabled('#connection_request_container_send');
         self::assertSelectorIsEnabled('#connection_request_container_cancel');
 
+        self::assertSelectorTextContains('.empty-connection-list-message', 'No items have been added yet. Add items to this list using the + button, then click the Send button.');
```
MKodde, line 64: Suggestion: inject the translator in this test, and simply check if `__('entity.create_connection_request.emptyConnectionListPlaceholder')` is on the page. That way, when the translation changes, you do not have to update you test.

MKodde (changes requested):
IRL we found a UX issue in the form where you could lock the connection request form. Aside from that I found these two things that could be addressed (or not)

tests/webtests/CreateConnectionRequestTest.php
```diff
@@ -61,15 +62,20 @@ public function test_a_connection_request_can_be_created()
...
         $this->fillFormField($form, '#connection_request_container_connectionRequests___name___name', 'Jesse');
         $this->fillFormField($form, '#connection_request_container_connectionRequests___name___email', 'jesse-james@gmail.com');
         $this->click($form, '.add_collection_entry');
 
         self::assertSelectorExists('.collection-list');
         self::assertSelectorIsVisible('.collection-list');
+        self::assertSelectorTextNotContains('.empty-connection-list-message', 'No items have been added yet.');
 
-        self::assertSelectorTextContains('.collection-list td:nth-child(1)', 'Institution 1');
+        self::assertSelectorTextContains('.collection-list tr.collection-entry td:nth-child(1)', 'Institution 1');
```
MKodde, line 78: No option to use the translator here? Feel free to ignore/resolve if you find this feedback insignificant.

MKodde (approved):
Looks great! I trust you now configured the submit + validation + disable logic in a manner the PO agrees with! So you've got yourself an approval. I did have one additional test suggestion, but feel free to continue without addressing that.

Issue #1318, opened 2024-11-25
Improve contrast on form labels


johanib:
The form labels on SP Dashboard are very light gray, causing them to fail the accessibility contrast check.

I propose setting the label color to dark gray. 

Before:
![new_service_before](https://github.com/user-attachments/assets/c7f0ec49-f886-404c-bf3b-c13db91b1881)
After:
![new_service_after](https://github.com/user-attachments/assets/35f89faf-fa80-4749-b024-1631f0f5c94e)
Before:
![registration_form_before](https://github.com/user-attachments/assets/37a3f055-e8e7-4f69-8ad0-00ec31b9762a)
After:
![registration_form_after](https://github.com/user-attachments/assets/a2098723-221d-46fa-9ffc-d0f7c759cc22)

PR #1319, opened 2024-11-25
Improve contrast on form labels

johanib:
Prior to this change, the contrast on form labels was too low.

Resolves https://github.com/SURFnet/sp-dashboard/issues/1318


MKodde (approved):
Nice work

PR #1323, opened 2024-11-27
Ensure correct invite url

johanib:
Prior to this change, the api call to invite would fail because the relative path would collapse. This change ensures the baseUri and call path are absolute/relative when needed so the calls get executed as expected.

See https://symfony.com/doc/7.2/reference/configuration/framework.html#reference-http-client-base-uri

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/Invite/InviteHttpClient.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/Invite/InviteHttpClient.php
@@ -43,8 +43,10 @@ class InviteHttpClient
             ],
         ];
 
+        $baseUri = InviteHttpClientUrlSanitizer::buildBaseUri($path, $host);
+
         $this->httpClient = HttpClient::createForBaseUri(
-            $host . $path,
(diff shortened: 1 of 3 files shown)


tests/unit/Infrastructure/Invite/InviteHttpClientTest.php
MKodde, line 1: Thanks for testing the sanitizer! Should those tests not be in a test file of its own? Otherwise, I suggest moving this test to the integration test namespace, as this on the verge of being an integration test.

MKodde (commented):
Nice optimization and added defence in depth!

See one suggestion for the test, olther than that gtg!

tests/unit/Infrastructure/Invite/InviteHttpClientTest.php
johanib: ah, I forgot to rename the test.

Its a pure unit test right? It only tests the sanitizer

MKodde: Think that threw me off. It indeed is a pure unit test :+1:

Issue #1321, opened 2024-11-27
Temporarily disable "type of service" field


quartje:
In https://github.com/SURFnet/sp-dashboard/pull/646 a new manage field was introduces to indicate the Type of service. The story to request this feature was not clear enough. Another issue will be opened with the rework needed to fix this feature.

In the meantime, we would like to release 6.2 _without_ this particular feature. So creating a feature branch for 6.2, and temporarily disable the field in the oidc and saml entry forms. and leaving it out from publishing to production.

Issue #1322, opened 2024-11-27
Type of service field is emptied when a non existing value is already present in manage


quartje:
In https://github.com/SURFnet/sp-dashboard/pull/646 and https://github.com/SURFnet/sp-dashboard/pull/648 and new Manage field was introduced. While the functionality works, some edge cases do not: There are some types which we have in Manage, but which should not be presented to customers. We can leave it out from the json where we define the types but that gives the following bug:

In manage, have an entity which has a "type" that is not present in the json.
In spdashboard, edit the entity (just have a change, like name)
The result is pushed back to manage, but the type of service is emptied. 

Proposed solution: In the attributes json in the assets, we have the option to define an attribute that exists, but should not be shown in the dashboard using the excludeOnEntityType attribute. Maybe we can use that mechanism here as well?

johanib:
@teunfransen Which of these types should be invisible to the users?

I think you mentioned:
* SURF
* Recommended

Any more?

```
[
  {
    "typeEn" : "SURF",
    "typeNl" : "SURF"
  },
  {
    "typeEn" : "Education",
    "typeNl" : "Onderwijs"
  },
  {
    "typeEn" : "Research",
    "typeNl" : "Onderzoek"
  },
  {
    "typeEn" : "Privacy/security",
    "typeNl" : "Privacy/beveiliging"
  },
  {
    "typeEn" : "Content/library",
    "typeNl" : "Content/bibliotheek"
  },
  {
    "typeEn" : "Business management",
    "typeNl" : "Bedrijfsvoering"
  },
  {
    "typeEn" : "Recommended",
    "typeNl" : "Aangeraden"
  },
  {
    "typeEn" : "Productivity",
    "typeNl" : "Productiviteit"
  },
  {
    "typeEn" : "Management of education/research",
    "typeNl" : "Organisatie van onderwijs/onderzoek"
  },
  {
    "typeEn" : "Collaboration",
    "typeNl" : "Samenwerken"
  },
  {
    "typeEn" : "Medical",
    "typeNl" : "Medisch"
  },
  {
    "typeEn" : "eCommerce",
    "typeNl" : "eCommerce"
  },
  {
    "typeEn" : "Tooling",
    "typeNl" : "Tooling"
  },
  {
    "typeEn" : "Repository",
    "typeNl" : "Repository"
  }
]

```

teunfransen:
Hi Johan, indeed, SURF and Recommended

PR #1324, opened 2024-11-27
Temporarily hide the TypeOfService selector and allow empty values to…

johanib:
… prevent unsubmittable forms

Resolves https://github.com/SURFnet/sp-dashboard/issues/1321

### DIFF
--- a/assets/scss/pages/base.scss
+++ b/assets/scss/pages/base.scss
@@ -63,3 +63,7 @@
 .hidden {
   display: none;
 }
+
+div.form-row:has(#dashboard_bundle_entity_type_metadata_typeOfService) {
+  display: none;
+}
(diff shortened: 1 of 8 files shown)


johanib:
@MKodde As discussed, I tested the other form as wel. There were no errors.

MKodde (approved):
If 'soft' disabling is OK with the PM, this is fine with me. Maybe refer to the ticket where this is going to be fixed in the PR description? Is saw a reference in the TODO comment, might be handy to link it here too?

johanib:
Will be re-enabled in https://github.com/SURFnet/sp-dashboard/issues/1322

MKodde (commented):
We reviewed this IRL; this temporary solution will be rolled back and tweaked in the near future

Issue #1331, opened 2024-12-13
When publishing an entity for test, at least 1 idp must be selected


teunfransen:
When publishing an entity to the TEST environment, the user is presented with an idp selection screen. 
Currently we have no validation on that screen. This generates an error when a user selects no IdP's.

In the new IdP selection (/idps) menu, at least 1 idp must be selected. If no idp has been selected, the validation signals (like the validation messages in the entity creation screens) in red the following message:
```
No connection has been selected. Please provide at least one connection to continue
```

johanib:
Validation added in https://github.com/SURFnet/sp-dashboard/pull/1334

As discussed, the validation should also be implemented for the other screen. 
But that screen is deprecated, and should be replaced:

https://github.com/SURFnet/sp-dashboard/pull/1335

PR #1334, opened 2024-12-17
Add validation: At least 1 idp must be selected

johanib:
Prior to this change, when no idp was selected in the Entity IDP selection when publishing an Entity, the form validation would handle, and and error was thrown.

This change adds form validation that at least one IDP must be selected.

Resolves https://github.com/SURFnet/sp-dashboard/issues/1331

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/AtLeastOneSelected.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/AtLeastOneSelected.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * Copyright 2024 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/AtLeastOneSelected.php
```diff
@@ -0,0 +1,53 @@
...
+{
+
+    /**
+     * @param string[] $fieldNames
+     */
+    #[HasNamedArguments]
+    public function __construct(
+        public array $fieldNames,
+        ?array $groups = null,
+        mixed $payload = null,
```
pablothedude, line 39: Isn't it possible to omit `mixed` here? If you use mixed here then I would also expect a test for different types of variables.

templates/EntityAcl/idps.html.twig
```diff
@@ -15,7 +15,9 @@
             {% endfor %}
         {% endfor %}
 
-        {{ form_errors(form) }}
+        {% for error in form.vars.errors %}
```
pablothedude, line 18: Is this custom code construct really needed?

pablothedude (approved):

templates/EntityAcl/idps.html.twig
johanib: @teunfransen @quartje 
Should this validation also be added this screen?
image

@quartje 
Teun mentioned that unchecking `Select all` is similar to not selecting anything (not sure if I understood that correctly). So it might not cause functional issues? But it might not be good UX.

teunfransen: I've tested this, and it's true

PR #1335, opened 2024-12-17
Remove ACL idp selection screen

johanib:
Prior to this change, the old idp selection screen was still used. This change links the old screen to the new screen.

As discussed with Teun/Bart, the screens should be identical.


johanib:
Discussed with Teun. 
For now, this is to much scopecreep. 

Closing. Will be fixed in https://github.com/SURFnet/sp-dashboard/issues/1336

Issue #1343, opened 2025-01-06
Contractual base is not set when doing a change request


quartje:
In https://github.com/SURFnet/sp-dashboard/pull/665 we've added the contractual base field. This is correctly applied when creating entities. When doing a change request however, the field is updated with the value "null". 
The expected behaviour is:
* When the field is not present, and it with the correct value
* When the field is present and the value is incorrect update it
* When the field is present and the value is correct, leave it untouched

johanib:
@quartje @teunfransen 
I'm having trouble reproducing this. Does this occur for a specific entity type? Or a specific state?
I did some step debugging and looked at the code with Michiel, but what seems to happen is:

When updating (publishing in general), the contractual base is set based on the service type. If the contractual base is not different from the value currently in manage, it will not update the field in manage.


https://github.com/user-attachments/assets/f065a341-7260-4a44-b037-69a6f9bc1a32

johanib:
I managed to reproduce using a oidcng test entity.

# In manage, manually add the contractual base.
# In sp-dashboard, update the entity
# The contractual base in manage will be gone

johanib:
This was an issue because only specific coin metadata gets updated by sp-dashboard. 
In this case 'contractualBase' was in the list of fields because it is used for production entities. 

But for test entities, it was erroneously set to null. Because this field is an exception, I made it so the current value in manage is used instead of null.

PR #1346, opened 2025-01-09

johanib:
Prior to this change, when saving a test entity, the contractual base field in manage would be set to null.
This is because the contractual base field is used for the production entities. Other random coin fields will not get overwritten.
This change makes sure the current value from Manage is reapplied, so sp-dasboard will not mutate the value.

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Domain/Service/ContractualBaseService.php
+++ b/src/Surfnet/ServiceProviderDashboard/Domain/Service/ContractualBaseService.php
@@ -53,4 +53,20 @@ class ContractualBaseService
         // 4. Set the coin value on the entity
         $entity->getMetaData()?->getCoin()->setContractualBase($contractualBase);
     }
+
+    public function writeContractualBaseForTestEntity(ManageEntity $entity, ?ManageEntity $originalEntity): void
+    {
+        if ($entity->getEnvironment() !== Constants::ENVIRONMENT_TEST) {
+            return;
+        }
+
+        if ($originalEntity === null) {
+            return;
+        }
+
+        $originalContractualBase = $originalEntity->getMetaData()?->getCoin()->getContractualBase();
+        if ($originalContractualBase !== null) {
+            $entity->getMetaData()?->getCoin()->setContractualBase($originalContractualBase);
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


tests/unit/Domain/Service/ContractualBaseServiceTest.php
```diff
@@ -55,6 +55,27 @@ public function testWriteContractualBase(
...
+            Constants::ENVIRONMENT_TEST,
+            Constants::TYPE_SAML,
+            Constants::SERVICE_TYPE_INSTITUTE,
+        );
+        $pristine->getMetaData()->getCoin()->setContractualBase(Constants::CONTRACTUAL_BASE_IX);
+
+        $this->service->writeContractualBaseForTestEntity($entity, $pristine);
+
+        $this->assertSame(Constants::CONTRACTUAL_BASE_IX, $entity->getMetaData()->getCoin()->getContractualBase());
+    }
```
MKodde, line 77: Nice test, personally I would add some coverage for the non-happy flows. 

Testing that contractual base is not set when:
- ENV=prod
- The pristine entity does not have the contractual base coin

src/Surfnet/ServiceProviderDashboard/Domain/Service/ContractualBaseService.php
```diff
@@ -53,4 +53,16 @@ public function writeContractualBase(ManageEntity $entity): void
         // 4. Set the coin value on the entity
         $entity->getMetaData()?->getCoin()->setContractualBase($contractualBase);
     }
+
+    public function writeContractualBaseForTestEntity(ManageEntity $entity, ManageEntity $originalEntity): void
+    {
+        if ($entity->getEnvironment() !== Constants::ENVIRONMENT_TEST) {
```
MKodde, line 59: Consider using the `isProduction` method of the `ManageEntity` (or adding an `isTest` variant for this purpose.

MKodde (approved):
Awesome solution, great to see you boiled the solution down to the most to the point changeset!

See some small suggestions, but feel free to merge as-is if you don't see the point in them.

tests/unit/Domain/Service/ContractualBaseServiceTest.php
johanib: added

src/Surfnet/ServiceProviderDashboard/Domain/Service/ContractualBaseService.php
johanib: Changing this is very inconvenient in the test mocks, so left it as is.

PR #1347, opened 2025-01-09

johanib:
Prior to this change, when saving a test entity, the contractual base field in manage would be set to null. This is because the contractual base field is used for the production entities. Other random coin fields will not get overwritten. This change makes sure the current value from Manage is reapplied, so sp-dasboard will not mutate the value.


johanib:
Will create updated pr

PR #1348, opened 2025-01-13

johanib:
Prior to this change, when saving a test entity, the contractual base field in manage would be set to null. This is because the contractual base field is used for the production entities. Other random coin fields will not get overwritten. This change makes sure the current value from Manage is reapplied, so sp-dasboard will not mutate the value.

(cherry picked from commit e548d941f4603a9fe8f11354b05b561040213e9d)


johanib:
Backport of https://github.com/SURFnet/sp-dashboard/pull/1348 to 6.2 release

PR #1349, opened 2025-01-14
Fix: Don't reset contractual base on production entities

johanib:
Prior to this change, when a change request was sent, the contractualBase would be set to null, because it is not present in the form. This change updates the contractualbase if necessary on update.

The expected behaviour is:
When the field is not present, add it with the correct value 
When the field is present and the value is incorrect update it 
When the field is present and the value is correct, leave it untouched

Resolves https://github.com/SURFnet/sp-dashboard/issues/1343

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Checkout openconext-devconf
         uses: actions/checkout@v4
-        with: 
+        with:
           path: OpenConext-devconf
           repository: OpenConext/OpenConext-devconf
 
--- a/tests/integration/Application/CommandHandler/Entity/EntityChangeRequestCommandHandlerTest.php
+++ b/tests/integration/Application/CommandHandler/Entity/EntityChangeRequestCommandHandlerTest.php
@@ -1,0 +1,179 @@
+<?php
+
+/**
+ * Copyright 2025 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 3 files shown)


tests/integration/Application/CommandHandler/Entity/EntityChangeRequestCommandHandlerTest.php
```diff
@@ -0,0 +1,179 @@
...
+
+        $this->ticketService->shouldReceive('createJiraTicket')
+            ->once()
+            ->andReturn($ticket);
+
+        $this->entityChangeRequestRepository->shouldReceive('openChangeRequest')
+            ->once()
+            ->withArgs(function($entity, $pristineEntity) use ($expectedDiff){
+                if($pristineEntity->diff($entity)->getDiff() !== $expectedDiff){
+                    var_dump($pristineEntity->diff($entity)->getDiff(), 'is not equal to' , $expectedDiff);
```
MKodde, line 124: :thinking: is this a dev leftover? Can you not do a $this->fail? or throw an exception?

tests/integration/Application/CommandHandler/Entity/EntityChangeRequestCommandHandlerTest.php
```diff
@@ -0,0 +1,179 @@
+<?php
+
+/**
+ * Copyright 2024 SURFnet B.V.
```
MKodde, line 4:
```suggestion
 * Copyright 2025 SURFnet B.V.
```

.github/workflows/test-integration.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: Checkout openconext-devconf
         uses: actions/checkout@v4
-        with: 
+        with:
```
MKodde, line 28: Personally I would not include this change in the PR. This change seems unrelated to the changeset, was it autoformatting? Or maybe it is good old boyscout work.

Feel free to leave it and resolve this feedback by yourself.

MKodde (changes requested):

.github/workflows/test-integration.yml
johanib: It's auto formatting, I cannot reinstate the space, so gonna leave it out.

tests/integration/Application/CommandHandler/Entity/EntityChangeRequestCommandHandlerTest.php
johanib: Good point, it should have been a fail ofc. Thanks for pointing it out.

PR #1351, opened 2025-01-20
Fix: Don't reset contractual base on production entities

johanib:
Prior to this change, when a change request was sent, the contractualBase would be set to null, because it is not present in the form. This change updates the contractual base if necessary on update.

The expected behaviour is:
When the field is not present, add it with the correct value When the field is present and the value is incorrect update it When the field is present and the value is correct, leave it untouched

Resolves https://github.com/SURFnet/sp-dashboard/issues/1343

(cherry picked from commit 0085f9a791f721b2560017905f8512aea783813f)

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Checkout openconext-devconf
         uses: actions/checkout@v4
-        with: 
+        with:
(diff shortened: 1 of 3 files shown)


johanib:
Same as https://github.com/SURFnet/sp-dashboard/pull/1349 but to 6.2 branch

PR #1354, opened 2025-01-27
DRAFT 1322 - fix type of service

johanib:


tests/unit/Domain/ValueObject/TypeOfServiceCollectionTest.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Surfnet\ServiceProviderDashboard\Tests\Unit\Domain\ValueObject;
+
+use PHPUnit\Framework\TestCase;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\TypeOfService;
+use Surfnet\ServiceProviderDashboard\Domain\ValueObject\TypeOfServiceCollection;
+
+class TypeOfServiceCollectionTest extends TestCase
+{
+
+    public function testFiltersHidden()
```
MKodde, line 30: Personally I would widen the test coverage of the filter. 

1. What happens when there are no items in the collection and you call filter? 
2. Or what happens if all values are filtered out?

I know these are no cases that should ever occur. But it might be good to test the collection according to those possibilities.

tests/webtests/EntityCreateOidcngTest.php
MKodde, line 1: I wonder, can the webtest be used to validate that the 'hidden' types of service are never in the DOM?

assets/type_of_service.json
```diff
@@ -1,7 +1,8 @@
 [
   {
     "typeEn" : "SURF",
-    "typeNl" : "SURF"
+    "typeNl" : "SURF",
+    "typeHidden": true
```
MKodde, line 5: Looking at the name, at first I expected these TOS (type of service) to become hidden form fields. While studying the actual solution. I found out you solved this problem in a much more elegant way.

I think the property could be slightly more descriptive. Although it now correctly conveys the purpose. It however does suggest `input type = hidden` to me for some reason. But now, the suggestion for giving a more descriptive name.

`isDeferred`
`isSupplementary`

Feel free to leave as is!

src/Surfnet/ServiceProviderDashboard/Domain/Repository/TypeOfServiceRepositoryFromConfig.php
```diff
@@ -59,7 +59,8 @@ private function load(): void
         }
         $this->collection = new TypeOfServiceCollection();
         foreach ($data as $entry) {
-            $typeOfService = new TypeOfService($entry->typeEn, $entry->typeNl);
+            $typeHidden = $entry?->typeHidden ?? false;
```
MKodde, line 62: Nitpick: My opinion is that the use of the null coalescing operator `??` combined with the null safe operator `?->` makes this line somewhat hard to understand at once. Consider rewriting this line into maybe an if-statment?

Might be my lesser experience in using the 'modern' operators :sauropod: . And feel free to leave as is.

MKodde (approved):
Looks great @johanib !

I only performed a code review, did not test this in the browser. If you'd like me to do that, please ping me for a quick peer review at your workstation.

assets/type_of_service.json
johanib: I understand your point.
I think it's difficult to find a more clear term.

The main reason for 'typeHidden', is that when you are updating `type_of_service.json`, `typeHidden` is the most clear option.
On first glance I would not know what to expect from options named isDeferred or isSupplementary.

So will leave as is.

src/Surfnet/ServiceProviderDashboard/Domain/Repository/TypeOfServiceRepositoryFromConfig.php
johanib: :t-rex: 
```suggestion
            $typeHidden = $entry->typeHidden ?? false;
```

johanib: Good sense of smell, it was redundant. Fixed

tests/webtests/EntityCreateOidcngTest.php
johanib: Yes! Added.

src/Surfnet/ServiceProviderDashboard/Domain/Repository/TypeOfServiceRepositoryFromConfig.php
MKodde: Much better like this

tests/unit/Domain/ValueObject/TypeOfServiceCollectionTest.php
MKodde: Thanks for adding the test coverage :1st_place_medal:

MKodde (approved):

Issue #1364, opened 2025-05-13
Bug met JIRA en SP dashboard


Tyskai:
Log regel
> "message":"Unable to create the Jira issue.","context":["JSON property \"expand\" in class \"JiraRestApi\\Issue\\IssueSearchResult\" must not be NULL"],"level":500,"level_name":"CRITICAL"

Andere belangrijke info
> Er is afgelopen woensdag een update van JIRA geweest. Er zou geen aanpassingen geweest moeten zijn aan de API. Er staat een vraag uit, maar daarover weten wij op dit moment niet meer dan dit.

Er zijn verschillende "verzoeken":
- Productieverzoeken leiden tot een foutmelding in de UI (change komt wel in Manage terecht, maar geen JIRA ticket).
- Change requests aan een productiekoppelingen leiden ook tot dezelfde foutmelding (in UI). In Manage komt geen change binnen, en geen ticket in JIRA. Deze foutmelding:
 ![Image](https://github.com/user-attachments/assets/2ea16e03-f71c-4917-974c-b28cfba85069)

- Connection request en delete request lijken wel goed te gaan (komen wel in Manage, krijgt wel een ticket in JIRA).

Het lijkt nu sterk samen te hangen met
[2025-05-14T16:01:20] [ERROR] (unknown)
>   ➤ Message : {"errorMessages":[],"errors":{"fixVersions":"Field 'fixVersions' cannot be set. It is not on the appropriate screen, or unknown."}}" at [JiraClient.ph](http://jiraclient.ph/)p line 286","context":{"exception":{"class":"JiraRestApi\JiraException","message":"CURL HTTP Request Failed: Status Code : 400, URL:https://sd-jira-staging.surf.nl/jira/rest/api/2/issue[\nError](https://sd-jira-staging.surf.nl/jira/rest/api/2/issue%5CnError) Message : {"errorMessages":[],"errors":{"fixVersions":"Field 'fixVersions' cannot be set. It is not on the appropriate screen, or unknown."}}","code":400,"file":"/var/www/html/vendor/lesstif/php-jira-rest-client/src/[JiraClient.ph](http://jiraclient.ph/)p:286"}},"level":500,"level_name":"CRITICAL","channel":"request","datetime":"2025-05-14T16:01:20.217118+02:00","extra":{}}

Deze komen hier voor:
> https://github.com/lesstif/php-jira-rest-client/issues/547
> https://github.com/lesstif/php-jira-rest-client/pull/548

[...]

PR #1365, opened 2025-05-13
Fix? for Jira tickets not being created

johanib:
Prior to this change, an error occurred looking up jira issues, resulting in jira issues not being created. By updating this library, error should no longer occur as this version of the library does not do strict type checking.

https://github.com/SURFnet/sp-dashboard/issues/1364

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


pablothedude (approved):

johanib:
Webtests are broken on main and 62, seem unrelated to this change.

PR #1433, opened 2026-02-27
Upgrade: major upgrades for Yarn

kayjoosten:
First we fixed the cypress test as that should have been working. Therefore its now part of the Github Actions again. 

We also made a mayor upgrade to alot of packages. 

For some of the cypress test they had a valid reasont to fail. Those where the accesability test. Some of the color did not have the correct contrast to other colors.

a11y-before-after


.github/workflows/test-integration.yml
```diff
@@ -93,4 +94,3 @@ jobs:
         working-directory: spdashboard
         run: |
           ${DOCKER_COMPOSE_CYPRESS} cypress run --browser chrome -C ./ci/qa/cypress.config.ts
-        continue-on-error: true
```
johanib, line 96: 🎊

ci/qa/phpstan-baseline.php
```diff
@@ -4507,24 +4501,6 @@
...
-	// identifier: missingType.parameter
-	'message' => '#^Method Surfnet\\\\ServiceProviderDashboard\\\\Infrastructure\\\\DashboardBundle\\\\Service\\\\CurlLogoValidationHelper\\:\\:validateLogo\\(\\) has parameter \\$url with no type specified\\.$#',
-	'count' => 1,
-	'path' => __DIR__ . '/../../src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/CurlLogoValidationHelper.php',
-];
-$ignoreErrors[] = [
-	// identifier: missingType.return
-	'message' => '#^Method Surfnet\\\\ServiceProviderDashboard\\\\Infrastructure\\\\DashboardBundle\\\\Service\\\\LogoValidationHelperInterface\\:\\:validateLogo\\(\\) has no return type specified\\.$#',
-	'count' => 1,
-	'path' => __DIR__ . '/../../src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/LogoValidationHelperInterface.php',
```
johanib, line 4520: 🥳

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -718,16 +718,6 @@ public function getAcsLocations(): ?array
         return null;
     }
 
-    public function getOrganizationUnitAttribute(): ?Attribute
```
johanib, line 721: Are you sure about that?? 

Seems it's referenced here `assets/Resources/metadata/attributes.json:74`

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
```diff
@@ -477,16 +475,6 @@ public function getProtocol(): string
         return Constants::TYPE_SAML;
     }
 
-    public function getOrganizationUnitAttribute(): ?Attribute
```
johanib, line 480: Why is this removed?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Resources/config/services.yml
```diff
@@ -308,12 +308,15 @@ services:
...
     tags:
       - { name: validator.constraint_validator, alias: logo }
 
-  Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service\CurlLogoValidationHelper:
-    arguments: [ '@logger' ]
+  surfnet.logo.http.guzzle:
+    class: GuzzleHttp\Client
+
+  Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service\GuzzleLogoValidationHelper:
+    arguments: [ '@surfnet.logo.http.guzzle', '@logger', '%env(bool:manage_verify_ssl)%' ]
```
johanib: Currently, this requires an entry in the changelog, as the environment wont boot without the env variable being present. 

1. I would like it to be optional.
2. If possible, make a `parameter`, load it from the env once into that parameter, instead of reading the env multiple times.

(is there no way to configure this on guzzle at a global level?)

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/GuzzleLogoValidationHelper.php
```diff
@@ -17,64 +17,68 @@
...
 namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service;
 
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\GuzzleException;
 use Psr\Log\LoggerInterface;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Exception\LogoInvalidTypeException;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Exception\LogoNotFoundException;
 
-class CurlLogoValidationHelper implements LogoValidationHelperInterface
+class GuzzleLogoValidationHelper implements LogoValidationHelperInterface
```
johanib: Why is this needed?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
```diff
@@ -37,8 +36,15 @@ class ValidLogoValidator extends ConstraintValidator
      */
     final public const STATUS_INVALID_TYPE = 'validator.logo.wrong_type';
 
-    public function __construct(private readonly LogoValidationHelperInterface $logoValidationHelper)
-    {
+    /**
+     * The status code used when the logo URL resolves to a private or reserved address.
+     */
+    final public const STATUS_PRIVATE_HOST = 'validator.logo.private_host';
```
johanib, line 42: Why is this needed?

docker-compose-ci-webtests.yml
```diff
@@ -0,0 +1,6 @@
+---
+services:
+  spdashboard:
+    environment:
+      APP_ENV: test
+      manage_verify_ssl: "false"
```
johanib: 🤔

docker-compose-ci.yml
```diff
@@ -3,8 +3,9 @@ services:
   spdashboard:
     image:  ghcr.io/openconext/openconext-basecontainers/php82-apache2-node20-composer2:latest
     environment:
-      APP_ENV: test
+      APP_ENV: dev
```
johanib: I would expect `ci` or `test` ?

templates/Service/overview.html.twig
```diff
@@ -2,6 +2,8 @@
 
 {% block body_container %}
     <main>
+        {{ render(controller('Surfnet\\ServiceProviderDashboard\\Infrastructure\\DashboardBundle\\Controller\\SiteNoticeController::showGlobalSiteNotice')) }}
```
johanib, line 5: Why is this needed?

 * 3

templates/Service/overview.html.twig
```diff
@@ -17,7 +19,7 @@
 
         {% for service in services %}
             <article class="fieldset card service">
-                {% set shortName = service.name|replace({' ': '-'}) %}
+                {% set shortName = service.name|replace({' ': '-', '.': ''}) %}
```
johanib, line 22: Why is this needed?

* 2

johanib (commented):
Congrats on getting the cypress tests green again and everything upgraded!

Some concerns:
* `manage_verify_ssl` is not a bad solution perse, but you needed to add this in way to many places to keep things running and it requires a new variable on production. We should be able to come to a more elegant solution. 
* Some changes seem way out of scope for this PR (probably needed, but unclear why). Probably a good reason, but it is unclear why these changes are needed. This should be clarified in the commit messages. `Prior to this change, xxx \n \n This change, xxx`. Not sure if possible, but consider cleaning up/merging some commits. 

Maybe we can discuss before rework to clear things up.

templates/Service/overview.html.twig
kayjoosten: overview.html.twig and my-services.html.twig both override the entire body_container block from base.html.twig. Because the site notice is rendered inside that block in the base template, it gets lost when the block is overridden. The render() call was added explicitly to both templates so the notice still appears on those pages.

templates/Service/overview.html.twig
kayjoosten: Service 2 in the test fixtures is "Ibuildings B.V." — without this change, the generated input id becomes add-for-test-Ibuildings-B.V.. The cy.htmlvalidate() check extends html-validate:recommended which enforces the id-pattern rule (kebab-case). Dots are not valid in kebab-case IDs, so the rule fails on the service overview page. Removing dots from the shortName fixes the htmlvalidate error.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
kayjoosten: A previous security commit added isPrivateHost() to ValidLogoValidator to block logo URLs that resolve to private/reserved IP ranges. In devconf and CI, hostnames like engine.dev.openconext.local resolve to Docker bridge IPs (172.x.x.x), which are in that private range, so logo validation was always failing in those environments. The environment injection skips the check outside of prod where the security concern actually applies.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveSamlEntityCommand.php
kayjoosten: I thought it wasnt use thanks for addressing this ill place it back

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/GuzzleLogoValidationHelper.php
kayjoosten: Too be honest i did this change early on and i didnt notice it was already curl i though i made it curl Ill revert is as the that will also eliminate the ssl env var i think.

templates/Service/overview.html.twig
johanib: okay, but would the proper fix not be to update the cypress selector, instead of changing the functionality of the app to match cypress? (not sure I understand correctly)

templates/Service/overview.html.twig
johanib: So this was a bug (already currently on prod), which is now fixed?

kayjoosten: Yes thats correct

johanib: Cool :partying_face: 
Probably good to make a note of the bugs you found  and fixed in the changelog, so the PM can validate / check.

templates/Service/overview.html.twig
kayjoosten: Maybe i don't understand correctly i thought it should always be kabab case for the shortname. That is what i could make out of the test. So if its not that important that yes it should be changed in the test. If the kabab case is mandatory than this was also a bug in in prod.

docker-compose-php-tests.yml
```diff
@@ -0,0 +1,5 @@
+---
+services:
+  spdashboard:
+    environment:
+      APP_ENV: test
```
johanib, line 5: Seems weird to set the app_env to test when in 'ci' mode?

ci/qa/phpstan.neon
```diff
@@ -4,4 +4,5 @@ parameters:
     level: 9
     paths:
         - ../../src
+    reportUnmatchedIgnoredErrors: false
```
johanib: 😱 

Unmatched ignored errors should be removed from the baseline. Simlpy update the baseline.

config/services.yaml
```diff
@@ -8,6 +8,8 @@ parameters:
     router.request_context.base_url: ''
     acs_location_route_name: 'dashboard_saml_consume_assertion'
     authentication_context_class_ref: ~
+    env(manage_verify_ssl): 'true'
+    manage_verify_ssl: '%env(bool:manage_verify_ssl)%'
```
johanib: I'm still struggling with this. Why was it not needed before?

johanib (commented):
Still struggling a bit with this. Lets go over this tomorrow in a meeting.

I tried doing some cleanup: https://github.com/SURFnet/sp-dashboard/pull/1441

But I think the build fails :(

Which is putting the finger on the sore place. It feels like a whole bunch of parameters need to be set correctly, but while looking at the project, it is not clear which. Which is probably your struggle as well. So lets put some though into how we can tackle this long term / how a proper setup would look, even if it's out of scope for now.

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -726,6 +728,7 @@ public function getOrganizationUnitAttribute(): ?Attribute
     public function setOrganizationUnitAttribute(?Attribute $organizationUnitAttribute): void
     {
         $this->organizationUnitAttribute = $organizationUnitAttribute;
+        $this->setAttribute('organizationUnitAttribute', $organizationUnitAttribute);
```
johanib: Did you find this bug? Please create a ticket so the TPM can track it and it gets into the release notes. 

* 1

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Service/CurlLogoValidationHelper.php
```diff
@@ -34,23 +34,22 @@ public function __construct(private readonly LoggerInterface $logger)
...
+    public function validateLogo(string $url): string
     {
         $this->logger->debug(sprintf('Validating logo: "%s" using curl', $url));
 
         $ch = curl_init();
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
         curl_setopt($ch, CURLOPT_VERBOSE, 1);
-        curl_setopt($ch, CURLOPT_HEADER, 1);
+        curl_setopt($ch, CURLOPT_HEADER, 0);
```
johanib: If this is not needed, maybe remove it? Or should it be set to 0?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
```diff
@@ -47,32 +53,39 @@ public function validate($value, Constraint $constraint): void
             return;
         }
 
+        if ($this->environment === 'prod' && $this->isPrivateHost($value)) {
```
johanib: I think the validator should receive an injected parameter `bool $allowLogoPrivateHosts`.

Using the environment as a check is a bit of a smell.

johanib: Also, what is the scenario we are covering here? Is including local images prohibited? Is it something specific to tests?

README.md
```diff
@@ -79,6 +81,15 @@ Some remarks:
 3. Running tests in PROD mode is encouraged (this disengages the web debug toolbar) saving false positives in a11y and html validation errors
 4. A service with ID = 2 should be present, it should have the add production entities option set.
 
+## Overview of available environments
+
+| Environment | Description                                                                  |
+|-------------|------------------------------------------------------------------------------|
+| dev         | Local development, real services or stubs, developer ergonomics              |
+| test        | PHPUnit (unit + integration + webtests (panther)), mock kernel, no real HTTP |
+| ci          | Cypress e2e, real kernel, real services under self-signed TLS, seeded DB     |
```
johanib, line 90: CI: Performs actual requests (instead of mocking them).  CI uses devcnof, test does not.

config/services.yaml
johanib: Discussed: Because we now use cypress to actually perform requests. 
Previously, the panther webtest, does not perform actual requests to manage.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
johanib: This is a security improvement. Is better that it was. This improvement partially prevents the server from fetching private resources.

johanib:
1 unresolved comment & the build is red. 


After those are fixed, we can merge and move on to the other prs.

cypress/support/commands/createEntity.js
```diff
@@ -114,7 +122,7 @@ Cypress.Commands.add('verifyCreation', () => {
     cy.viewEntity();
     cy.checkContainsValue('ACS location', 'https://oidc.dev.support.surfconext.nl/saml/SSO');
     cy.checkCorrectTextValue('Entity ID', 'https://tiffany.aching.do/id');
-    cy.checkCorrectTextValue('Logo URL', 'https://generative-placeholders.glitch.me/image?width=600&height=300');
+    cy.checkCorrectTextValue('Logo URL', 'https://engine.dev.openconext.local/images/logo.png');
```
johanib, line 125: 👍

johanib (approved):

PR #1442, opened 2026-03-19
fix: organizationUnitAttribute silently dropped from Manage payload

kayjoosten:
## What

\`setOrganizationUnitAttribute()\` in both \`SaveSamlEntityCommand\` and \`SaveOidcngEntityCommand\` stored the value in a dedicated property but never called \`setAttribute()\`. \`EntityMergeService\` builds the Manage payload via \`getAttribute('organizationUnitAttribute')\`, which reads from the \`$attributes[]\` array — not the dedicated property. The attribute was therefore always null and silently dropped from the payload.

\`SaveOidcngEntityCommand\` was also missing the property declaration entirely.

## Why it wasn't caught

The setter compiled fine and no existing tests verified that the attribute round-tripped through \`getAttribute()\` after being set via the setter.

## Changes

- \`SaveSamlEntityCommand\`: add \`setAttribute()\` call in \`setOrganizationUnitAttribute()\`
- \`SaveOidcngEntityCommand\`: add missing property declaration + \`setAttribute()\` call
- \`CHANGELOG.md\`: note the fix under Unreleased

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
+++ b/src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
@@ -728,6 +728,7 @@ class SaveOidcngEntityCommand implements SaveEntityCommandInterface
     public function setOrganizationUnitAttribute(?Attribute $organizationUnitAttribute): void
     {
         $this->organizationUnitAttribute = $organizationUnitAttribute;
+        $this->setAttribute('organizationUnitAttribute', $organizationUnitAttribute);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
```diff
@@ -728,6 +728,7 @@ public function getOrganizationUnitAttribute(): ?Attribute
     public function setOrganizationUnitAttribute(?Attribute $organizationUnitAttribute): void
     {
         $this->organizationUnitAttribute = $organizationUnitAttribute;
+        $this->setAttribute('organizationUnitAttribute', $organizationUnitAttribute);
     }
```
MKodde: I see why this fixed the problem. The attributes are managed in the `attributes` field, and no longer via the attribute specific fields (which would be the case before the optimization in 2022). 

The question is: why was the `organizationUnitAttribute` not also part of this optimization back then?

My assumption and proposed solution is to completely get rid of the getter and setter for this field, and let the 'magic' accessors do their job.

MKodde (changes requested):

johanib (approved):

src/Surfnet/ServiceProviderDashboard/Application/Command/Entity/SaveOidcngEntityCommand.php
johanib: :heavy_check_mark:

PR #1444, opened 2026-03-19
fix: service card HTML ids contain invalid dots

kayjoosten:
## What

Service cards on the overview page generated HTML element IDs with dots in them (e.g. \`add-for-test-Ibuildings-B.V.\`).

## Why it's a problem

Dots are not valid in kebab-case HTML IDs. The \`html-validate:recommended\` ruleset enforces \`id-pattern\`, which rejects them. This was a live bug on production for any service whose name contains a dot.

## Fix

Strip \`.\` in addition to replacing spaces with \`-\` when building the ID from \`service.name\`.

## Changes

- \`templates/Service/overview.html.twig\`: add \`'.': ''\` to the \`replace\` filter map
- \`CHANGELOG.md\`: note the fix under Unreleased

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 Remove integration with Teams
 Add integration with Invite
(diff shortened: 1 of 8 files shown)


johanib (approved):
allowLogoPrivateHosts has bleeded through. Just a heads up, if you dont think its a problem, leave it in.

PR #1483, opened 2026-06-02
Clarify / fix how the webtests are run in devconf

johanib:
The defconf path conflicted with https://github.com/OpenConext/Stepup-Gateway/blob/main/.github/workflows/test-integration.yml#L21


kayjoosten (approved):

kayjoosten:
Only test are still failing

PR #1488, opened 2026-06-04
Feature/issue 1379 resource server publish error

### LINKED ISSUE #1379: fout bij het toevoegen van een resource server aan een OIDC-client
SPD instantie: https://sp.surfconext.nl/entity/detail/1684/32becf66-50bc-42bc-916c-d8c3c1e26794/production

Hierbij lukte het de gebruiker niet om een resource server toe te voegen. Mij lukte dat ook niet via het SPD. Wel via Manage.

De volgende foutmelding verscheen in SPD:

Image

Bijbehorend ticket: https://servicedesk.surf.nl/jira/browse/CXT-84467

Daarna nog een keer gebeurd, zie comment

kayjoosten:
# Bug #1379 — Error when editing a resource server

## Symptom

When a user edits an existing resource server and saves, both a **success message and an error message** appear simultaneously. The change is never applied in Manage.

## Root cause

When a user edits a resource server, form data is merged into the existing Manage entity via `MetaData::merge()`.

The resource server form has **no `NameIDFormat` field** — the user never needs to set it. As a result, `SaveOidcngResourceServerEntityCommand::getNameIdFormat()` always returns `null`.

`MetaData::merge()` did the following:

```php
// Wrong: if the new value is null → overwrite with null
$this->nameIdFormat = is_null($metaData->getNameIdFormat()) ? null : $metaData->getNameIdFormat();
```

The correct logic is: *if the new value is null, preserve the existing value.* This is what `Coin::merge()` already does correctly:

```php
// Correct: if the new value is null → keep the old value
$this->excludeFromPush = is_null($coin->getExcludeFromPush()) ? $this->excludeFromPush : $coin->getExcludeFromPush();
```

## Step-by-step failure

1. Existing resource server in Manage has `NameIDFormat = "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"`
2. After merge, `NameIDFormat = null` (the form supplies no value for this field)
3. The diff between the two versions includes `metaDataFields.NameIDFormat: null`
4. That `null` is sent to Manage via the merge API
5. Manage rejects it and returns a response without `id`
6. SPD throws an error — but due to a secondary bug also shows a success message → confusing dual message

## When was this introduced?

Present since **2020-09-08** (`30d6531a` — "Add Oidc TNG resource server support"). The wrong null-handling was in the code from day one when resource server support was added. Only formally reported in 2025 because resource servers are rarely edited after initial creation.

## Fix

| File | Change |
|------|--------|
| `MetaData::merge()` | Preserve `nameIdFormat` when the incoming value is `null` |
| `EntityControllerTrait` | Resource servers in production bypass the change-request flow and are published directly |
| `EntityEditController` | Success flash is only shown when no error flash was added by the command handler |

### DIFF
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
@@ -156,7 +156,7 @@ trait EntityControllerTrait
             case $entity->getEnvironment() === Constants::ENVIRONMENT_TEST:
                 $publishEntityCommand = new PublishEntityTestCommand($entity, $applicant);
                 break;
-            case $isEntityChangeRequest:
+            case $isEntityChangeRequest && !$entity?->isOidcngResourceServer():
                 $publishEntityCommand = new EntityChangeRequestCommand($entity, $applicant);
                 break;
             case $entity->getEnvironment() === Constants::ENVIRONMENT_PRODUCTION:
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityEditController.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityEditController.php
@@ -94,7 +94,9 @@ class EntityEditController extends AbstractController
                         $response = $this->publishEntity($entity, $command, $isProductionEntityEdit, $request);
 
                         if ($response instanceof Response) {
-                            if ($environment !== Constants::ENVIRONMENT_PRODUCTION) {
+                            if ($environment !== Constants::ENVIRONMENT_PRODUCTION
+                                && !$flashBag->has('error')
+                            ) {
                                 $this->addFlash('info', 'entity.edit.metadata.flash.success');
                             }
                             return $response;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- Fix: editing an existing resource server (oauth20_rs) always failed with a confusing dual success+error message and no change applied in Manage. Root cause: `MetaData::merge()` overwrote `nameIdFormat` with `null` when the form command returned `null` (the RS form has no NameIDFormat field), causing the diff to send `metaDataFields.NameIDFormat: null` to Manage, which rejected the update. Additionally, published production RS entities were incorrectly routed through the change-request flow instead of being published directly.
 Remove integration with Teams
 Add integration with Invite
 - Fix: `organizationUnitAttribute` was silently dropped from the Manage payload because `setOrganizationUnitAttribute()` did not call `setAttribute()`. The attribute is now correctly included when saving SAML and OIDC entities.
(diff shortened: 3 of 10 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityEditController.php
```diff
@@ -94,7 +94,9 @@ public function edit(Request $request, string $environment, string $manageId, in
                         $response = $this->publishEntity($entity, $command, $isProductionEntityEdit, $request);
 
                         if ($response instanceof Response) {
-                            if ($environment !== Constants::ENVIRONMENT_PRODUCTION) {
+                            if ($environment !== Constants::ENVIRONMENT_PRODUCTION
+                                && !$flashBag->has('error')
```
johanib, line 98: 🤔 `!$flashBag->has('error')` 

This is setup for failure... ?

The intent is to prevent the error from being shown twice?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
```diff
@@ -156,7 +156,7 @@ private function createPublishEntityCommandFromEntity(
             case $entity->getEnvironment() === Constants::ENVIRONMENT_TEST:
                 $publishEntityCommand = new PublishEntityTestCommand($entity, $applicant);
                 break;
-            case $isEntityChangeRequest:
+            case $isEntityChangeRequest && !$entity?->isOidcngResourceServer():
```
johanib, line 159: man, this switch 🤯 

I'm assuming the $entity is already in production mode at this point. Would it be a more systematic fix to move the `case` below this one up a position? So that it gets checked/executed first? (or that would break a lot of stuff).

My cornern here is: Why is oidcng resource server different from any other service that it needs an exception here?

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityEditController.php
kayjoosten: Almost the case now is that it shows a succes and a error. So when there is a error thrown it ofcourse should not show the succes response

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
kayjoosten: Moving ENVIRONMENT_PRODUCTION above $isEntityChangeRequest would route ALL production entity edits through direct publish, bypassing the change-request/service desk flow entirely. That would break SAML and OIDC entities in production which specifically need service desk review before changes are applied.

RS entities are the exception because they do not go through the service desk changes are applied directly (same as test entities). This is already visible in PublishEntityProductionCommandHandler::isNewResourceServer() which treats new RS entities as direct publishes. This change extends that same logic to edits.

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityEditController.php
johanib: Whats the scenario here? Why is there a success message? Are multiple operations performed? If so, the status messages should provide that context: "Manage push succeeded" / "Push to XXX failed." If both are to manage, it should not push a "success" message in the first place.

What if a 'warning' message is potentially shown?

If this is not fixable another reasonable way, we can accept it as a pragmatic fix..

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Controller/EntityControllerTrait.php
johanib: :+1:

johanib (approved):

johanib (approved):

PR #1489, opened 2026-06-04
Remove python / java from daily security check.

johanib:
Cleaning this up prevents dependabot creating useless pr's for these irrelevant github actions


kayjoosten (commented):
There is only a merge conflict left to solve but looks good to me

PR #1503, opened 2026-07-10
fix: block server-side fetch of metadata/logo URLs pointing at internal hosts

kayjoosten:
## Summary
- Metadata URL was fetched server-side with no check on the target host, allowing SSRF against internal infrastructure (loopback, private ranges, our own internal ranges even though those are publicly addressed, cloud metadata endpoints).
- Enforced at the actual fetch point (`Fetcher::fetch`), not only in form validation, since metadata import can bypass the entity form.
- Redirect hops are re-checked; outgoing request is pinned to the validated IP to close a DNS-rebinding gap.
- Logo URL validator reused the same shared check (previously only blocked private ranges, not our own public-but-internal ranges).
- TLS certificate verification, previously unconditionally disabled on the metadata fetch, is now enabled by default.

## Test plan
- [x] `composer check` (phpcs, phpmd, phpstan, unit + integration tests) passes
- [x] New unit tests for `HostBlocklistChecker` (private ranges, explicit CIDR denylist, cloud metadata endpoint)
- [x] New/updated tests for `Fetcher`, `ValidMetadataUrlValidator`, `ValidLogoValidator` covering blocked-host and allow-flag behavior

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -74,6 +74,14 @@ saml_remote_idp_certificate=MIIDuDCCAqCgAwIBAgIJAPdqJ9JQKN6vMA0GCSqGSIb3DQEBBQUA
 # The default timeout for Curl requests when retrieving metadata
 metadata_url_timeout=30
 
+# CIDR ranges (JSON array) that must not be reachable via a user-supplied
+# metadata/logo URL fetched server-side, in addition to the always-blocked
+# RFC1918/loopback/link-local/reserved ranges. Real internal ranges are
+# environment-specific and must be set by Ansible on deploy — do NOT put
+# real ranges here. The example below is illustrative only: it already
+# falls within the link-local range that is blocked by default.
+metadata_logo_fetch_blocked_ip_ranges='["169.254.169.254/32"]'
+
 # Manage defaults
 
 ## Manage test instance
--- a/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
+++ b/src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
@@ -20,6 +20,7 @@ namespace Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Valida
 
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Exception\LogoInvalidTypeException;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Exception\LogoNotFoundException;
+use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service\HostBlocklistCheckerInterface;
 use Surfnet\ServiceProviderDashboard\Infrastructure\DashboardBundle\Service\LogoValidationHelperInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
@@ -43,6 +44,7 @@ class ValidLogoValidator extends ConstraintValidator
 
     public function __construct(
         private readonly LogoValidationHelperInterface $logoValidationHelper,
+        private readonly HostBlocklistCheckerInterface $hostBlocklistChecker,
         private readonly bool $allowLogoPrivateHosts = false,
     ) {
     }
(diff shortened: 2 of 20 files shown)


src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
```diff
@@ -53,7 +55,7 @@ public function validate($value, Constraint $constraint): void
             return;
         }
```
johanib, line 57: **Logo download is not pinned to the validated IP — DNS-rebinding SSRF bypass remains open**

`isBlocked()` here resolves and checks the host (DNS lookup #1), but `CurlLogoValidationHelper::validateLogo()` then runs a fresh `curl_exec()` on the raw URL with no `CURLOPT_RESOLVE` (DNS lookup #2, independent resolution). An attacker-controlled DNS name can return a public IP for this check and an internal IP (e.g. `169.254.169.254`) for the actual download. This is the same rebinding gap that was closed for `Fetcher`, but the logo path is left unprotected — so the "pinned to the validated IP" guarantee does not hold for logos.

Suggested approach: mirror the `Fetcher` pattern — have `resolve`/`isBlocked` return the validated IP and pass it into `validateLogo()`, then pin with `curl_setopt($ch, CURLOPT_RESOLVE, ["$host:$port:$ip"])` in `CurlLogoValidationHelper` (keeping `CURLOPT_FOLLOWLOCATION` off as it is now).

src/Surfnet/ServiceProviderDashboard/Legacy/Metadata/Fetcher.php
```diff
@@ -59,6 +68,62 @@ public function fetch($url): string
...
+     */
+    private function buildGuzzleOptions(string $url, ?string $ip): array
+    {
+        $guzzleOptions = [
+            'timeout' => $this->timeout,
+            'verify' => $this->verifySsl,
+            'allow_redirects' => false,
+        ];
+
+        $host = parse_url($url, PHP_URL_HOST);
```
johanib, line 128: **Uppercase URL scheme breaks the `CURLOPT_RESOLVE` port, silently disabling the IP pin**

`str_starts_with($url, 'https://')` is case-sensitive. A URL like `HTTPS://host/…` (schemes are case-insensitive and Guzzle lowercases them before connecting on 443) computes port `80`, so the pin becomes `host:80:ip` while curl actually connects to `host:443`. The RESOLVE entry doesn't match, curl falls back to its own DNS resolution, and the rebinding protection is bypassed on the metadata path too.

Suggested approach: derive the scheme case-insensitively, e.g. `strtolower((string) parse_url($url, PHP_URL_SCHEME)) === 'https' ? 443 : 80`.

.env.dist
```diff
@@ -74,6 +74,12 @@ saml_remote_idp_certificate=MIIDuDCCAqCgAwIBAgIJAPdqJ9JQKN6vMA0GCSqGSIb3DQEBBQUA
 # The default timeout for Curl requests when retrieving metadata
 metadata_url_timeout=30
 
+# CIDR ranges (JSON array) that must not be reachable via a user-supplied
+# metadata/logo URL, in addition to the always-blocked RFC1918/loopback/
+# link-local/reserved ranges. Real internal ranges are environment-specific
+# and must be set by Ansible on deploy — do NOT put real ranges here.
+blocked_ip_ranges='["169.254.169.254/32"]'
```
johanib: `blocked_ip_ranges` Can we make this more specified?
The name does not indicate that it is used to specifically block 'uploads' from blocked ip ranges.

I think the comment is accurate, but maybe we can help operators better understand this parameter by adding more context.

johanib: Default value is fine as an example, but note that is an example only, as its within the private ranges that are blocked by default.

johanib (commented):
What I expected from this PR:
* The same fix we applied recently to the logo validater > now applied to metadata

This PR seems to do that, with the addition of more strict checks (IP pinning + redirect re-check + TLS verify) in addition. 👍 

Please check the comments below

src/Surfnet/ServiceProviderDashboard/Infrastructure/DashboardBundle/Validator/Constraints/ValidLogoValidator.php
johanib: ValidLogoValidator now calls resolve($host) → isIpBlocked($ip) → validateLogo($value, $ip); CurlLogoValidationHelper::pinResolvedIp() sets CURLOPT_RESOLVE and keeps FOLLOWLOCATION off. Exactly the suggested approach.

src/Surfnet/ServiceProviderDashboard/Legacy/Metadata/Fetcher.php
johanib: Fetcher.php:131 now uses strtolower((string) parse_url($url, PHP_URL_SCHEME)) === 'https' ? 443 : 80; same helper in CurlLogoValidationHelper.

.env.dist
johanib: Renamed to metadata_logo_fetch_blocked_ip_ranges, with a comment explaining it applies to server-side fetches of user-supplied metadata/logo URLs and is additional to the always-blocked ranges.

johanib: Comment now states real ranges must be set by Ansible on deploy and that the 169.254.169.254/32 example "is illustrative only: it already falls within the link-local range that is blocked by default."

johanib (approved):

